PR #69, opened 2024-11-10
Added test reporters and coverage analysis

Slartibartfass2:


.github/semantic.yml
Merseleo, line 1: Only for my personal interest: what is this?

src/app.css
Merseleo, line 1: Is this because no rebase was done so far? (seems to be same for some other files, like the src/lib/utils, src/routes/theme/...

.gitignore
Merseleo, line 1: Maybe I should add this update to the other PR #78 and then you can rebase on develop and include the updates? Or do you think it is better to add them here?

e2e/demo.test.ts
Merseleo, line 1: Maybe should rename this folder? Suggestion: e2e_tests/

vite.config.ts
```diff
@@ -1,10 +1,24 @@
...
 
 export default defineConfig({
-    plugins: [sveltekit()],
-
+    plugins: [sveltekit(), svelteTesting()],
+    // Tell Vitest to use the `browser` entry points in `package.json` files, even though it's running in Node
+    resolve: process.env.VITEST ? { conditions: ["browser"] } : undefined,
     test: {
-        include: ["test/**/*.{test,spec}.{js,ts}"],
+        include: ["test/**/*.{test,spec}.{js,ts}", "integration/**/*.{test,spec}.{js,ts}"],
```
Merseleo: Do we have an integration directory? For components testing?

.github/workflows/code_quality_checks.yml
Merseleo, line 1: In your config you added something regarding integration tests, do we want an empty job for the integration test already?

.github/workflows/code_quality_checks.yml
```diff
@@ -48,6 +51,17 @@ jobs:
...
+
+            - name: Report code coverage
+              uses: zgosalvez/github-actions-report-lcov@v3
+              with:
+                  coverage-files: coverage/lcov.info
+                  artifact-name: code-coverage-report
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  update-comment: true
+
     e2e-test:
```
Merseleo, line 87: Maybe add to this job the upload of the playwright report (https://playwright.dev/docs/ci).

Merseleo (commented):
Again, I think this is a good basis for all the testing and further configuration will follow when writing the tests. However, maybe you could add a meaningful test example, if this is possible (i. e. for a e2e test and a component test, which is probably not possible so far as we don't have any components completed, and normal tests). 
And as far as I understood examples etc. correctly the configs look good for now.


One last request: maybe you should add a short section about testing in the Readme.md, so how we test, how to run them locally and what is reported in the CI (optionally). Or do you think differently?

.github/semantic.yml
Slartibartfass2: This is a config file for the Semantic PRs GitHub App I added, which is like a workflow and checks whether the commits in a PR adhere to the conventional commits schema.

vite.config.ts
Slartibartfass2: Currently, we don't because we have no tests but as soon as we add them, e.g. in #74 this directory will also be tracked in git.

.github/workflows/code_quality_checks.yml
Slartibartfass2: Both unit and integration tests are executed by vites with the `test:unit` command (I see that this name might be misleading).

.github/workflows/code_quality_checks.yml
Slartibartfass2: Good catch, I added it.

src/app.css
Slartibartfass2: Since this PR has been open for some while and I replaced the default branch main with develop, this PR had main still as base, I changed that now, and this isn't part of the changes anymore.

e2e/demo.test.ts
Slartibartfass2: What do you think about the following structure?
```
test/
├─ e2e/
├─ integration/
└─ unit/
```

.gitignore
Slartibartfass2: See my comment in your PR: https://github.com/SE-UUlm/snowballr-frontend/pull/78#discussion_r1844978171

Slartibartfass2 (commented):
I'll add a real component test in #74, I don't think it makes much sense to add a demo test now.

I'll add a section in the README as soon as we are finished discussing the location of the tests.

.github/workflows/code_quality_checks.yml
Slartibartfass2: I'm currently looking into the [workspace feature](https://vitest.dev/guide/workspace.html) of vitest, which could solve this.

We could have two configs for unit and integration tests and run them separately.

.github/semantic.yml
Merseleo: Ah perfect.

src/app.css
Merseleo: Thank you, it more clearly now.

.gitignore
Merseleo: I saw and updated it. As mine is probably merged before, I think this is fine and further updates of the .gitignore can be added from the other PRs (like the playwright report).

e2e/demo.test.ts
Merseleo: Afaik Svelte suggests to add the unit tests in the folders of the corresponding source files, but imo this structure is the best solution!

If you want to ask @luca-schlecker too or have any other idea, lets discuss it, otherwise resolve this thread on your own.

vite.config.ts
```diff
@@ -1,10 +1,22 @@
...
+    // Tell Vitest to use the `browser` entry points in `package.json` files, even though it's running in Node
+    resolve: process.env.VITEST ? { conditions: ["browser"] } : undefined,
     test: {
-        include: ["test/**/*.{test,spec}.{js,ts}"],
+        reporters: process.env.GITHUB_ACTIONS ? ["default", "github-actions"] : ["default"],
+        sequence: {
+            shuffle: true,
+        },
+        coverage: {
+            include: ["src/**"],
```
Merseleo, line 15: Why now include src/** and not test/**?

.github/workflows/code_quality_checks.yml
Merseleo: Perfect.

.github/workflows/code_quality_checks.yml
Merseleo: Yes, I think you find a very good solution. Imo the config files are also separated better now and the names are not misleading anymore. Perfect, thank you!

.github/workflows/code_quality_checks.yml
```diff
@@ -68,3 +105,81 @@ jobs:
...
+
+            - name: Report code coverage
+              uses: zgosalvez/github-actions-report-lcov@v4
+              with:
+                  coverage-files: final-lcov.info
+                  artifact-name: code-coverage-report
+                  github-token: ${{ secrets.GITHUB_TOKEN }}
+                  update-comment: true
+
+    check:
```
Merseleo, line 169: What is the "check" job?

.github/workflows/code_quality_checks.yml
Merseleo: Perfect, thank you.

vite.config.ts
Slartibartfass2: Here we define which files are tracked for the coverage report. Without this, some config files in the root directory will be included.

e2e/demo.test.ts
Slartibartfass2: Afaik, we already discussed this and decided to have tests separately, which is why I'll resolve this thread.

Merseleo (dismissed):
I think we are on the right track.
The new section in the README.md surely is helpful for new people and for us at beginning too. But one last question (must not be done in this PR): do you think it is helpful to host the report anywhere and link to them in the README.md or add a coverage badge or something like that?

vite.config.ts
Merseleo: Ah ok, makes sense

.github/workflows/code_quality_checks.yml
Slartibartfass2: You can read about it [here](https://svelte.dev/docs/cli/sv-check).

TL;DR it mainly checks if the TypeScript code compiles.

Slartibartfass2:
> I think we are on the right track. The new section in the README.md surely is helpful for new people and for us at beginning too. But one last question (must not be done in this PR): do you think it is helpful to host the report anywhere and link to them in the README.md or add a coverage badge or something like that?

we could use github pages to host, but I am not convinced that's necessary. If you want to check out the coverage, one could either download the report or just run the tests in the repo.

About the coverage badge: Maybe that's possible, but I'm not sure. Perhaps we can research this and add this later.

.github/workflows/code_quality_checks.yml
Merseleo: Ah ok, seems to be useful, good idea

Merseleo (approved):
LGTM

PR #89, opened 2024-11-19
Added paper view

### LINKED ISSUE #34: [Use Case]: Open Paper View

Project Member

### Describe the use case

As project member, I want to open the paper view for a specific paper.

### Acceptance criteria

- show paper view

### Anything else?

Related to #30, #41, #42, #43, #44, #45, #46, #47, #50, #51, #52, #53, #54, #55, #61 and #98

Slartibartfass2:
Closes #34

## What I have made

- the root layout file has now a container with the style that is used in every page of our application
- added a reusable tooltip component
- added some aria labels to existing components for improved accessibility
- added paper view, consisting of the following components
  - bookmark button (logic is implemented in #99 and #100)
  - navigation button = previous/next paper button (due to the tooltip being unable to be an a tag when providing a `href` prop (like a button would do), this button uses `goto` to navigate to the previous/next paper)
  - paper cards are the two card components on the paper view, which display paper information
  - the content of these cards will be implemented in other issues, this also includes the testing for these cards, testing it now wouldn't make sense
  - decision buttons to either accept, decline or mark a paper as undecided (maybe button)
    - there is a component for each decision, which seems a lot, but providing different styles depending on a type/state prop was less readable
- added a setup test file to mock the `goto` method used in the navigation buttons
- the paper view component is shown on the project dependent and independent pages
- this PR adds the general structure and the views will be extended in future issues, I wouldn't add more to it now

## Checklist

  - ~~[ ] unit tests~~ (no logic outside of minor component logic was added)
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (we do not have those yet)


src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -0,0 +1,112 @@
...
+                addPaperToReadingList();
+                break;
+            case "removing":
+            case "added":
+                removePaperFromReadingList();
+                break;
+            default:
+                throw new Error(`unexpected state '${state}'`);
+        }
+    }
```
luca-schlecker: I would prefer having something like this in the Html, dropping `state` entirely and calculating `tooltipText` from `isBookmarked` directly:
```svelte
{#if isHovered}
    {#if isBookmarked}
        <BookmarkMinus />
    {:else}
        <BookmarkPlus />
    {/if}
{:else}
    {#if isBookmarked}
        <Bookmark fill="bg-primary" />
    {:else}
        <Bookmark />
    {/if}
{/if}
```

This would decrease handler complexity and increase readability imo.

src/lib/components/composites/Tooltip.svelte
```diff
@@ -0,0 +1,57 @@
...
+        {#snippet trigger()}
+            I'm a trigger
+        {/snippet}
+        {#snippet content()}
+            This shows a tooltip
+        {/snippet}
+    </Tooltip>
+```
+-->
+<Tooltip.Provider>
```
luca-schlecker, line 45: Is there a reason we don't wrap the whole application in a single tooltip provider?

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -0,0 +1,85 @@
...
+</div>
+<main class="flex flex-col h-full w-full px-2 py-4 gap-5">
+    <div class="flex flex-row w-full h-full gap-5">
+        <PaperDetailsCard />
+        <PaperResearchContextCard />
+    </div>
+    {#if showButtonBar}
+        <div class="flex flex-row w-full h-fit justify-between gap-4" data-testid="button-bar">
+            <!-- TODO: Implementation of navigation buttons will be done in #46 and #47 -->
+            <PaperNavigationButton direction="left" href="" />
```
luca-schlecker, line 70: Maybe adding a tooltip property to the button primitive would be a nice addition. `PaperNavigationButton` adds a whole new component containing primitive logic just for it to be used two times.
What do you think about 'inlining' it using that modified button primitive?

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -0,0 +1,85 @@
...
+            {#if userConfig.isReviewMode}
+                <!-- flex grow is very high so that it grows first, before the navigation buttons do -->
+                <!-- max-width is max-width of buttons + gap, which is the reason why they have fixed values -->
+                <div class="flex flex-grow-1000 max-w-[62rem] gap-[1rem] justify-center">
+                    <DeclineButton paperId={paper.id} />
+                    {#if userConfig.showMaybeButton}
+                        <MaybeButton paperId={paper.id} />
+                    {/if}
+                    <AcceptButton paperId={paper.id} />
+                </div>
```
luca-schlecker, line 80: Same here as with the `PaperNavigationButton`. Inlining may be useful.

src/lib/components/composites/paper-components/paper-view/cards/PaperCard.svelte
```diff
@@ -0,0 +1,57 @@
...
+                {#each tabs as tab}
+                    <Tabs.Trigger
+                        value={tab.value}
+                        class="data-[state=active]:border-b-primary h-fit rounded-none border-b-2 border-b-transparent bg-transparent px-4 pb-2 transition-none"
+                    >
+                        {tab.label}
+                    </Tabs.Trigger>
+                {/each}
+            </Tabs.List>
+            <Card.Content class="p-5">
```
luca-schlecker, line 39: Why is `Tabs.Content` not used here?

src/lib/components/composites/paper-components/paper-view/cards/PaperCard.svelte
luca-schlecker, line 1: Would it be desirable to extract just the specially styled tab bar into a component that then could be used inside a card?

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
+    import PaperCard from "./PaperCard.svelte";
+    import PaperCardContent from "./PaperCardContent.svelte";
+
+    const tabs = [
+        { value: "1", label: "References/Citations" },
```
luca-schlecker: Maybe stick to forward- and backward-references?

luca-schlecker (changes requested):
Great work! There were some things though that stood out to me. Looking forward to hearing your thoughts on these points.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
Slartibartfass2: The Tooltip Trigger is a button itself, so there's no need to add it to the button primitive.
I think it's nice to hide the implementation details in this `PaperView` component, which is big enough, so that one can easily see which buttons are where without having to figure that out by their code and styling.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
Slartibartfass2: Same as with the `PaperNavigationButton`s, this would add a lot of duplicated code and make the component less readable.

src/lib/components/composites/PaperBookmarkButton.svelte
Slartibartfass2: Great suggestion, I like it.

src/lib/components/composites/paper-components/paper-view/cards/PaperCard.svelte
Slartibartfass2: `Tabs.Content` is used in `PaperCardContent` to switch between the different card contents.

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
Slartibartfass2: Changed it.

src/lib/components/composites/Tooltip.svelte
Slartibartfass2: I didn't think of that yet, in the usage description it was always at one place. I think in principle we could do that, but I don't really see the advantage of that. It would remove the advantage of being able to control e.g. the delay for a certain tooltip and would probably be just confusing for others to read.

src/lib/components/composites/paper-components/paper-view/cards/PaperCard.svelte
Slartibartfass2: We don't use it anywhere else, but I like the idea and moved it to a separate component.

Slartibartfass2 (commented):
Thanks for the review, I'll rebase when we're finished so that you see the changes I made.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
luca-schlecker: I can see why you'd like to keep it this way. Fine with me :+1: .

luca-schlecker (approved):
LGTM. :+1:

PR #95, opened 2024-11-19
Add search bar component

### LINKED ISSUE #79: [Feature]: Add Search Bar component

_No response_

Currently, we don't have any custom component for a search bar (used for searching shortcuts, papers, references, ...).

Add a component for the Search Bar.

Alternatively, we could add a style for a shadcn input component.

### Anything else?

_No response_

Merseleo:
Closes #79

## What I have made

- I added a search bar component, which allows users to search, whereas the search function is passed to the component
- I updated the *vite.config.ts* to ignore the routes and app.d.ts from coverage (tested by e2e tests)

## Checklist

  - ~~[ ] unit tests~~ (it is a component which has no units to test)
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (it is a component which can not be tested with e2e tests)


src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -0,0 +1,68 @@
...
+    interface Props {
+        placeholderText?: string;
+        onSearch: (searchText: string) => void;
+    }
+    const { placeholderText = "Search", onSearch }: Props = $props();
+
+    let searchInput: string = $state("");
+
+    // timeout in ms to wait after last key input before start search
+    const timeout: number = 500;
```
Slartibartfass2: If the unit is not inferred by type, I would add it in the name.
```suggestion
    const timeoutInMs: number = 500;
```

src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -0,0 +1,68 @@
...
+            if (timeoutId !== null) clearTimeout(timeoutId);
+
+            // Wait a certain amount of time and start search, if no new key is pressed
+            timeoutId = setTimeout(() => onSearch(searchInput), timeout);
+        }
+    };
+</script>
+
+<!--
+@component
```
Slartibartfass2, line 44: Great that you can add documentation for components like this. Didn't know that.

src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+It starts the search on the following events:
+  - the user pressed enter
+  - the user click the search icon
+  - the user enters the search text and wait
+If the user enters the 'Esc' key, the search is aborted.
+
+Usage:
+```tsx
+    <SearchBar placeholderText={yourText} onSearch={yourCallback    } />
```
Slartibartfass2: Let's keep the description in one tense and use the svelte highlighting (tsx is for react + typescript).
```suggestion
It starts the search on the following events:
  - the user presses enter
  - the user clicks the search icon
  - the user enters the search text and waits
If the user enters the 'Esc' key, the search is aborted.

Usage:
```svelte
    <SearchBar placeholderText={yourText} onSearch={yourCallback} />
```

src/routes/+page.svelte
```diff
@@ -1,10 +1,20 @@
...
 </script>
 
-<SimpleNavigationBar {user} title="SnowballR" tabs={[]} defaultTabValue="" />
+<div class="flex flex-col gap-5">
+    <!-- example ussage of the navigation bar -->
+    <SimpleNavigationBar {user} title="SnowballR" tabs={[]} defaultTabValue="" />
+
+    <!-- example usage of the search bar ... -->
+    <SearchBar placeholderText={searchPlaceholderText} onSearch={(input) => console.log(input)} />
+    <SearchBar onSearch={(input) => console.log(input)} />
```
Slartibartfass2: This is no example usage for the navigation bar, it is, it's actual usage.
Let's move the examples to the theme page or maybe another page called examples, to keep the demos but don't block the pages from being designed.

src/lib/components/primitives/input/input.svelte
```diff
@@ -0,0 +1,22 @@
...
+        value = $bindable(),
+        class: className,
+        ...restProps
+    }: WithElementRef<HTMLInputAttributes> = $props();
+</script>
+
+<input
+    bind:this={ref}
+    class={cn(
+        "border-input bg-background ring-offset-background placeholder:text-muted-foreground flex h-10 w-full rounded-md border px-3 py-2 text-default file:border-0 file:bg-transparent file:text-sm file:font-medium focus-visible:outline-none disabled:cursor-not-allowed disabled:opacity-50",
```
Slartibartfass2: Let's use the theme that we already defined.
```suggestion
        "border-input-border-slate bg-background ring-offset-background placeholder:text-placeholder flex h-fit w-full rounded-md border px-4 py-2 text-default file:border-0 file:bg-transparent file:text-sm file:font-medium focus-visible:outline-none disabled:cursor-not-allowed disabled:opacity-50",
```

I also noticed that the `text-placeholder style doesn't use the predefined color for it, can you change that:
```pcss
.text-placeholder {
    @apply text-base font-medium tracking-normal text-input-placeholder-slate italic;
}
```

src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -0,0 +1,68 @@
...
+```
+-->
+<div class="relative w-full">
+    <!-- inspired by https://github.com/shadcn-ui/ui/issues/1562 -->
+    <Input
+        type="text"
+        placeholder={placeholderText}
+        bind:value={searchInput}
+        onkeyup={processNewInput}
+        class="px-4 py-3 pr-10"
```
Slartibartfass2: Now that this style is already in input, you only have to make space for the search button.
```suggestion
        class="pr-10"
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { render, screen } from "@testing-library/svelte";
+import userEvent from "@testing-library/user-event";
+
+describe("SearchBarComponent", () => {
+    test("When all required props are provided, then the search bar does not throw any error", () => {
+        render(SearchBar, {
+            props: { onSearch: (input: string) => console.log(input) },
+        });
+        assert.doesNotThrow(() => screen.getByPlaceholderText("Search"));
+        expect(screen.getByPlaceholderText("Search")).toBeInTheDocument();
```
Slartibartfass2: These assertions do the same things, don't they?

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    test("When placeholder text prop is provided, then the search bar show the provided placeholder text and not the default text", () => {
```
Slartibartfass2:
```suggestion
    test("When placeholder text prop is provided, then the search bar shows the provided placeholder text and not the default text", () => {
```

test/integration/search-bar/search-bar.test.ts
Slartibartfass2, line 1: Let's keep the When ..., then ... style consistent in this file.

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+                placeholderText: "Search paper",
+                onSearch: (input: string) => console.log(input),
+            },
+        });
+
+        assert.throws(() => screen.getByPlaceholderText("Search"));
+        assert.doesNotThrow(() => screen.getByPlaceholderText("Search paper"));
+    });
+
+    test("When the user type a search string into the search bar and click the search icon, the search is conducted", async () => {
```
Slartibartfass2:
```suggestion
    test("When the user enters a search string into the search bar and click the search icon, then the search is conducted", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        render(SearchBar, {
+            props: { onSearch: (input: string) => (searchInput = input) },
+        });
+
+        await userEvent.type(screen.getByRole("textbox"), "Test");
+        await userEvent.click(screen.getByRole("button"));
+        assert.equal(searchInput, "Test");
+    });
+
+    test("When the user type a search string into the search bar and press enter, the search is conducted", async () => {
```
Slartibartfass2:
```suggestion
    test("When the user enters a search string into the search bar and press enter, then the search is conducted", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        render(SearchBar, {
+            props: { onSearch: (input: string) => (searchInput = input) },
+        });
+
+        await userEvent.type(screen.getByRole("textbox"), "Test");
+        await userEvent.type(screen.getByRole("textbox"), "{enter}");
+        assert.equal(searchInput, "Test");
+    });
+
+    test("When the user type a search string into the search bar waits, the search is conducted", async () => {
```
Slartibartfass2:
```suggestion
    test("When the user enters a search string into the search bar waits, then the search is conducted", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        await userEvent.type(screen.getByRole("textbox"), "Test");
+        assert.equal(searchInput, "start");
+
+        setTimeout(() => {
+            assert.equal(searchInput, "Test");
+        }, 500);
+    });
+
+    test("When the user type no search string into the search bar, no search is conducted", async () => {
```
Slartibartfass2:
```suggestion
    test("When the user doesn't enter a search string into the search bar, then no search is conducted", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        });
+
+        await userEvent.type(screen.getByRole("textbox"), "{enter}");
+        assert.equal(searchInput, "start");
+
+        await userEvent.click(screen.getByRole("button"));
+        assert.equal(searchInput, "start");
+    });
+
+    test("When the user escapes the search bar, no search is conducted and the search input cleared", async () => {
```
Slartibartfass2:
```suggestion
    test("When the user escapes the search bar, then no search is conducted and the search input cleared", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        let searchInput: string = "start";
+        render(SearchBar, {
+            props: { onSearch: (input: string) => (searchInput = input) },
+        });
+
+        await userEvent.type(screen.getByRole("textbox"), "Test");
+        await userEvent.type(screen.getByRole("textbox"), "{esc}");
+        assert.equal(searchInput, "start");
```
Slartibartfass2: According to coverage the case for escape isn't executed. Is this the correct format to simulate key events?

Slartibartfass2 (changes requested):
Great job, the search bar looks great and the logic looks reasonable. The documentation is as good as expected :)

Just some minor changes left.

I noticed that the `onSearch` method is triggered again, when the user navigates in the text with the arrow buttons. Maybe add a check whether the input has changed and only then trigger the method, except when the enter key is pressed again.

Slartibartfass2:
Minor side quest:
I would add `exclude: ["src/routes/**/+*.{svelte,ts}", "src/app.d.ts"],` to the `coverage` object in `vite.config.ts` to only include components and logic into our coverage reports. The pages are later tested by the e2e tests.

Somehow the branch coverage isn't shown, but this needs to be solved separately.

Slartibartfass2:
[This](https://testing-library.com/docs/svelte-testing-library/example#two-way-data-binding) could also be interesting for testing the search input.

Merseleo:
> Great job, the search bar looks great and the logic looks reasonable. The documentation is as good as expected :)
> 
> Just some minor changes left.
> 
> I noticed that the `onSearch` method is triggered again, when the user navigates in the text with the arrow buttons. Maybe add a check whether the input has changed and only then trigger the method, except when the enter key is pressed again.

Great catch, thank you. I fixed it.

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: Agree, I will change it (and keep it in mind for similar cases).

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: Yep, I doesn't work in Webstorm, but I thought we should add it for all VS Code users.

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: I changed it.

src/routes/+page.svelte
Merseleo: I create a separate page, where we can put our custom components, so the homepage doesn't contain them anymore.

src/lib/components/primitives/input/input.svelte
Merseleo: I adapted the .text-placeholder style as well as the input styling (matching the theme from Figma).

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: Changed.

test/integration/search-bar/search-bar.test.ts
Merseleo: I will try it ...

test/integration/search-bar/search-bar.test.ts
Merseleo: Good catch, I will delete one.

test/integration/search-bar/search-bar.test.ts
Merseleo: Changed.

test/integration/search-bar/search-bar.test.ts
Merseleo: Done.

test/integration/search-bar/search-bar.test.ts
Merseleo: Done.

test/integration/search-bar/search-bar.test.ts
Merseleo: Done.

test/integration/search-bar/search-bar.test.ts
Merseleo: Done.

test/integration/search-bar/search-bar.test.ts
Merseleo: Done.

test/integration/search-bar/search-bar.test.ts
Merseleo: I think there is indeed a problem, I will check it and try to solve it.

Merseleo:
> Minor side quest: I would add `exclude: ["src/routes/**/+*.{svelte,ts}", "src/app.d.ts"],` to the `coverage` object in `vite.config.ts` to only include components and logic into our coverage reports. The pages are later tested by the e2e tests.
> 
> Somehow the branch coverage isn't shown, but this needs to be solved separately.

I added it.

src/lib/components/composites/search-bar/SearchBar.svelte
Slartibartfass2: This change isn't applied yet.

src/lib/components/composites/search-bar/SearchBar.svelte
Slartibartfass2: There's still the gigantic whitespace after `yourCallback`.

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,91 @@
...
+                placeholderText: "Search paper",
+                onSearch: (input: string) => console.log(input),
+            },
+        });
+
+        assert.throws(() => screen.getByPlaceholderText("Search"));
+        assert.doesNotThrow(() => screen.getByPlaceholderText("Search paper"));
+    });
+
+    test("When the user enters a search string into the search bar and click the search icon, then the search is conducted", async () => {
```
Slartibartfass2: Missed this last time.
```suggestion
    test("When the user enters a search string into the search bar and clicks the search icon, then the search is conducted", async () => {
```

test/integration/search-bar/search-bar.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            props: { onSearch: (input: string) => (searchInput = input) },
+        });
+
+        await userEvent.type(screen.getByRole("textbox"), "Test");
+        assert.notEqual(searchInput, "Test");
+        await userEvent.click(screen.getByRole("button"));
+        assert.equal(searchInput, "Test");
+    });
+
+    test("When the user enters a search string into the search bar and press enter, then the search is conducted", async () => {
```
Slartibartfass2: Missed this last time.
```suggestion
    test("When the user enters a search string into the search bar and presses enter, then the search is conducted", async () => {
```

Slartibartfass2 (changes requested):
Nearly done!

I still have some questions regarding the behavior:
- Should we trigger the onSearch method again, when enter or the button is pressed if the input hasn't changed? Is this the refresh functionality?
- When there's input and I press escape, the onSearch method isn't triggered. I think this could be misleading, e.g. when filtering the papers in the project view I want to clear the filter when pressing escape

what's your opinion about those behaviors?

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: Sorry, accidentally restored this change.

test/integration/search-bar/search-bar.test.ts
Merseleo: As i ...

test/integration/search-bar/search-bar.test.ts
Merseleo: Fixed it.

test/integration/search-bar/search-bar.test.ts
Merseleo: The test should be corrected now.

Merseleo (commented):
Regarding your questions on the behavior:
1) I would interpret it as a "refresh" option, I mean isn't this common search functionality that pressing the search button again or pressing enter multiple times start the search again and refreshes the results (at least DuckDuckGo etc. have this behavior).
2) Ok, I thought clearing the input with escape has the meaning of aborting the search. Same here: if you look at the behavior of search engines for the escape character, they often loose focus and don't search (and also not clear the input, but this is another discussion). But if you want, we can ask @luca-schlecker or @araschke what they think the behavior should be?!

Slartibartfass2 (approved):

Slartibartfass2:
Merge after #102

PR #102, opened 2024-11-24
Added build job

### LINKED ISSUE #96: [Feature]: Add build CI job

_No response_

Currently, we don't have a build job in the CI. Uploading the build artifact may be helpful and will be required when finishing a release later on.

Add a CI build job, which uses the build command in the package.json and upload the created artifact.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #96

## What I have made

- added a build job that uploads the build artifact
- added a lighthouse job that uses the build artifact to analyze the pages
  - this a long-running job, I wouldn't make it required
  - maybe I would only let it run when something is merged to develop or main
  - in the future we could deploy the reports e.g. using GitHub Pages

## Checklist

- [ ] ~~I have updated the documentation accordingly and commented my code~~ (only CI jobs)
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ (only CI jobs)
- [ ] ~~(for reviewer) I have checked the implementation against the requirements~~ (only CI jobs)

### DIFF
--- a/.github/lighthouse.config.js
+++ b/.github/lighthouse.config.js
@@ -1,0 +1,13 @@
+export default {
+    extends: "lighthouse:default",
+    settings: {
+        formFactor: "desktop",
+        screenEmulation: {
+            mobile: false,
+            width: 1920,
+            height: 1080,
+            deviceScaleFactor: 1,
+            disabled: false,
+        },
+    },
+};
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,38 @@
+name: Build
+
+on:
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+
+env:
+    node_version: ${{ vars.NODE_VERSION }}
+
+jobs:
+    build:
+        name: Build
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+
+            - name: Setup Node.js
+              uses: actions/setup-node@v4
+              with:
+                  node-version: ${{ env.node_version }}
+
+            - name: Install dependencies
+              run: npm install
+
+            - name: Build application
+              run: npm run build
+
+            - name: Archive build artifacts
+              uses: actions/upload-artifact@v4
+              if: ${{ !cancelled() }}
+              with:
+                  name: build-artifact
+                  path: .svelte-kit/output/
+                  if-no-files-found: error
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,87 @@
...
+
+            - name: Archive build artifacts
+              uses: actions/upload-artifact@v4
+              if: ${{ !cancelled() }}
+              with:
+                  name: build-artifact
+                  path: .svelte-kit/output/
+                  if-no-files-found: error
+
+    lighthouse:
```
Merseleo: Maybe only run this job, [wenn the pull request was approved](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved), which should happen before a merge to develop / main, but no block the ci all the time.

.github/lighthouse.config.js
```diff
@@ -0,0 +1,13 @@
+export default {
+    extends: "lighthouse:default",
+    settings: {
+        formFactor: "desktop",
+        screenEmulation: {
+            mobile: false,
+            width: 1440,
+            height: 1024,
```
Merseleo: I know we used this resolution in Figma, but why don't we check for the [most common resolution](https://www.browserstack.com/guide/ideal-screen-sizes-for-responsive-design): 1920 x 1080

Merseleo (changes requested):
Imo it looks good for now, maybe we want to extend the assertions etc. for Lighthouse later, but let's start with these jobs. By the way a nice idea to build the frontend and check it with Lighthouse.

.github/lighthouse.config.js
Slartibartfass2: We can use this too, as long as it's a fixed size.

.github/workflows/build.yml
Slartibartfass2: This seems too late in my opinion. When the PR is approved, there's no reason to see the results of this job.

I discussed this quick with @luca-schlecker, and we think it's best to just do it when something was pushed to main or develop. The developer/reviewer can check the report manually on their own machine.

Merseleo (commented):
Ok then let's run the lighthouse job on a requested review.

Merseleo (commented):
I think we are done with it.

Merseleo (approved):
Seems to work now, so let's merge ...

PR #111, opened 2024-11-27
Add paper list entry component

### LINKED ISSUE #101: [Feature]: Add Paper Entry component

_No response_

Currently, we don't have any component for paper entries, which display the the id, title, author and review information

Add a custom component for paper entries displaying the id, title, author and review information.
Therefore, the _PaperInfo_ component can be reused (see #74).

_No response_

### Anything else?

The component should look like the following:

Bildschirmfoto 2024-11-22 um 14 13 45

Merseleo:
Closes #101 

## What I have made

- added a component for displaying the basic paper information and optionally review information about this paper
- added event handlers to this component to navigate to the corresponding paper view on double click and executing a custom function on single click (or navigate to the paper view, if no other functionality is desired)
- added a new npm command for running the tests and opening the coverage: `npm run show-coverage`
- outsource avatar from user menu to own component, that optionally can display the review decision of the corresponding user

## Checklist

  - ~~[ ] unit tests~~ (this PR adds a new component, which have no testable units)
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (this PR adds a new component, which cannot be tested e2e)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,12 +11,13 @@
         "format": "prettier --write .",
         "lint": "prettier --check . && eslint .",
         "test": "npm run test:unit && npm run test:integration && npm run test:e2e",
-        "show-coverage": "npm run test && cd coverage/ && open index.html",
+        "show-coverage": "open coverage/index.html",
         "test:unit": "vitest run --project unit",
         "test:unit:ui": "npm run test:unit -- --ui",
         "test:integration": "vitest run --project integration",
         "test:integration:ui": "npm run test:integration -- --ui",
-        "test:e2e": "playwright test"
+        "test:e2e": "playwright test",
+        "test:and-open-coverage": "npm run test && npm run show-coverage"
     },
     "devDependencies": {
         "@playwright/test": "^1.49.0",
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
         "format": "prettier --write .",
         "lint": "prettier --check . && eslint .",
         "test": "npm run test:unit && npm run test:integration && npm run test:e2e",
+        "show-coverage": "npm run test && cd coverage/ && open index.html",
         "test:unit": "vitest run --project unit",
         "test:unit:ui": "npm run test:unit -- --ui",
         "test:integration": "vitest run --project integration",
--- a/src/lib/components/composites/navigation-bar/UserMenu.svelte
+++ b/src/lib/components/composites/navigation-bar/UserMenu.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-    import * as Avatar from "$lib/components/primitives/avatar/index.js";
     import * as DropdownMenu from "$lib/components/primitives/dropdown-menu/index.js";
     import BookOpen from "lucide-svelte/icons/book-open";
     import Archive from "lucide-svelte/icons/archive";
@@ -9,14 +8,13 @@
     import type { ComponentType } from "svelte";
     import type { Icon } from "lucide-svelte";
     import type { User } from "$lib/model/backend";
+    import UserAvatar from "$lib/components/composites/user-avatar/UserAvatar.svelte";
 
     interface Props {
         user: User;
     }
     const { user }: Props = $props();
 
-    const getInitial = (text: string) => (text.length > 0 ? text[0].toUpperCase() : "");
-    const userInitials = `${getInitial(user.firstName)}${getInitial(user.lastName)}`;
     const menuItems: {
         // The icon library still uses the deprecated ComponentType type
         icon: ComponentType<Icon>;
@@ -52,11 +50,8 @@
 </script>
 
 <DropdownMenu.Root>
-    <DropdownMenu.Trigger>
-        <Avatar.Root>
-            <Avatar.Fallback>{userInitials}</Avatar.Fallback>
-        </Avatar.Root>
-    </DropdownMenu.Trigger>
+    <UserAvatar {user} />
+    <DropdownMenu.Trigger></DropdownMenu.Trigger>
     <DropdownMenu.Content class="w-60" side="bottom" sideOffset={0} align="start">
         <DropdownMenu.Group>
             <DropdownMenu.GroupHeading>
--- a/src/lib/components/composites/navigation-bar/UserMenu.svelte
+++ b/src/lib/components/composites/navigation-bar/UserMenu.svelte
@@ -50,8 +50,9 @@
 </script>
 
 <DropdownMenu.Root>
-    <UserAvatar {user} />
-    <DropdownMenu.Trigger></DropdownMenu.Trigger>
+    <DropdownMenu.Trigger>
+        <UserAvatar {user} />
+    </DropdownMenu.Trigger>
     <DropdownMenu.Content class="w-60" side="bottom" sideOffset={0} align="start">
         <DropdownMenu.Group>
             <DropdownMenu.GroupHeading>
--- a/src/lib/components/composites/paper-components/PaperEntry.svelte
+++ b/src/lib/components/composites/paper-components/PaperEntry.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+    import PaperInfo from "$lib/components/composites/paper-components/PaperInfo.svelte";
+    import type { Paper } from "../../../../app";
+    import { ReviewStatus } from "../../../../types";
+
+    interface PaperEntryProps {
+        paper: Paper;
+    }
+
+    const { paper }: PaperEntryProps = $props();
+
+    const reviewStatusColor: Record<ReviewStatus, string> = {
+        [ReviewStatus.Unreviewed]: "border-unreviewed-gray",
+        [ReviewStatus.Accepted]: "border-accept-green",
+        [ReviewStatus.Maybe]: "border-maybe-yellow",
+        [ReviewStatus.Declined]: "border-decline-red",
+    };
+
+    function showStatus(): boolean {
+        return paper.showReviewStatus && paper.reviewData !== undefined;
+    }
+</script>
+
+<div
+    class="w-full border {showStatus() ? 'border-l-0' : ''} border-container-border-grey rounded-md"
+>
+    <div
+        class="{showStatus()
+            ? `border-l-4 ${reviewStatusColor[paper.reviewData.finalDecision]}`
+            : ''} rounded-md px-3 py-1.5"
+    >
+        <PaperInfo {paper} />
+    </div>
+</div>
--- a/src/lib/components/composites/paper-components/PaperEntry.svelte
+++ b/src/lib/components/composites/paper-components/PaperEntry.svelte
@@ -1,32 +1,30 @@
 <script lang="ts">
     import PaperInfo from "$lib/components/composites/paper-components/PaperInfo.svelte";
-    import type { Paper } from "../../../../app";
-    import { ReviewStatus } from "../../../../types";
+    import { type Paper, ReviewDecision } from "$lib/model/backend";
 
     interface PaperEntryProps {
         paper: Paper;
+        showReviewStatus: boolean;
     }
 
-    const { paper }: PaperEntryProps = $props();
+    const { paper, showReviewStatus = false }: PaperEntryProps = $props();
 
-    const reviewStatusColor: Record<ReviewStatus, string> = {
-        [ReviewStatus.Unreviewed]: "border-unreviewed-gray",
-        [ReviewStatus.Accepted]: "border-accept-green",
-        [ReviewStatus.Maybe]: "border-maybe-yellow",
-        [ReviewStatus.Declined]: "border-decline-red",
+    const reviewDecisionColor: Record<ReviewDecision | "unreviewed", string> = {
+        [ReviewDecision.Accepted]: "border-accept-green",
+        [ReviewDecision.Maybe]: "border-maybe-yellow",
+        [ReviewDecision.Declined]: "border-decline-red",
+        ["unreviewed"]: "border-unreviewed-gray",
     };
-
-    function showStatus(): boolean {
-        return paper.showReviewStatus && paper.reviewData !== undefined;
-    }
 </script>
 
 <div
-    class="w-full border {showStatus() ? 'border-l-0' : ''} border-container-border-grey rounded-md"
+    class="w-full border {showReviewStatus
+        ? 'border-l-0'
+        : ''} border-container-border-grey rounded-md"
 >
     <div
-        class="{showStatus()
-            ? `border-l-4 ${reviewStatusColor[paper.reviewData.finalDecision]}`
+        class="{showReviewStatus
+            ? `border-l-4 ${reviewDecisionColor[paper.reviewData?.finalDecision ?? 'unreviewed']}`
             : ''} rounded-md px-3 py-1.5"
     >
         <PaperInfo {paper} />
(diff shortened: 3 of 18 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
         "format": "prettier --write .",
         "lint": "prettier --check . && eslint .",
         "test": "npm run test:unit && npm run test:integration && npm run test:e2e",
+        "show-coverage": "npm run test && cd coverage/ && open index.html",
```
Slartibartfass2: How about:
```suggestion
        "show-coverage": "open coverage/index.html",
```
I often run only the integration test and then want to open the coverage, so I think we should have that separated. We could also add a command that combines both, e.g. `test:show-coverage`

src/lib/components/composites/navigation-bar/UserMenu.svelte
```diff
@@ -52,11 +50,8 @@
 </script>
 
 <DropdownMenu.Root>
-    <DropdownMenu.Trigger>
-        <Avatar.Root>
-            <Avatar.Fallback>{userInitials}</Avatar.Fallback>
-        </Avatar.Root>
-    </DropdownMenu.Trigger>
+    <UserAvatar {user} />
+    <DropdownMenu.Trigger></DropdownMenu.Trigger>
```
Slartibartfass2: The avatar has to be in the trigger. Otherwise, it's not clickable.
```suggestion
    <DropdownMenu.Trigger>
        <UserAvatar {user} />
    </DropdownMenu.Trigger>
```

src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -0,0 +1,62 @@
+<script lang="ts">
+    import * as Avatar from "$lib/components/primitives/avatar/index.js";
+    import { ReviewDecision, type User } from "$lib/types";
+    import Check from "lucide-svelte/icons/check";
+    import { X } from "lucide-svelte";
```
Slartibartfass2: I am not sure if this makes a difference:
```suggestion
    import X from "lucide-svelte/icons/x";
```

src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -0,0 +1,62 @@
...
+        >
+            <X size={16} color="#ffffff" strokeWidth="3" />
+        </div>
+    {:else if reviewDecision === ReviewDecision.Maybe}
+        <div
+            class="absolute top-6 left-6 flex justify-center items-center rounded-full bg-maybe-yellow h-5 w-5"
+        >
+            <text class="text-white">?</text>
+        </div>
+    {/if}
```
Slartibartfass2: Do you think there's a way to modularize this a bit more? The divs all have the same style except the color.
I think this is optional, I once tried using some sort of config for the decision buttons but ended up in creating a component for each of them.

tests/integration/navigation-bar/paper-navigation-bar.test.ts
```diff
@@ -62,6 +62,7 @@ describe("PaperNavigationBar", () => {
                 },
                 backRef: "/",
                 paper: {
+                    id: undefined,
```
Slartibartfass2: Was there a problem, when omitting the ID?

src/routes/+page.svelte
```diff
@@ -8,4 +8,6 @@
 <svelte:head>
     <title>SnowballR</title>
 </svelte:head>
-<SimpleNavigationBar {user} title="SnowballR" tabs={[]} defaultTabValue="" />
+<div class="flex flex-col gap-3">
+    <SimpleNavigationBar {user} title="SnowballR" tabs={[]} defaultTabValue="" />
+</div>
```
Slartibartfass2: What was the reason for this change?

src/lib/components/composites/paper-components/PaperEntry.svelte
```diff
@@ -0,0 +1,91 @@
...
+
+    // Mapping of review status to border color of paper entry
+    const reviewStatusColor: Record<ReviewStatus, string> = {
+        [ReviewProgress.Unreviewed]: "border-unreviewed-gray",
+        [ReviewDecision.Accepted]: "border-accept-green",
+        [ReviewDecision.Maybe]: "border-maybe-yellow",
+        [ReviewDecision.Declined]: "border-decline-red",
+    };
+
+    const showStatus = () => showReviewStatus && paper.reviewData !== undefined;
```
Slartibartfass2: Why does this have to be a method and not just a boolean value?
```suggestion
    const isStatusVisible = showReviewStatus && paper.reviewData !== undefined;
```

src/lib/components/composites/paper-components/PaperEntry.svelte
```diff
@@ -0,0 +1,91 @@
...
+     */
+    const handleClick = (e: MouseEvent) => {
+        if (e.detail === 1) {
+            if (timeoutId !== null) clearTimeout(timeoutId);
+            timeoutId = setTimeout(() => onClick(), 250);
+        } else if (e.detail === 2) {
+            if (timeoutId !== null) clearTimeout(timeoutId);
+            navigateToPaperView();
+        }
+    };
```
Slartibartfass2: This method looks a bit confusing at first. I think we can make it more self-describing.
```suggestion
    const handleClick = (event: MouseEvent) => {
        if (timeoutId !== null) {
            clearTimeout(timeoutId);
        }
        const numberOfClicks = event.detail;

        if (numberOfClicks === 1) {
            timeoutId = setTimeout(onClick, 250);
        } else if (numberOfClicks === 2) {
            navigateToPaperView();
        }
    };
```
- renamed `e` to `event`
- moved `timeoutId != null` check to the beginning
  - check is made for single and double click
  - else case isn't reached when we navigate to the paper view after a double click
- introduced new variable `numberOfClick` to make the meaning of `event.detail` more clear
- `() => onClick()` can be just `onClick` (the former just wraps `onClick` in another method)

What are your thoughts on these changes?

src/lib/components/composites/paper-components/PaperEntry.svelte
```diff
@@ -0,0 +1,91 @@
...
+Usage:
+```svelte
+    <PaperEntry paper={paper} projectId={1} showReviewStatus={true} />
+```
+-->
+<button
+    type="button"
+    class="flex flex-row items-center justify-end pe-4 gap-3 w-full border {showStatus()
+        ? 'border-l-0'
+        : ''} border-container-border-grey rounded-md hover:bg-container-border-grey/35"
```
Slartibartfass2: Optionally, you could use the `cn` utils method here to add additional style.
There's a similar case below.
```suggestion
    class={cn(
        "flex flex-row items-center justify-end pe-4 gap-3 w-full border border-container-border-grey rounded-md hover:bg-container-border-grey/35",
        isStatusVisible ? "border-l-0" : "",
    )}
```

src/lib/components/composites/paper-components/PaperEntry.svelte
```diff
@@ -0,0 +1,91 @@
...
+>
+    <div
+        class="flex flex-auto {showStatus()
+            ? `border-l-4 ${reviewStatusColor[paper.reviewData?.reviewStatus ?? ReviewProgress.Unreviewed]}`
+            : ''} rounded-md px-3 py-1.5"
+    >
+        <PaperInfo {paper} />
+    </div>
+    {#if showStatus()}
+        {#each paper.reviewData!.individualDecisions as [user, individualDecision]}
```
Slartibartfass2: The `!` is not necessary here because TypeScript checks that `reviewData` is defined.
Alternatively, you can remove the `paper.reviewData !== undefined` check and use `paper.reviewData?.individualDecisions ?? []`.

Afaik, `!` is a bad practice because you fool the compiler to believe that something is defined.

Slartibartfass2 (changes requested):
The component looks great, nice job!

I still have two questions:
1. Should we increase the double click cooldown? There's no standardization, so this leaves us to decide what's best. [Microsoft](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setdoubleclicktime#parameters) uses 500ms as default and I think that's reasonable, but I'm open to discussion.
2. When the component is hovered, there's little contrast between the hovered component and the avatar, do you think we should change this?

package.json
Merseleo: I think your second suggestion sounds like a good idea, I will add this command and adapt the first one.

src/lib/components/composites/navigation-bar/UserMenu.svelte
Merseleo: Oh my fault, thank you for the hint.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: I don't think so, but it looks better, so I will change it.

src/routes/+page.svelte
Merseleo: Because first I tested the components on this page and therefore wanted this layout, but if I am not mistaken, this will be changed after the rebase anyway?

tests/integration/navigation-bar/paper-navigation-bar.test.ts
Merseleo: No, if I remember correctly this was added while the rebase or so and I wanted to delete it, but forgot it, I will change it.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: I try to simplify it a bit, but I think an own component is not necessary as it is really simple and especially the icons change, so the only thing in common is the style.

src/lib/components/composites/paper-components/PaperEntry.svelte
Merseleo: It doesn't have to, I changed it.

src/lib/components/composites/paper-components/PaperEntry.svelte
Merseleo: I think your code is more elegant and I would change the timeout to 500ms, as you proposed (and Microsoft uses).
250ms was only a first try to make it faster, but I agree that the interval between the two clicks for a double click could be too short.

src/lib/components/composites/paper-components/PaperEntry.svelte
Merseleo: Theoretically I could and used It at the beginning, but personally I like the the current style more. 
But if you want to make it consistent, I can change it?

src/lib/components/composites/paper-components/PaperEntry.svelte
Merseleo: As showStatus() was a method, I got an error or at least a hint, that paper.reviewData could be null so I needed to add this check (or I was lost). But as the check is made via a variable now, it is completely fine to drop the ! and stick with best practice and not the bad practice.

Merseleo (commented):
Thank you for this great review, some very good catches and nice feedback.

Regarding your second question: I agree, that it is not the best solution, but I wanted to wait for your feedback first and see if you had any ideas. However, I will add my idea and you can give feedback on it.

Hopefully, I didn't forgot anything.

P.S. I will rebase, after we completed this review and will re-request a review from you, but I think there will not be any major changes.

src/routes/+page.svelte
Slartibartfass2: I'm not aware of any changes. In my PR of the Paper View (#89) I only added styling to the global container in the +layout.svelte.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Slartibartfass2: I think this is fine, but also add `h-5 w-5` to the style, which is the same for each div.

src/lib/components/composites/paper-components/PaperEntry.svelte
Slartibartfass2: I took a second look at this today and found some weird race conditions. If you click a second time, just when the timer stops, both events are fired. This is easy to achieve with a high cooldown, like 1 second.

I think this is due to the different interpretation of `numberOfClicks` and that the `timeoutId` isn't reset.

I investigated this for a while and found this, to be a sound solution:
- if the timeout wasn't set yet, it's the first click
  - start the timeout
  - when the timeout finishes, the `timeoutId` will be reset and the single click event fired
- when the timeout was set, it's the second click
  - clear the running timeout
  - reset `timoutId`
  - fire double click event
```ts
    const handleClick = () => {
        if (timeoutId === null) {
            timeoutId = setTimeout(() => {
                timeoutId = null;
                onClick();
            }, 350);
        } else {
            clearTimeout(timeoutId);
            timeoutId = null;
            navigateToPaperView();
        }
    };
```

What are your thoughts on this? Could you reproduce the behavior, which I described?

Also, on a second thought, 500 ms is too much, the single click has a high delay. We could do maybe 350ms?

Lastly, don't forget to adjust the documentation, there's a different amount of delay in it.

src/lib/components/composites/paper-components/PaperEntry.svelte
Slartibartfass2: No, that's up to you.

src/lib/components/composites/paper-components/PaperEntry.svelte
Slartibartfass2: great!

src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -0,0 +1,64 @@
...
+        firstName: "Max",
+        lastName: "Mustermann",
+    }} reviewDecision: ReviewDecision.Accept />
+```
+-->
+<div class="relative">
+    <Avatar.Root>
+        <Avatar.Fallback class="group-hover/paper-list-entry:bg-white"
+            >{userInitials}</Avatar.Fallback
+        >
```
Slartibartfass2: Minor, but this weirds me out 😅 
```suggestion
        <Avatar.Fallback class="group-hover/paper-list-entry:bg-white">
            {userInitials}
        </Avatar.Fallback>
```

Slartibartfass2 (changes requested):
This is nearly done! Great job.

Slartibartfass2:
To your hover style change:
What do you think about `bg-slate-200`, this makes everything a bit darker and keeps the contrast.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: Me too, I changed it.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: Sorry, missed it.

src/routes/+page.svelte
Merseleo: Ah, you are right. The styling is in the +layout.svelte. I will delete it.

src/lib/components/composites/paper-components/PaperEntry.svelte
Merseleo: I am also fine with 350ms, I also experimented with the delay an 250ms was the best for me, but I understand that it could be too fast for some people and we should increase it. Thus, i will change it (and update the documentation ;))

Merseleo (commented):
Regarding your proposal to change the background of the user avatar to `bg-slate-200` on hover looks better, thank you!

Slartibartfass2 (changes requested):
Let's rebase and finish this!

Slartibartfass2 (approved):

PR #112, opened 2024-11-28
Added Sign Up page

### LINKED ISSUE #2: [Use Case]: Sign Up

User

### Describe the use case

As user, I want so sign up, if I don't have an account already.

### Acceptance criteria

- show sign up screen
- when the user enters valid credentials, they should be signed up, signed in and redirected to the page they previously tried to access
- the user can enter a first and a last name, an email and a password
- the first and last name have to be non-blank and have a maximum length
- the password is non-blank, needs to be at least 8 characters long, has to contain some special characters and at least one number and has a maximum length

### Anything else?

Related to #1, #4 and #58

grafik

Slartibartfass2:
Closes #2 

## What I have made

**Note**: Due to the Backend not being able to handle user creation, this doesn't work with the deployed server until it's fixed. For that, I created the following issues:
- Backend: https://github.com/SE-UUlm/snowballr-backend/issues/267
- Frontend: https://github.com/SE-UUlm/snowballr-frontend/issues/134

<br />

- added schemas to validate user input in `src/lib/schemas.ts`
  - we use the zod library to create schemas
  - the first and last name has to consist of 1-100 non-whitespace characters
  - the email must be a valid email format (uses predefined email schema)
  - the password has to consist of 8–128 characters including 2 lower and 2 upper case letters, 2 numbers and 2 special characters, as well as it can only use a fixed set of characters not including whitespace
  - zod has a fixed set of issue types and if you define a rule on your own it's always `custom`. To differentiate different custom rules, we add a `subCode` parameter that can be read after the validation was processed
  - those schemas can not only be used for the sign-up and the sign-in page, but also for the new-password page and for changing the username in the settings
- I moved the `newpassword`, `resetpassword`, `signin` and `signup` route to the `auth` group to apply a layout that consists of a window that is always centered, as it is for all these pages
- added a highly customizable input component
  - this component tries to abstract as much functionality as possibly to reduce code duplication when using input components
  - when a schema is passed, the user can validate the input by binding the component and calling the `validate` method
  - the `getValue` method has the purpose to not have to also bind the value next to the component ref
- added `PasswordInput` as a reusable password input component
  - it handles toggling the visibility of the input
  - it applies the password schema
- added sign-up page
  - I tried the `Form` component at first, but it didn't work out for me, so I added the logic myself using the `Input` component
  - If the submit button is pressed or the enter key is pressed while focusing an input component, the inputs are validated and the sign-up API is called
- I spent probably too much time in over optimizing the schema test suite, but now it's straightforward to use
  - see `schemaTest` documentation for more details

## Checklist

  - [x] unit tests
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (not able to do that currently)


.github/workflows/build.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 env:
     node_version: ${{ vars.NODE_VERSION }}
+    PUBLIC_API_BASE_URL: http://localhost:8080
```
Merseleo: I also noticed, that this env variable was missing in my PR. A general suggestion, as manually setting the URL in multiple files could cause problems, is to save the concrete URL as [Github Secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow)?

.github/workflows/lighthouse.yml
```diff
@@ -37,10 +37,11 @@ jobs:
                       const globber = await glob.create('**/src/routes/**/+page.svelte');
                       const files = await globber.glob();
                       const urls = files
-                          .map(x => x.match(/routes\/([a-zA-Z\/\[\]]*?)\/?\+page\.svelte/)[1])
-                          .map(x => x.replace('[paperId]', '42').replace('[projectId]', '12'))
+                          .map(x => x.match(/routes\/([\(a-zA-Z\/\[\]\)]*?)\/?\+page\.svelte/)[1])
+                          .map(x => x.replace('[paperId]', '42').replace('[projectId]', '12').replace(/\([a-z\-]+\)\//g, ''))
```
Merseleo, line 41: Why do you "delete" all the nested routes for the project?

src/lib/controller/backend-controller.ts
```diff
@@ -64,9 +64,17 @@ export class BackendController implements IBackendController {
...
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     async createUser(personalInfo: UserSpec, password: string): Promise<User> {
-        throw new Error("Method not implemented.");
+        const payload = {
+            email: personalInfo.email,
+            sender: {
+                firstName: personalInfo.firstName,
+                lastName: personalInfo.lastName,
+                password: password,
+            },
```
Merseleo, line 74: Why do you want to encapsulate this data here? Currently the new API definition does not nest any data for a user? So is there any special reason?

src/routes/+layout.svelte
Merseleo, line 1: Maybe you could add an automatic redirect to the signup page (oder signin in the other PR), so we start there?!

src/lib/components/composites/input/Input.svelte
```diff
@@ -0,0 +1,139 @@
...
+        placeholder = "",
+        required,
+        type,
+        link,
+        inputClass = "",
+        schema,
+        onButtonClick,
+        buttonContent,
+        buttonProps,
+        value = $bindable(),
```
Merseleo, line 45: Maybe rename to inputValue or is this something different?

src/lib/components/composites/input/Input.svelte
```diff
@@ -0,0 +1,139 @@
...
+
+    let errorMessages = $state<string[]>([]);
+    let isFirstValidation = $state(true);
+
+    /**
+     * Validates the input value against the provided schema and returns whether the input was valid.
+     *
+     * - If no schema is provided, this method always returns true
+     *
+     * @returns {boolean} True if the input value is valid, false otherwise.
```
Merseleo:
```suggestion
     * @returns {boolean} True if the input value conforms to the schema, false otherwise.
```

src/lib/components/composites/input/Input.svelte
```diff
@@ -0,0 +1,139 @@
...
+        value = $bindable(),
+        class: className,
+        ...restProps
+    }: Props = $props();
+
+    let errorMessages = $state<string[]>([]);
+    let isFirstValidation = $state(true);
+
+    /**
+     * Validates the input value against the provided schema and returns whether the input was valid.
```
Merseleo:
```suggestion
     * Validates the input value against the provided schema.
```

src/lib/components/composites/input/Input.svelte
```diff
@@ -0,0 +1,139 @@
...
+     * Same as binding the `value` prop.
+     */
+    export const getValue = () => value;
+
+    /**
+     * After the initial input was validated, validate the input on every input event.
+     */
+    function onInput() {
+        if (!isFirstValidation) {
+            validate();
```
Merseleo: Why is it not possible to call the validate() function directly, so why do you distinguish between first validation and other validation?

src/lib/components/composites/input/Input.svelte
```diff
@@ -0,0 +1,139 @@
...
+        />
+        {#if buttonContent}
+            <button class="absolute right-4" onclick={onButtonClick} type="button" {...buttonProps}>
+                {@render buttonContent()}
+            </button>
+        {/if}
+    </div>
+    {#each errorMessages as errorMessage}
+        <p class="text-sm text-red-500">{errorMessage}</p>
+    {/each}
```
Merseleo: A suggestion: Do not only display the errors in red text, but two columned (depending on width) with Red Cross icons (like declined paper) with shorter text. I've got inspired by https://dribbble.com/shots/20671956-Stratis-UI-Login-modals for example. What do you think of it?

src/lib/components/composites/input/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
...
+Customized {@link Input} component for password input.
+
+Usage:
+```svelte
+    <PasswordInput bind:value={password} bind:this={passwordInput} showForgotPasswordLink />
+```
+-->
+<Input
+    class={className}
+    inputId="password"
```
Merseleo:
```suggestion
    inputId="password-input"
```

src/lib/components/composites/input/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+<!--
+@component
+Password Input Element.
+
+Customized {@link Input} component for password input.
+
+Usage:
+```svelte
+    <PasswordInput bind:value={password} bind:this={passwordInput} showForgotPasswordLink />
```
Merseleo: You have no props showForgotPasswordLink yet?!

src/lib/components/composites/input/Input.svelte
Merseleo: I know this is difficult if it handled in this component, but it would be cool if the error not disappears as it solved, but gets green with a tick or something like that.

Merseleo: And maybe change the border color to red (or green), if the validation fails (or succeeds)?

src/lib/schemas.ts
```diff
@@ -0,0 +1,124 @@
...
+                "Password must contain at least 2 numbers",
+            );
+        }
+
+        if (!hasMinNumberOfCharacterSet(password, specialCharacters, 2)) {
+            addCustomIssue(
+                context,
+                ZodIssueSubCode.not_enough_special_characters,
+                `Password must contain at least 2 special characters (${specialCharacters})`,
+            );
```
Merseleo: Maybe indicate these results by a password strength meter and do not block it directly?

So only check, that at least an uppercase letter, lowercase letter, number and special character is in it and indicate further password strength by the strength meter?

src/routes/(auth)/signup/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        const userSpec: UserSpec = {
+            firstName: firstNameInput.getValue(),
+            lastName: lastNameInput.getValue(),
+            email: emailInput.getValue(),
+        };
+        const user = await BackendController.getInstance().createUser(
+            userSpec,
+            passwordInput.getValue(),
```
Merseleo, line 33: Is it right to send the initial password in plain text?

Merseleo: And how do I test this properly? Not possible yet, right?

tests/integration/input/input.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        // When input is validated, then error messages are shown
+        expect(input.component.validate()).toBe(false);
+        await waitFor(() => {
+            errorMessages = document.getElementsByTagName("p");
+            expect(errorMessages).toHaveLength(1);
+        });
+        expect(errorMessages[0]).toHaveTextContent("String must contain at least 10 character(s)");
+    });
+
+    test("When a button is provided, then button is shown", () => {
```
Merseleo:
```suggestion
    test("When a button is provided, then button is shown and can be clicked.", () => {
```

tests/integration/input/password-input.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        // Error messages do not exist
+        let errorMessages = document.getElementsByTagName("p");
+        expect(errorMessages).toHaveLength(0);
+
+        // When input is validated, then no error messages are shown
+        expect(input.component.validate()).toBe(true);
+        errorMessages = document.getElementsByTagName("p");
+        expect(errorMessages).toHaveLength(0);
+        expect(input.component.getValue()).toBe(validPassword);
```
Merseleo:
```suggestion
```

Checked in test below, so code is a duplicate.

tests/integration/input/password-input.test.ts
```diff
@@ -0,0 +1,79 @@
...
+                value: validPassword,
+            },
+        });
+
+        // When input is validated, then no error messages are shown
+        expect(input.component.validate()).toBe(true);
+        const errorMessages = document.getElementsByTagName("p");
+        expect(errorMessages).toHaveLength(0);
+        expect(input.component.getValue()).toBe(validPassword);
+    });
```
Merseleo: Please add another test for the case, the password is not valid.

tests/unit/schemas.test.ts
Merseleo, line 1: Nice job, I agree over engineered, but very cool tests.

Merseleo (changes requested):
Puh, that was a lot of work from you, but greatly done. Some things we need to add later, as you already wrote in your description, but in general I think most of it is good.
However, there are some things I don't understand, maybe you explain some things to me.

Mainly, there are only some design discussions open ;)

I hope, I don't forgot anything in this PR, I don't want to rule out the possibility of further questions arising in the second round.

.github/workflows/lighthouse.yml
Slartibartfass2: This is only for groups. I updated the regex to also match routes in parentheses. Because groups are only to organize your pages but aren't part of the URL of a sub-page, we remove the group name from the path.

src/lib/controller/backend-controller.ts
Slartibartfass2: I wrote this, when we didn't have the new API. This is how it's done in the current backend.

src/lib/components/composites/input/Input.svelte
Slartibartfass2: My idea was, that it would bother the user to show them the validation errors before they even had the chance to input any credentials. So now the user can input their credentials and only in the case where their credentials didn't meet our requirements we show them the error messages.

.github/workflows/lighthouse.yml
Merseleo: Ah right, I forgot the groups.

src/lib/controller/backend-controller.ts
Merseleo: Ok, make sense. But probably you now understand, that I want to specify the API as fast as possible.

.github/workflows/build.yml
Slartibartfass2: Yes, that makes sense, I added it as an Action variable (not a secret because it's just a URL)

src/lib/components/composites/input/Input.svelte
Slartibartfass2: `value` is 'inherited' by `HTMLInputAttributes`. We could do something like
```ts
value: inputValue = $bindable(),
```

src/lib/controller/backend-controller.ts
Slartibartfass2: Yes, totally!

src/routes/(auth)/signup/+page.svelte
Slartibartfass2: AFAIK, this is the way to do it. I checked different websites and they do the same.
When the Webserver is deployed, it will use TLS to encrypt the connection.

We cannot test this properly yet due to the missing E2E setup and the wrong backend behavior.

src/lib/components/composites/input/PasswordInput.svelte
Slartibartfass2: Yes, I moved this to the sign in PR. I remove it here.

src/lib/components/composites/input/PasswordInput.svelte
Slartibartfass2: Changed it.

tests/integration/input/password-input.test.ts
Slartibartfass2: Must've forgotten this. I've added one.

src/lib/schemas.ts
Slartibartfass2: A password strength meter would be a massive implementation, don't you think?
There are many aspects that are used by such tool:
- dictionaries of words in a specific language
- dictionaries of common passwords
- substitutions of single characters like '4's and 'A's or '3's and 'E's
- sequences
- keyword proximity

In my opinion, this would be another PR with low priority because the current implementation works and assures a decent password.

src/lib/components/composites/input/Input.svelte
Slartibartfass2: I like the idea and added it to the current implementation. It was a lot, but I made it work.

Slartibartfass2 (commented):
Thank you a lot for the time you spent on this review. I'm excited to hear what you think of the new changes.

tests/unit/schemas.test.ts
Slartibartfass2: Thanks :))

src/routes/+layout.svelte
Slartibartfass2: I added this in the sign in PR because there I can check whether the user is signed in.

.github/workflows/build.yml
Merseleo: Ok, thank you.

src/routes/+layout.svelte
Merseleo: Ok 👍

src/lib/components/composites/input/Input.svelte
Merseleo: Ah ok, did not know that. Then it is probably wise to not rename it.

src/lib/components/composites/input/Input.svelte
Merseleo: Ok, understand. Did not consider the fact that it could be annoying to see the errors before any input, but as I rethought it, it makes sense indeed. So let's keep this behavior, but may I request a comment containing the reason you described to me, so that other developers understand it directly?

src/lib/components/composites/input/Input.svelte
Merseleo: I think you forgot to color the borders depending on the validation. Or is this too much?

src/lib/components/composites/input/PasswordInput.svelte
Merseleo: Ok, thank you.

src/lib/schemas.ts
Merseleo: I agree, I didn't consider all necessary changes at the beginning. But do you generally agree that I could be useful and want to create this Issue or do you think we should drop this idea completely?

src/routes/(auth)/signup/+page.svelte
Merseleo: Ok.

And i forgot that the connection will be TLS encrypted later, so the problem I thought exists, disappears.

src/lib/schemas.ts
```diff
@@ -0,0 +1,125 @@
...
+                ZodIssueSubCode.invalid_characters,
+                `only lower or upper case letters, numbers and the following special characters ${specialCharacters}`,
+            );
+        }
+
+        if (!hasMinNumberOfCharacterSet(password, upperCaseLetters, 2)) {
+            addCustomIssue(
+                context,
+                ZodIssueSubCode.not_enough_upper_case_letters,
+                "at least 2 upper case letter",
```
Merseleo:
```suggestion
                "at least 2 uppercase letters",
```

src/lib/schemas.ts
```diff
@@ -0,0 +1,125 @@
...
+            );
+        }
+
+        if (!hasMinNumberOfCharacterSet(password, lowerCaseLetters, 2)) {
+            addCustomIssue(
+                context,
+                ZodIssueSubCode.not_enough_lower_case_letters,
+                "at least 2 lower case letter",
```
Merseleo:
```suggestion
                "at least 2 lowercase letters",
```

Merseleo (changes requested):
I like the new validation visualization, but I think you missed some things. We are on the right track!

src/lib/components/composites/input/Input.svelte
Slartibartfass2: Yes, I strongly agree.

src/lib/schemas.ts
Slartibartfass2: I think it's a nice addition but with very low priority. There are numerous things we still have to do and I wouldn't like to invest any more time into this until we finished the essential UI and logic.

src/lib/components/composites/input/Input.svelte
Slartibartfass2: Yes I forgot that. I added it now.

Slartibartfass2 (commented):
Thank you again for reviewing this PR!

Slartibartfass2:
I'll rebase this, as soon as we're finished.

src/lib/components/composites/input/Input.svelte
Merseleo: I like the comment, makes this code more understandable imo.

src/lib/components/composites/input/Input.svelte
Merseleo: I like the new border 👍

src/lib/schemas.ts
Merseleo: Ok, understandable ... so let's "forget" this idea, as we will probably do not have the time (considering the progress, especially of the backend).

Merseleo (approved):
I think we are done, so if the rebase doesn't change the code massively, you can merge!

Issue #82, opened 2024-11-18
[Feature]: Add Backend API Interface


luca-schlecker:
### Is your feature request related to a problem? Please describe

There is no way to talk to the backend yet.

### Describe the solution you'd like

A centralized and uniform way of talking to the backend independent of URLs or transportation protocols. This decreases coupling and makes it easier to access the backend whilst also decreasing the potential for errors.

PR #113, opened 2024-11-28
draft: initial backend api interface

luca-schlecker:
The PR for #82 isn't yet in its final form, this only serves as an initial draft discussion opportunity.


src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+    createProject(spec: ProjectSpec): Promise<Project>;
+    project(projectId: number): IProjectController;
+
+    getUsers(): Promise<User[]>;
+    createUser(personalInfo: UserSpec, password: string): Promise<User>;
+    user(userId: number): IUserController;
+    thisUser(): IUserController;
+
+    getAuthors(): Promise<Author[]>;
+    createAuthor(spec: AuthorSpec): Promise<void>;
```
Slartibartfass2: I think it would be great to return the created `Author` here.
```suggestion
    createAuthor(spec: AuthorSpec): Promise<Author>;
```

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface IUserController {
+    get(): Promise<User>;
+    block(): Promise<void>;
+    changePassword(newPassword: string): Promise<void>;
+    update(newSpec: UserSpec): Promise<User>;
+
+    getProjects(): Promise<Project[]>;
+    getReadingList(): Promise<ReadingListEntry[]>;
+    addToReadingList(paperId: number): Promise<void>;
+    removeFromReadingList(readingListEntryId: number): Promise<void>;
```
Slartibartfass2: Currently, I see an issue with this, when I don't have the reading list entry ID.
The combination of user ID and paper ID should be enough to refer to a reading list entry.
In my opinion, the `removeFromReadingList` method should take the paper ID instead. This way, we don't have to store the reading list entry ID anywhere.

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+    get(): Promise<Project>;
+    archive(): Promise<void>;
+    getZip(): Promise<Blob>;
+    getCsv(): Promise<Blob>;
+
+    getStageCount(): number;
+    stage(stageIndex: number): IStageController;
+
+    getMembers(): Promise<User[]>;
+    inviteMember(email: string): Promise<void>;
```
Slartibartfass2: Can we rename that to `inviteUser`?

src/lib/backend/Model.ts
```diff
@@ -0,0 +1,98 @@
...
+export enum ReviewDecision {
+    No,
+    Maybe,
+    Yes,
+}
+
+export type PaperDecision = ReviewDecision.Yes | ReviewDecision.No;
+
+export interface ReviewDecisionMatrix {
+    reviewers: number;
```
Slartibartfass2: Is there a reason why this is not part of a project directly?
Also, what do you think about the name `numberOfReviewers`?

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface IStageEntryController {
+    get(): Promise<StageEntry>;
+    remove(): Promise<void>;
+
+    getReviews(): Promise<Review[]>;
+    review(userId: number): IReviewController;
+}
+
+export interface IStageController {
+    getCsv(): Promise<Blob>;
```
Slartibartfass2: Can we make `getCsv()` more descriptive? Without reading the OpenAPI spec, I wouldn't know that this returns the papers as csv including their review decision.

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    getPapers(): Promise<StageEntry[]>;
+    addPaper(paperId: number): Promise<void>;
```
Slartibartfass2: I think it would be great to return the created `StageEntry` here.
```suggestion
    addPaper(paperId: number): Promise<StageEntry>;
```

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+    getCsv(): Promise<Blob>;
+
+    getStageCount(): number;
+    stage(stageIndex: number): IStageController;
+
+    getMembers(): Promise<User[]>;
+    inviteMember(email: string): Promise<void>;
+    removeMember(userId: number): Promise<void>;
+
+    getCriterias(): Promise<Criterion[]>;
```
Slartibartfass2: The plural of criterion is criteria.
```suggestion
    getCriteria(): Promise<Criterion[]>;
```

src/lib/backend/Model.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export interface Project extends ProjectSpec {
+    id: number;
+}
+
+export interface ProjectSpec {
+    name: string;
+    reviewDecisionMatrix: ReviewDecisionMatrix;
+    mergeThreshold: number;
```
Slartibartfass2: In the figma design, we decided on `similarityThreshold` let's keep it consistent and use it here too.

src/lib/backend/Model.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export interface ReviewSpec {
+    finished: boolean;
+    decision: ReviewDecision;
+    selectedCriteriaIds: number[];
+}
+
+export interface StageEntry {
+    paperId: number;
```
Slartibartfass2: Is there a reason why the `paperId` is listed here again, while it's already in `Paper`.

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+    get(): Promise<Criterion>;
+    remove(): Promise<void>;
+    update(spec: CriterionSpec): Promise<void>;
+}
+
+export interface IProjectController {
+    get(): Promise<Project>;
+    archive(): Promise<void>;
+    getZip(): Promise<Blob>;
+    getCsv(): Promise<Blob>;
```
Slartibartfass2: Can we make `etCsv()` more descriptive? Without reading the OpenAPI spec, I wouldn't know that this returns the papers as csv.

src/lib/backend/Backend.ts
```diff
@@ -0,0 +1,111 @@
...
+export interface ICriterionController {
+    get(): Promise<Criterion>;
+    remove(): Promise<void>;
+    update(spec: CriterionSpec): Promise<void>;
+}
+
+export interface IProjectController {
+    get(): Promise<Project>;
+    archive(): Promise<void>;
+    getZip(): Promise<Blob>;
```
Slartibartfass2: Can we make `getZip()` more descriptive? Without reading the OpenAPI spec, I wouldn't know that this returns the replication package.

Slartibartfass2 (changes requested):
Currently, we use kebab-case for our file TypeScript file names, and at least for the svelte `+page.ts` files this should stay the same. What are your thoughts on naming those files `model.ts` and `backend-api.ts`.
Also, what are your thoughts on the following directory structure:
```
.
└── lib/
    ├── model/
    │   ├── backend/
    │   │   ├── paper.ts
    │   │   ├── project.ts
    │   │   └── ...
    │   └── internal/
    │       └── ...
    └── backend-api.ts
```

src/lib/backend/Model.ts
luca-schlecker: I am unsure how I'd like to implement the final data structure. This only serves the purpose of defining what will be sent in the end.

src/lib/backend/Model.ts
luca-schlecker: No

src/lib/backend/Model.ts
Slartibartfass2: What would be the problems when moving this to `Project`?

src/lib/model/backend.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export interface Paper {
+    id: number;
+    doi: string;
+    title: string;
+    abstrakt: string;
+    year: number;
+    type: string;
+    authorIds: number[];
+    backwardReferencedPaperIds: number[];
```
Slartibartfass2, line 75: For the full list of references, we would also need the forward referenced paper IDs, can we do that?

Slartibartfass2 (changes requested):
LGTM so far.

Let's merge this into develop so we can use the types in other PRs.
I've noticed that you resolved the conversations yourself, I'm used to it that the reviewer does this, to assure that the point of discussion was resolved.
Are you familiar with other workflows? Please elaborate.

Also, what were your thoughts on https://github.com/SE-UUlm/snowballr-frontend/pull/113#pullrequestreview-2472263099? Should we keep all types in one file or try to separate them?

luca-schlecker:
I resolved them myself as I implemented the requested change. It served to me as a sort of todo-list.
Will leave them open in the future. 👍

I preferred having them as a single file. Typescript is able to import just a specific type and I felt like neither the file was long enough nor the types complicated enough to split them apart.

src/lib/model/backend.ts
luca-schlecker: I initially intended forward references to be accessed solely through `getForwardReferencedPapers` of a paper as they are a direct computational result of backward references. On second thought it would be a nice API addition to just put them in a project. This would ease the addition of forward references without having to go to the referencing paper.

Will be adding this.

Slartibartfass2: Great!

Slartibartfass2:
Just one last thing: Can you remove the `types.ts` in lib? Now that we have your types, we don't need them anymore.

Slartibartfass2 (changes requested):
Just one last thing: Can you remove the `types.ts` in lib? Now that we have your types, we don't need them anymore.

src/lib/backend/Model.ts
Slartibartfass2: As discussed, the number of reviewers is relevant for the decision matrix, so we keep it there for now.

Slartibartfass2 (changes requested):
LGTM.

The check fails because the props in the tests are missing properties. As discussed, I recommend a creation method, e.g.

In `./tests/model-builder.ts`:
```ts
import type { User } from "$lib/model/backend";

function createUser(user: Partial<User>): User {
    const defaultUser: User = {
        id: 1,
        firstName: "John",
        lastName: "Doe",
        email: "john.doe@example.com",
        isAdmin: false,
        status: "active",
    };
    return Object.assign(defaultUser, user)
}
```

Slartibartfass2 (approved):
Great! Let's merge it.

PR #117, opened 2024-12-02
Moved types to types.ts file

### LINKED ISSUE #116: [Feature]: Move types to types.ts

_No response_

The types in app.d.ts don't get recognized correctly and no errors are thrown when the structure isn't met. This omits type safety.

Move types to a types.ts, where we can define internal types.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #116

## What I have made

- moved types to types.ts to enable type safety and a common location for our types

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (only refactoring)
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only refactoring)


src/lib/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface Paper extends PaperSpec {
+    readonly id: number;
+}
+
+export interface PaperSpec {
+    readonly title: string;
+    readonly authors: string[];
+    readonly isBookmarked: boolean;
+}
```
Merseleo, line 9: Why did you separate paper and paper specification? Is there any use case, where the paper id would have a negative impact or is not existing?

src/routes/archivedprojects/+page.svelte
```diff
@@ -1,10 +1,8 @@
...
 
-    const user = {
-        firstName: "Max",
-        lastName: "Mustermann",
-    };
+    const { data } = $props();
+    const { user } = data;
 </script>
 
 <SimpleNavigationBar {user} title="Archived Projects" backRef="/" tabs={[]} defaultTabValue="" />
```
Merseleo, line 8: Why do you explicitly create a variable for user here and in the routes/+page.svelte you directly access the user
with user={data.user}?!

Same on following pages ...

Merseleo (changes requested):
Only two questions, but in general a good starting point for the types, thank you!

src/lib/types.ts
Slartibartfass2: When a new paper is created, there's only the data of the paper but no ID. the paperspec can then be used for the creation of the paper. The same names are used in the backend types, which makes it easier to integrate later.

src/routes/archivedprojects/+page.svelte
Slartibartfass2: This is just convenience. When the name is the same as the argument, one can only use `{user}`.
We can expect more pattern matching later, i.e. `const { user, paper, ... } = data;`

Slartibartfass2 (commented):
Thank you for the review, I answered your comments.

src/lib/types.ts
Merseleo: Ah ok, makes sense (I forgot the case of creating a paper.

src/routes/archivedprojects/+page.svelte
Merseleo: Ok, then why did you not follow this pattern in the routes/+page.svelte file?

Slartibartfass2: I missed this, good catch, I changed it there.

Merseleo (approved):
Perfect, thank you for moving the types and adapting everything! Let's merge ;)

PR #120, opened 2024-12-02
Renamed test folder to 'tests'

### LINKED ISSUE #119: [Feature]: Rename test folder to tests

_No response_

the tsconfig expects the test folder to be named `tests`, otherwise path aliases are not recognized correctly, and the editor will show an error.

Rename the directory.

The tsconfig includes can be overwritten but I think renaming makes more sense.

### Anything else?

_No response_

Slartibartfass2:
Closes #119 

## What I have made

- removed demo component
- renamed test directory to 'tests'
- now the path aliases are recognized correctly
- this also enables type safety in the props in the integration tests, which is great
- I carefully checked if I missed any renaming, please check again

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (only renamed directory)
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only renamed directory)


Merseleo (approved):
I think you did not miss anything, looks perfect!

Issue #121, opened 2024-12-03
[Use Case]: View Open Reviews


Merseleo:
### Who is the primary actor in this use case?

User

### Describe the use case

As a user, I want to see a list of all open reviews, i. e. paper which are either unreviewed or undecided, across all SLRs, in which I collaborate.

### Acceptance criteria

- show open reviews list on the home page
- if a paper is either unreviewed or undecided, i. e. has the status 'Maybe', and is part of a SLR, in which the user collaborates, it is shown in the list, otherwise not

PR #122, opened 2024-12-04
Show projects list on homepage

### LINKED ISSUE #6: [Use Case]: View Project List

User

### Describe the use case

As user, I want to see a list of projects, in which I collaborate.

### Acceptance criteria

- show project list on the landing/home page
- if the user is a member of a project, the project should be shown in the list
- if the user is not a member of a project, the project shouldn't be shown in the list

### Anything else?

_No response_

Merseleo:
Closes #6

> **Hint**: For properly testing the new homepage, you need to either run the mock backend server (and checkout `feat/add-demo-projects`) or do load the data from the server, but statically in the load function (in +page.ts) from the homepage.

## What I have made

- added basic layout to the _Homepage_ (with Todo for "Open Reviews" list, see #121)
- added component for project list entry, which can navigate to the project dashboard on click and display project metadata
- added component for named list, which get an array of list items and display the list in a uniform layout
  - the list expects a component as child, in which a entry from the item list can be wrapped (as well as the corresponding skeleton component)
 - show list of active and archived projects on the _Homepage_  

- from now on the primitive components are not considered in test coverage anymore, although they would probably have 100% coverage when used in composite components

## Checklist

  - ~~[ ] unit tests~~ (this PR adds two components and their usage, so there are no separate units that need to be tested, but only the components which are tested using integration tests)
  - [x] integration tests


src/lib/utils/clsx-helper.ts
Slartibartfass2, line 1: Are you sure about this? We will have to adjust the import for every new component?
What about adjusting the `components.json` to have a separate shadcn utils directory?

.idea/codeStyles/Project.xml
Slartibartfass2, line 1: Do you know what these files do? Do they adhere to our prettier and eslint config?

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+<button
+    type="button"
+    class="flex flex-col lg:flex-row h-fit w-full gap-3 lg:gap-10 items-start lg:items-center justify-between px-5 py-2
+    border border-container-border-grey rounded-md hover:bg-container-border-grey/35 {project.archived
+        ? 'opacity-25'
+        : ''}"
+    onclick={() => goto(`/project/${project.id}/dashboard`)}
+>
+    <div class="flex flex-col h-fit w-fit items-start">
+        <h2 class="truncate">{project.name}</h2>
```
Slartibartfass2, line 41: This truncate doesn't seem to work.
image

With less space, this creates a horizontal scroll bar:
image

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+Furthermore this component is clickable and navigates to the corresponding project homepage.
+
+Usage:
+```svelte
+    <ProjectListEntry project={demoProject} members={memberUserSpecArray} stage={3} stageProgress={70} />
+```
+-->
+<button
+    type="button"
+    class="flex flex-col lg:flex-row h-fit w-full gap-3 lg:gap-10 items-start lg:items-center justify-between px-5 py-2
```
Slartibartfass2: `gap-3` seems a lot to me. What do you think about `gap-2` or `gap-1.5`?
```suggestion
    class="flex flex-col lg:flex-row h-fit w-full gap-2 lg:gap-10 items-start lg:items-center justify-between px-5 py-2
```
```suggestion
    class="flex flex-col lg:flex-row h-fit w-full gap-1.5 lg:gap-10 items-start lg:items-center justify-between px-5 py-2
```

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+    const { project, members, stage, stageProgress }: ProjectListEntryProps = $props();
+
+    /**
+     * Combine the first- and lastname of the members array into a string as following:
+     * [{firstName: "John", lastName: "Doe", email: "john@doe.com"}, ... ] -->
+     * "John Doe, ..."
+     */
+    const getMemberNames = () => {
+        return members.map((member: User) => `${member.firstName} ${member.lastName}`).join(", ");
+    };
```
Slartibartfass2: I added a similar method in #124 for the authors. Can you make this a public method somewhere that takes an object with `firstName` and `lastName` so that we can pass either authors or users and get the same result?

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+
+Usage:
+```svelte
+    <ProjectListEntry project={demoProject} members={memberUserSpecArray} stage={3} stageProgress={70} />
+```
+-->
+<button
+    type="button"
+    class="flex flex-col lg:flex-row h-fit w-full gap-3 lg:gap-10 items-start lg:items-center justify-between px-5 py-2
+    border border-container-border-grey rounded-md hover:bg-container-border-grey/35 {project.archived
```
Slartibartfass2: You also used `hover:bg-container-border-grey/35` in the `PaperListEntry` component.
We could store this color separately or this hover behavior as a custom class. What do you think about this?

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+                <span class="italic">no members</span>
+            {/if}
+        </div>
+    </div>
+    <div
+        class="flex flex-row w-full gap-x-5 lg:gap-x-2.5 items-center justify-start lg:justify-end"
+    >
+        <span class="h-fit w-fit text-nowrap">Stage {stage}</span>
+        <Progress
+            class="h-2.5 w-full min-w-16 lg:max-w-32"
```
Slartibartfass2: The contrast between the progress background and the entry background seems a bit low to me, and it gets even lower when hovering.
What's your opinion on slate-200 by default and slate-300 while hovering (similar to the user avatar in `PaperListEntry`)?
```suggestion
            class="h-2.5 w-full min-w-16 lg:max-w-32 bg-slate-200 group-hover/project-list-entry:bg-slate-300"
```

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+            {:else}
+                <span class="italic">no members</span>
+            {/if}
+        </div>
+    </div>
+    <div
+        class="flex flex-row w-full gap-x-5 lg:gap-x-2.5 items-center justify-start lg:justify-end"
+    >
+        <span class="h-fit w-fit text-nowrap">Stage {stage}</span>
+        <Progress
```
Slartibartfass2, line 51: Do you think it would make sense to show the progress in % when hovering the progress bar?
There would be a tooltip in develop now.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: Same for the truncate of the members.

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+
+Furthermore this component is clickable and navigates to the corresponding project homepage.
+
+Usage:
+```svelte
+    <ProjectListEntry project={demoProject} members={memberUserSpecArray} stage={3} stageProgress={70} />
+```
+-->
+<button
+    type="button"
```
Slartibartfass2: Is there a reason why this is no a tag?
It would enable opening the project in a new tab via middle mouse click or Ctrl + left click (depending on the user's settings)

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -0,0 +1,71 @@
...
+        return members.map((member: User) => `${member.firstName} ${member.lastName}`).join(", ");
+    };
+</script>
+
+<!--
+@component
+Container displaying important project information.
+
+This component shows the
+  - project id and name
```
Slartibartfass2: The project ID isn't displayed, should we?

src/lib/components/composites/utils/NamedList.svelte
```diff
@@ -0,0 +1,82 @@
...
+            <h2>{listName}</h2>
+        {/if}
+        <ul class="space-y-4 pb-1 scroll-box">
+            {#each loadedItems as item}
+                <li>
+                    {@render listItemComponent?.(item)}
+                </li>
+            {/each}
+        </ul>
+        <!-- eslint-disable-next-line @typescript-eslint/no-unused-vars -->
```
Slartibartfass2: This is not used.
```suggestion
```

src/lib/components/composites/utils/NamedList.svelte
```diff
@@ -0,0 +1,82 @@
...
+    {:catch error}
+        <h2>{listName}</h2>
+        <div class="flex flex-row items-center gap-x-2 p-4">
+            <CircleAlert size={16} class="text-neutral-500" />
```
Slartibartfass2: What about making the icon a bit bigger and the error text red?
```suggestion
            <CircleAlert size={20} class="text-neutral-500" />
```

src/lib/utils/statistics-helper.ts
```diff
@@ -0,0 +1,29 @@
+import { type Paper, ReviewDecision } from "$lib/model/backend";
+
+/**
+ * Calculate the progress of a stage.
+ *
+ * The progress is defined as the number of papers that are already accepted or declined (so no further
+ * decision need to be made) relative to the total number of papers (including the unreviewed a paper that need
+ * the decision of the arbiter).
+ *
+ * If no papers are provided, the progress is set to 0 per default.
```
Slartibartfass2: Minor typos
```suggestion
 * Calculate the progress of a stage.
 *
 * The progress is defined as the number of papers that are already accepted or declined (so no further
 * decision is required) relative to the total number of papers (including the unreviewed papers that need
 * the decision of the arbiter).
 *
 * If no papers are provided, the progress is set to 0 by default.
```

Shouldn't it be "including the reviewed papers that need the final decision of the arbiter"?

src/lib/utils/statistics-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param papers
```
Slartibartfass2: Either provide a comment or omit this param doc tag.

src/lib/utils/statistics-helper.ts
```diff
@@ -0,0 +1,29 @@
...
+        (paper) =>
+            paper.reviewData !== undefined &&
+            (paper.reviewData.finalDecision === ReviewDecision.Accepted ||
+                paper.reviewData.finalDecision === ReviewDecision.Declined),
+    ).length;
+
+    return Math.round((decidedPapers / totalPapers) * 100);
+}
+
+export { calculateStageProgress };
```
Slartibartfass2, line 28: Is there a reason why export it like that and not directly at the function declaration?

src/routes/+page.svelte
```diff
@@ -1,12 +1,29 @@
...
+        <h2>Open Reviews</h2>
+    </section>
+    <NamedList listName="Projects" items={projectMetadata} showNumberOfListItems={true}>
+        {#snippet listItemComponent(componentData)}
+            <ProjectListEntry {...componentData} />
+        {/snippet}
+        {#snippet listItemSkeleton()}
+            <ProjectListEntrySkeleton />
+        {/snippet}
+    </NamedList>
```
Slartibartfass2: Maybe wrap this in a section or make `NamedList` a section itself.

src/routes/+page.ts
```diff
@@ -0,0 +1,123 @@
...
+import {
+    type Paper,
+    type Project,
+    type ProjectMetadata,
+    type StageEntry,
+    type User,
+} from "$lib/model/backend";
+import { calculateStageProgress } from "$lib/utils/statistics-helper";
+
+// import type { Project, ProjectMetadata } from "$lib/model/backend";
```
Slartibartfass2: Leftover?
```suggestion
```

src/lib/components/composites/utils/NamedList.svelte
```diff
@@ -0,0 +1,82 @@
...
+is added to the list name / header, like 'yourListName (10)'.
+
+While the list is loading, it displays 10 skeleton list items.
+-->
+<div class="flex flex-col h-full w-full px-5 gap-y-5">
+    {#await items}
+        <h2>{listName}</h2>
+        <ul class="space-y-4 pb-1 scroll-box">
+            <!-- eslint-disable-next-line @typescript-eslint/no-unused-vars -->
+            {#each Array(10) as _}
```
Slartibartfass2: I can imagine that 10 skeleton items would be too much for some pages, e.g., the forward and backward references in the paper view. Shall we make it a prop?

src/lib/components/composites/utils/NamedList.svelte
```diff
@@ -0,0 +1,82 @@
...
+        numberOfItems?: number;
+    }
+
+    const {
+        listName,
+        items,
+        listItemComponent,
+        listItemSkeleton,
+        showNumberOfListItems = false,
+        numberOfItems = -1,
```
Slartibartfass2: Is there a reason why you don't make this undefined by default?
As a user, I would expect to display this number, regardless of what value I would pass as prop.

src/routes/+page.ts
```diff
@@ -0,0 +1,123 @@
...
+
+async function requestProjectMetadata(project: Project): Promise<ProjectMetadata> {
+    const members: User[] = await BackendController.getInstance().project(project.id).getMembers();
+    const currentStage: number = await BackendController.getInstance()
+        .project(project.id)
+        .getStageCount();
+    const allPapersInCurrentStage: StageEntry[] = await BackendController.getInstance()
+        .project(project.id)
+        .stage(currentStage)
+        .getPapers();
```
Slartibartfass2: You can first create the project controller and reuse it for the subsequent requests.
```suggestion
    const projectController = BackendController.getInstance().project(project.id);
    const members: User[] = await projectController.getMembers();
    const currentStage: number = await projectController.getStageCount();
    const allPapersInCurrentStage: StageEntry[] = await projectController
        .stage(currentStage)
        .getPapers();
```

src/routes/+page.ts
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Loads projects for the user logged in.
+ *
+ * Therefore, request the project ids of the projects, the user logged in is member of and use
+ * these ids to request:
+ * <ul>
+ *     <li>the project members</li>
+ *     <li>the current project stage</li>
+ *     <li>the progress of the current stage</li>
+ * </ul>
```
Slartibartfass2: This HTML doesn't get rendered in the VSCode tooltip. Wouldn't it just work with Markdown lists?
image

```suggestion
 * - the project members
 * - the current project stage
 * - the progress of the current stage
```
which results in
image

src/routes/+page.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * TODO: check, whether this can be handled with a single request, e.g. on route /projects/[id]/projectMetadata/.
+ */
+export const load: PageLoad = () => {
+    const projectMetadata = BackendController.getInstance()
```
Slartibartfass2: Minor: When reading `projectMetadata` I would expect the metadata of a single project. Do you think we should use a plural name?

src/routes/+page.ts
Slartibartfass2: Furthermore, I got this error message:

 Loading http://localhost:3000/projects/2/stages/0/papers using `window.fetch`. For best results, use the `fetch` that is passed to your `load` function: https://svelte.dev/docs/kit/load#making-fetch-requests

Did you try anything with this? Let's investigate this further in a separate issue.

src/lib/controller/user-controller.ts
```diff
@@ -28,7 +28,34 @@ export class UserController implements IUserController {
...
 
     async getProjects(): Promise<Project[]> {
-        throw new Error("Method not implemented.");
+        const response = await this.client.get("projects");
+        try {
+            const projects: Project[] = await response.json();
+            return Promise.resolve(projects);
+        } catch (error: unknown) {
+            return Promise.reject(error);
+        }
```
Slartibartfass2: Now that you added that noop-catch this isn't required. I tested this in different scenarios, e.g., when an error is thrown in this method and if the response is HTML and every time the error is caught and `Error: Could not load projects.` is displayed.

So this can just be
```suggestion
        return await this.client.get("projects").then((response) => response.json());
```

This applies to the other methods too.

src/lib/components/composites/utils/NamedList.svelte
Slartibartfass2, line 1: The name utils seems a bit unfit in this case because the `NamedList` component isn't really a utility component, is it?
What's your opinion on `common` or just `list`/`lists` similar to the `input` directory in my sign up PR?

tests/integration/project-components/project-list-entry-skeleton.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { render, screen } from "@testing-library/svelte";
+
+describe("ProjectListEntrySkeletonComponent", () => {
+    test("When project list entry skeleton is rendered, then no data are displayed but only skeletons.", () => {
+        render(ProjectListEntrySkeleton);
+
+        // expect skeleton for project name, members and stage, but not progress
+        screen
+            .getAllByTestId("skeleton", { exact: false })
+            .forEach((skeleton) => expect(skeleton).toBeInTheDocument());
```
Slartibartfass2: Isn't this redundant? If you can fetch them with `getAllByTestId`, they have to be in the document.

tests/integration/project-components/project-list-entry.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect(screen.getByText("Demo Project")).toBeInTheDocument();
+        expect(screen.getByText("John Doe, Jane Doe")).toBeInTheDocument();
+        expect(screen.getByText("Stage 1")).toBeInTheDocument();
+
+        // Project (stage) progress is visualized using a progress bar
+        expect(screen.queryByText("60")).not.toBeInTheDocument();
+        expect(screen.queryByTestId("stage-progress-bar")).toHaveValue(60);
+    });
+
+    test("When all no members are provided, then the list entry shows a hint.", () => {
```
Slartibartfass2:
```suggestion
    test("When no members are provided, then the list entry shows a hint.", () => {
```

tests/integration/utils/named-list.test.ts
```diff
@@ -0,0 +1,120 @@
+import { expect, test, describe } from "vitest";
+import { render, screen } from "@testing-library/svelte";
+import NamedList from "$lib/components/composites/utils/NamedList.svelte";
+// @ts-expect-error "Snippets have implicetly type any"
```
Slartibartfass2: This is unused
```suggestion
```

tests/integration/utils/named-list.test.ts
```diff
@@ -0,0 +1,120 @@
...
+                items: componentData,
+                listItemComponent: listItemComponent,
+                listItemSkeleton: listItemSkeleton,
+            },
+        });
+
+        // List title is shown
+        expect(screen.getByText("Test List")).toBeInTheDocument();
+
+        setTimeout(() => {
```
Slartibartfass2: I'm not sure if this is executed because the method doesn't wait for the timeout to be finished.

AFAIK, you can use `waitFor` for this.

The same applies for the other tests.

Slartibartfass2 (changes requested):
Wow! This was a big one. You did an impressive job implementing the nested list, I really like it.

There are some minor things open to discuss.

tests/integration/project-components/project-list-entry-skeleton.test.ts
Merseleo: You are right, especially if their count is three, they must be in the document. Thank you, i deleted it.

tests/integration/utils/named-list.test.ts
Merseleo: Good catch thank, a relic of the old days when they were not yet correctly typed.

tests/integration/utils/named-list.test.ts
Merseleo: You are right, I made this mistake before and solved it with setTimeout, bur encapsulated in a promise. And then I can await the promise resolution, I will fix this.

tests/integration/utils/named-list.test.ts
Merseleo: Good catch thank, a relic of the old days when they were not yet correctly typed.

tests/integration/utils/named-list.test.ts
Merseleo: This also explains, why all tests passed after the first implementation, as most of the cases were not executed ...

tests/integration/utils/named-list.test.ts
Merseleo: No, 180° turn, we need this line. If I delete it, svelte check complains: 

Bildschirmfoto 2024-12-18 um 15 38 23

.idea/codeStyles/Project.xml
Merseleo: Yes, I had problems with the repo and cloned it again and changed the settings, so eslint and prettier use our configs and run on save and these settings were saved. If I remember correctly, you wanted me to save these settings back then?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: I can agree on 2, but 1.5 is too tight imo. What do you think of the change?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Probably a good idea to add a custom style, especially as this behavior could be useful again. I add it.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: At the beginning I thought it would be a great idea, like the PaperListEntry, but when dealing with a paper an ID can be useful, but what would be advantage of showing the ID of the project? I don't see it and imo it looks cleaner without the ID, so I would not add it, but update the comment.

src/lib/components/composites/utils/NamedList.svelte
Merseleo: You are right, a relic of old days (at the beginning I only display "Error" and no reason, which is now stored in the error object. I delete this comment.

src/lib/components/composites/utils/NamedList.svelte
Merseleo: I agree, a size of 20 is better, but I also tried red as text color and my preference is gray as I not an error, which can handled by the user directly and need further attention, but a hint, so the user doesn't wonder about the page behavior.

src/lib/utils/statistics-helper.ts
Merseleo: You are right, I will delete the param doc tag, as the name is self-describing enough, if you are ok with that?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Good catch, forgot to test the design on long titles. I try to fix this.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Great suggestion, I integrate this change.

src/lib/utils/statistics-helper.ts
Merseleo: No special reason, I like the style of exporting all non-internal functions at the end of the file more as it seems more clear to me, what is exported and what not (if more functions are added). But if you want to do it the other way, I am also ok with that and will change it?

src/routes/+page.svelte
Merseleo: Good idea, but I prefer to wrap it with a section, as the named list must not always be a section.

src/routes/+page.ts
Merseleo: Yep, a leftover, I delete it.

src/routes/+page.ts
Merseleo: Yes, I mixed it up with the documentation style in C++

src/lib/utils/clsx-helper.ts
Merseleo: I am sure, that I want to separate the utility functions a bit and create separate files, but I agree that this would be annoying to adjust the import for every new component. I will update the `components.json`. Are you ok with this solution?

src/lib/components/composites/utils/NamedList.svelte
Merseleo: I agree, `list` seems the best option.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Ok, did not know that. I sourced this method out and generalized it a bit, please have  look at it and provide feedback, whether you can use it that way?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: It could make sense, but you are only click away form this information, as the progress as percentage will be probably visible on the project dashboard, but if you want this feature, I can add it after the rebase. So it's your decision ...

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: No I read the wrong specification: As [this post](https://stackoverflow.com/questions/1827965/is-putting-a-div-inside-an-anchor-ever-correct) described, you are not allowed to nest everything inside an `a` tag, but as there are interactive elements and we are using HTML 5, theoretically it would allowed to use an `a` tag. I try it.

Merseleo: I tested it and I like this solution, so for now I will keep it.

src/lib/components/composites/utils/NamedList.svelte
Merseleo: Good idea.

src/lib/components/composites/utils/NamedList.svelte
Merseleo: You are right, `undefined` is the way better default value, changed it.

src/routes/+page.ts
Merseleo: You are right, as a new developer I would also expect the metadata of only project. I rename it.

src/routes/+page.ts
Merseleo: Thank you for this hint, I try to remember it for the future 👍.

No i dient not try anything with that, but I also noticed this warning and read the documentation regarding the fetch. But as we will switch to gRPC soon, we will delete the fetch calls anyway, so I did not see any reason, why to investigate further. Do you want me to investigate?

src/lib/controller/user-controller.ts
Merseleo: Thank you, forgot this.

Merseleo: Hopefully, I did not introduce a new error, as I cleaned this up.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: I tried many different things, but did not get it to work. Maybe you have an idea.

Merseleo (commented):
"A few minor things left" is quite relative, but thank you for the detailed review and great suggestions. Hopefully, I did not forget to work on any of your hints (expect the truncation problem).

src/lib/utils/clsx-helper.ts
Slartibartfass2: This is perfect!

.idea/codeStyles/Project.xml
Slartibartfass2: Yes that's great. I tested it with WebStorm and it seems alright.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: That's fine!

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: That's great. Thank you!

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: I think we can leave this for now. We can also add it later, when we notice that it should do that.

src/lib/utils/statistics-helper.ts
Slartibartfass2: No, that's fine. I like both ways.

src/routes/+page.ts
Slartibartfass2: No, let's see what happens when we use gRPC.

src/lib/controller/user-controller.ts
Slartibartfass2: Is there a reason why you didn't do that in `getAllProjects`?

tests/integration/utils/named-list.test.ts
Slartibartfass2: Alright. VS Code tells me that this is unused.

Can we find a solution here by adding a type declaration?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: It still seems that it doesn't work as expected:
image

The red area is the div around the progress bar. I found a solution by changing `w-full` of this div to `w-fit`.
The progress doesn't take the space when `w-full` is applied. However, we can add widths according to the screen size.
image
This looks better now imo.

I came to this solution, what are your thoughts about that?
```svelte
    <div
        class="flex flex-row w-fit gap-x-5 lg:gap-x-2.5 items-center justify-start lg:justify-end"
    >
        <span class="h-fit w-fit text-nowrap">Stage {stage}</span>
        <Progress
            class="h-2.5 w-24 sm:w-32 sm:bg-red-500 md:w-48 md:bg-green-400 lg:w-28 lg:bg-yellow-400 xl:w-40 xl:bg-blue-500 2xl:w-52 2xl:bg-purple-500 bg-slate-200 group-hover/project-list-entry:bg-slate-300"
            value={stageProgress}
            data-testid="stage-progress-bar"
        />
    </div>
```
The colors are just for knowing where the media queries start and end.

Slartibartfass2 (changes requested):
Great work again! We're nearly finished.

Don't forget to remove the example code in the `+page.ts`.

Furthermore, do you think we should add a hint when the list is empty or do you think the `Projects (0)` is enough?

tests/integration/utils/named-list.test.ts
Merseleo: I give it a second try...

src/lib/controller/user-controller.ts
Merseleo: Yes, because it forgot it ;)

tests/integration/utils/named-list.test.ts
Merseleo: I think I solved the problem by using the `createRawSnippet` API to create the Snippets. But please have a look at the solution and check, if you get any errors.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Ok weird, your rendering seems to be different, but when I test it in chrome I also have the problem. 

But I like your idea and integrated it.

Merseleo (commented):
Thank you for your suggestions and I deleted the demo data, so hopefully everything is done now, so I can rebase.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: You didn't set a default width. Was that on purpose?
Because now, no progress bar is shown at all when the screen is too narrow.
image

tests/integration/utils/named-list.test.ts
Slartibartfass2: This works!

Slartibartfass2 (changes requested):
We've come far, this looks great now.
How is your opinion on what commented in https://github.com/SE-UUlm/snowballr-frontend/pull/122#pullrequestreview-2515335907

You can rebase now. I think after this we're done.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: No, I will add a default with, however I do not really understand, how the min width is calculated here ...

Merseleo (commented):
Yes I saw your comment and thought about it. I didn't consider an empty list, but I agree that we should handle this case. Imo a hint would be the best. I will add it and rebase it and then you could give your final feedback, agree?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: What do you think of the new default width?

Slartibartfass2: I think that's good

Slartibartfass2 (dismissed):

Slartibartfass2 (approved):

PR #124, opened 2024-12-08
Added Sign In page

### LINKED ISSUE #1: [Use Case]: Sign In

User

### Describe the use case

As user, I want to sign in, when I'm not signed in and already have an account.

### Acceptance criteria

- show sign in screen, when the user intends to access a page but is not signed in
- when the user enters valid credentials, they should be redirected to the page they previously tried to access
- the user can enter an email and a password

### Anything else?

Related to #2, #3 and #5

grafik

Slartibartfass2:
Closes #1 

## What I have made

- Added Sign In Page using the input elements designed in #112 
  - the input fields don't use any schemas because all they need to do is matching the credentials on the server
- Added Sign In and Sign Out logic (this was a pain in my ass)
  - the backend endpoint sends back a user and a token, we only use the token because the user is stored in the tokens' payload
  - the token is stored in the local storage and used for every http request that follows
  - when the sign-out button is pressed, the token is removed from the local storage
  - so far so good, now the painful part
  - initially, I tried to decode the jwt in the root `layout.ts`, to that the user is easily accessible on any page
  - **but** svelte loads the `page.ts` and `layout.ts` not always on the client but initially on the web server, where the local storage is obviously not accessible
  - even though it also loads these pages in the client, that's not always the case (afaik only when the user loads a page in the browser, but not if they navigate between different pages of the same app)
  - this really sucked, so now we access the local storage on each page separately
  - note that the call to `getCurrentUser` is also sometimes executed on the server but always another time on the client
  - in between the response is undefined, and the navigation bars handle this case as seen below
  
2024-12-14 13_14_47-SnowballR – Mozilla Firefox

## Checklist

  - ~~[ ] unit tests~~ (I tried mocking the local storage, but somehow I couldn't mock the headers)
  - [x] integration tests

### DIFF
--- a/src/lib/components/composites/navigation-bar/UserMenu.svelte
+++ b/src/lib/components/composites/navigation-bar/UserMenu.svelte
@@ -9,9 +9,10 @@
     import type { Icon } from "lucide-svelte";
     import type { User } from "$lib/model/api/user";
     import UserAvatar from "$lib/components/composites/user-avatar/UserAvatar.svelte";
+    import { goto } from "$app/navigation";
 
     interface Props {
-        user: User;
+        user?: User;
     }
     const { user }: Props = $props();
 
@@ -47,6 +48,11 @@
             href: "/settings/account",
         },
     ];
+
+    function signOut() {
+        localStorage.removeItem("token");
+        goto("/signin");
+    }
 </script>
 
 <DropdownMenu.Root>
@@ -56,7 +62,11 @@
     <DropdownMenu.Content class="w-60" align="start" side="bottom" sideOffset={0}>
         <DropdownMenu.Group>
             <DropdownMenu.GroupHeading>
-                {`${user.firstName} ${user.lastName}`}
+                {#if user}
+                    {`${user.firstName} ${user.lastName}`}
+                {:else}
+                    Loading User...
+                {/if}
             </DropdownMenu.GroupHeading>
             <DropdownMenu.Separator />
             <DropdownMenu.Group>
@@ -73,7 +83,7 @@
                 {/each}
             </DropdownMenu.Group>
             <DropdownMenu.Separator />
-            <DropdownMenu.Item>
+            <DropdownMenu.Item onclick={signOut}>
                 <LogOut class="mr-2 size-4" />
                 <span>Sign out</span>
                 <DropdownMenu.Shortcut>⇧⌘Q</DropdownMenu.Shortcut>
--- a/src/lib/components/composites/navigation-bar/UserMenu.svelte
+++ b/src/lib/components/composites/navigation-bar/UserMenu.svelte
@@ -10,6 +10,7 @@
     import type { User } from "$lib/model/api/user";
     import UserAvatar from "$lib/components/composites/user-avatar/UserAvatar.svelte";
     import { goto } from "$app/navigation";
+    import { BackendController } from "$lib/controller/backend-controller";
 
     interface Props {
         user?: User;
@@ -49,9 +50,10 @@
         },
     ];
 
-    function signOut() {
-        localStorage.removeItem("token");
-        goto("/signin");
+    async function signOut(): Promise<void> {
+        await BackendController.getInstance()
+            .signOut()
+            .then(() => goto("/signin"));
     }
 </script>
 
--- a/src/routes/(auth)/signin/+page.svelte
+++ b/src/routes/(auth)/signin/+page.svelte
@@ -1,4 +1,47 @@
+<script lang="ts">
+    import Input from "$lib/components/composites/input/Input.svelte";
+    import PasswordInput from "$lib/components/composites/input/PasswordInput.svelte";
+    import { Button } from "$lib/components/primitives/button/index.js";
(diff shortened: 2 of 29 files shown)


src/lib/components/composites/navigation-bar/UserMenu.svelte
```diff
@@ -47,6 +48,11 @@
             href: "/settings/account",
         },
     ];
+
+    function signOut() {
+        localStorage.removeItem("token");
+        goto("/signin");
```
Merseleo: I thought we want to explicitly logout? So you need to call the logout method from the BackendController.

src/lib/controller/backend-controller.ts
```diff
@@ -82,8 +91,13 @@ export class BackendController implements IBackendController {
     }
 
     thisUser(): IUserController {
-        // TODO: exchange this by logged in user (will be fixed in https://github.com/SE-UUlm/snowballr-frontend/pull/124/)
-        return this.user(0);
+        const jwt = localStorage.getItem("token");
+        if (!jwt) {
+            throw new Error("No JWT stored in the local storage");
```
Merseleo: The reason for this error is, that no user is logged in, but is requested, right? Maybe you can add the reason to error message?

src/lib/model/backend.ts
```diff
@@ -97,3 +97,7 @@ export interface Author {
 }
 
 export type AuthorSpec = Omit<Author, "id">;
+
+export interface SignInResponse {
+    token: string;
```
Merseleo: If I understood it correctly, you need a token and a refresh token, right? So I guess you need to store both and define both in the interface.

tests/integration/input/password-input.test.ts
```diff
@@ -94,4 +100,16 @@ describe("PasswordInput", () => {
             expect(input).toHaveAttribute("type", "password");
         });
     });
+
+    test("When `showForgotPasswordLink` is false, then link is not shown", () => {
```
Merseleo:
```suggestion
    test("When `showForgotPasswordLink` is false, then the link is not shown", () => {
```

src/routes/(auth)/signin/+page.svelte
```diff
@@ -1,4 +1,86 @@
...
+        </form>
+        <div class="text-center text-sm">
+            You don't have an account?
+            <a href="/signup" class="underline"> Sign Up </a>
+        </div>
+        {#if error}
+            <Alert.Root variant="destructive">
+                <CircleAlert class="size-4" />
+                <Alert.Title>Error</Alert.Title>
+                <Alert.Description>{error}</Alert.Description>
```
Merseleo:
```suggestion
                <Alert.Title>{error}</Alert.Title>
```

What do you think of only providing the error and not showing: "Error: ...", as it is red and the error message already indicates that it is an error?

src/routes/(auth)/signin/+page.svelte
```diff
@@ -1,4 +1,86 @@
...
+            <Input
+                class="w-full"
+                inputId="email"
+                label="Email"
+                placeholder="john.doe@example.com"
+                required
+                type="email"
+                bind:value={email}
+            />
+            <PasswordInput class="w-full" showForgotPasswordLink bind:value={password} />
```
Merseleo: Imo the we should not use the PasswordInput component directly here, as it shows what is missing (so at least two numbers etc.), but is this really necessary / usual to show on the login page?

Merseleo (commented):
There are only a few small things, we can discuss. To properly test the "Login" we need a running and correct working backend, so let's a bit and hopefully we will get this soon, to test the "Login".

src/lib/components/composites/navigation-bar/UserMenu.svelte
Slartibartfass2: Yes, you're right. I forgot that when I implemented this back then.

I changed it now.

src/lib/controller/backend-controller.ts
Slartibartfass2: I replaced this with the `getCurrentUser` method, so this is outdated.

src/lib/model/backend.ts
Slartibartfass2: I'll wait for the new API for this.

src/routes/(auth)/signin/+page.svelte
Slartibartfass2: No, it isn't this was introduced due to the changes in the sign-up PR and rebasing of this branch. I added a flag so that these inputs aren't validated, as they shouldn't.

src/routes/(auth)/signin/+page.svelte
Slartibartfass2: I agree that's enough.

Slartibartfass2 (commented):
Thank you for the review

tests/integration/input/password-input.test.ts
Slartibartfass2: Added it

luca-schlecker:
Closing in favor of #373

PR #126, opened 2024-12-08
Added basic api controller structure

### LINKED ISSUE #125: [Feature]: Add basic API controller structure

_No response_

Currently, we can't use the backend API, because we don't have the implementations.

Add classes that implement the controller APIs. They don't need to have logic, but should be able to use.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #125 

## What I have made

- Added .env.example file as example for the .env file that is used to pass environment variables to the application
  - the data in it can be imported using `import { PUBLIC_API_BASE_URL } from "$env/static/public";`
  - each key requires the prefix `PUBLIC_` for code that runs on the client
- Added controller implementations according to interfaces
  - all methods currently throw an exception because they're not implemented yet (but will be)
- Added an `HttpClient` that handles fetch calls for different HTTP methods
- the client is created for each controller individually with a fixed path that may be passed by another wrapping controller

## Checklist

  - [x] unit tests
  - ~~[ ] integration tests~~ (no components were added)
  - ~~[ ] end-to-end tests~~ (not available yet)


.github/workflows/code_quality_checks.yml
Merseleo, line 1: I think there is better way to create an .env file for the pipeline than copying the example file: https://www.restack.io/p/ci-cd-machine-learning-answer-env-files-github-actions-cat-ai. Maybe you could have a look and give me your opinion? (same for the lighthouse job)

package.json
```diff
@@ -10,12 +10,12 @@
...
+        "test": "vitest && npm run test:e2e",
         "show-coverage": "open coverage/index.html",
-        "test:unit": "vitest run --project unit",
-        "test:unit:ui": "npm run test:unit -- --ui",
-        "test:integration": "vitest run --project integration",
-        "test:integration:ui": "npm run test:integration -- --ui",
+        "test:unit": "vitest --project unit",
+        "test:unit:ui": "npm run test:unit -- --ui --watch --open",
+        "test:integration": "vitest --project integration",
+        "test:integration:ui": "npm run test:integration -- --ui --watch --open",
```
Merseleo, line 18: Very good idea, thank you for adding this (seems helpful)!

src/lib/backend-api.ts
```diff
@@ -46,7 +46,7 @@ export interface IStageEntryController {
 
     getReviews(): Promise<Review[]>;
     addReview(spec: ReviewSpec): Promise<Review>;
-    review(userId: number): IReviewController;
+    review(reviewerId: number): IReviewController;
```
Merseleo, line 49:
```suggestion
    getReviewController(reviewerId: number): IReviewController;
```

Merseleo: Same for the other controller, imo the names are kind of misleading or meaningless, what do you think?

src/lib/backend-api.ts
```diff
@@ -90,8 +90,8 @@ export interface IAuthorController {
 }
 
 export interface IBackendController {
-    login(username: string, password: string): Promise<User>;
-    logout(): Promise<void>;
+    signIn(username: string, password: string): Promise<User>;
```
Merseleo: Is it now possible to sign in with both the username and the email?

src/lib/controller/author-controller.ts
Merseleo, line 1: Since when do we have a separate route for the authors?

src/lib/controller/backend-controller.ts
```diff
@@ -0,0 +1,105 @@
...
+        throw new Error("Method not implemented.");
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async createProject(spec: ProjectSpec): Promise<Project> {
+        throw new Error("Method not implemented.");
+    }
+
+    project(projectId: number): IProjectController {
+        return new ProjectController(projectId);
```
Merseleo, line 60: Why do you always return a new controller? Wouldn't it be better to store the controller for the user or project etc. and check, whether already a controller exists before creating a new one. I can image that this implementation will create a lot of unnecessary / duplicate objects?

Merseleo: Same for the other methods, which return a controller instance ...

src/lib/controller/http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param path The path to the API endpoint. This path will be appended to the `PUBLIC_API_BASE_URL` environment variable.
+     *
+     * @example
+     * const client = new HttpClient("api/v1/example");
+     * // When the `PUBLIC_API_BASE_URL` environment variable is `http://localhost:3000`
+     * // the base URL of the API will be `http://localhost:3000/api/v1/example`
+     */
+    constructor(path: string = "") {
+        this.basePath = `${PUBLIC_API_BASE_URL}${path ? `/${path}` : ""}`;
```
Merseleo, line 31:
```suggestion
        this.basePath = `${PUBLIC_API_BASE_URL}${path ? `/${path}` : "/"}`;
```

Don't you need a forward slash in default case?

tests/unit/controller/http-client.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    http.get(`${PUBLIC_API_BASE_URL}`, () => {
+        return HttpResponse.json({
+            data: "get-root",
+        });
+    }),
+
+    http.post(`${PUBLIC_API_BASE_URL}/echo`, async ({ request }) => {
+        return HttpResponse.json(await request.json());
+    }),
+];
```
Merseleo: Maybe move this "mock" of a http server into a separate file?

tests/unit/controller/http-client.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        const httpClient = new HttpClient("test");
+
+        const response = await httpClient.delete();
+
+        expect(response).toBeDefined();
+        expect(response.status).toBe(200);
+        expect(await response.json()).toEqual({ data: "delete-test" });
+    });
+
+    test("When get request is called without path, then correct data is returned", async () => {
```
Merseleo:
```suggestion
    test("When get request is called without path, then root data is returned", async () => {
```

tests/unit/controller/http-client.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        const httpClient = new HttpClient();
+
+        const response = await httpClient.get();
+
+        expect(response).toBeDefined();
+        expect(response.status).toBe(200);
+        expect(await response.json()).toEqual({ data: "get-root" });
+    });
+
+    test("When get request is called with body, then correct data is returned", async () => {
```
Merseleo:
```suggestion
    test("When post request is called with body, then correct data is returned", async () => {
```

Your GET requests do not have a body ;)

tests/unit/controller/http-client.test.ts
Merseleo, line 1: I know it does not improve the coverage, but for completeness I think it would be nice to check the behavior, if no valid request (wrong path) is sent?!

src/lib/controller/project-controller.ts
```diff
@@ -0,0 +1,71 @@
+import type { Criterion, CriterionSpec, Project, ProjectSpec, User } from "$lib/model/backend";
+import type { ICriterionController, IProjectController, IStageController } from "../backend-api";
+import { CriterionController } from "./criterion-controller";
+import { HttpClient } from "./http-client";
+import { StageController } from "./stage-controller";
+
+export class ProjectController implements IProjectController {
+    private client: HttpClient;
+    private path: string;
```
Merseleo: I think they can be readonly / final (same in other controllers too)?!

Merseleo (changes requested):
Imo a good foundation for the communication with the backend, however there are still some open questions from my side (+ some minor suggestions)

.github/workflows/code_quality_checks.yml
Slartibartfass2: Currently, I only do this, so that the code compiles. We can think about passing real values later imo.

package.json
Slartibartfass2: Yes, apparently `--ui` doesn't start the tests in watch mode automatically, which doesn't make sense to me.

src/lib/backend-api.ts
Slartibartfass2: I think the reason was that it can be used like this:
```ts
BackendController.getInstance().project(2).stage(34).paper(2).review(4).get();
```
With more descriptive names, this would look like the following code
```ts
BackendController.getInstance()
        .getProjectController(2)
        .getStageController(34)
        .getPaperController(2)
        .getReviewController(4)
        .get();
```

In my opinion, either way is fine. Do you want me to change this?

src/lib/backend-api.ts
Slartibartfass2: No `username` is the email, I fixed this in #124, but I can do it already here if you like.

src/lib/controller/author-controller.ts
Slartibartfass2: See https://snowballr.informatik.uni-ulm.de:8443/swagger/#/authors

But I don't see any use cases where we would use that.

src/lib/controller/backend-controller.ts
Slartibartfass2: Where would you store the controller?
I thought, we could create the controller e.g. when navigating to the project dashboard and there we can use it to make all the calls. When we navigate to another page, won't exist anymore.

Currently, this is a nice way to pass the IDs and paths and easily call the API. When using singletons, the paths and IDs would need to be handled differently.

src/lib/controller/http-client.ts
Slartibartfass2: Afaik, trailing slashes can usually be omitted.

src/lib/controller/project-controller.ts
Slartibartfass2: Yes true, I changed that.

tests/unit/controller/http-client.test.ts
Slartibartfass2: Good idea, I changed it.

tests/unit/controller/http-client.test.ts
Slartibartfass2: Yes I agree, I added it.

Slartibartfass2 (commented):
Thanks for the review, I applied the changes.

tests/unit/controller/http-client.test.ts
Slartibartfass2: Edit: This is a bit tricky because in the CI this is executed differently than locally. They don't recommend using not mocked endpoint:
```
[MSW] Warning: intercepted a request without a matching request handler:

  • GET https://snowballr.informatik.uni-ulm.de:8443/api/non-existing

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
```
So in this case, I'll remove it again.

src/lib/controller/http-client.ts
Merseleo: They can be omitted, especially since it is only the base path and all further relative paths starts with a slash, but I thought is is better practice to provide it (https://stackoverflow.com/questions/5948659/when-should-i-use-a-trailing-slash-in-my-url)

src/lib/controller/http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+export class HttpClient {
+    private readonly basePath: string;
+
+    /**
+     * Creates a new HTTP client.
+     *
+     * @param path The path to the API endpoint. This path will be appended to the `PUBLIC_API_BASE_URL` environment variable.
+     *
+     * @example
+     * const client = new HttpClient("api/v1/example");
```
Merseleo, line 26:
```suggestion
     * const client = new HttpClient("/api/v1/example");
```

or am I wrong (for your current implementation)?

tests/unit/controller/http-client.test.ts
Merseleo: Ah ok, if they propose to not test this behavior, I will believe them ...

.github/workflows/code_quality_checks.yml
Merseleo: But why do it later and do not start with this example value now?

package.json
Merseleo: It is what it is ...

src/lib/backend-api.ts
Merseleo: Ok, i understand why the current solution looks nice. Probably we should stay with it for more readable code. Your demonstration showed that this is probably more useful than self-describing method names (if you agree, resolve it).

src/lib/backend-api.ts
Merseleo: It would be great ...

src/lib/controller/author-controller.ts
Merseleo: Ok, I already thought I missed the use case. But as this endpoint exists in the backend, I think it is ok to add it and delete the controller later, if no meaningful use case appears (if you agree, close this conversation).

src/lib/controller/backend-controller.ts
Merseleo: Ok, i understand the advantage with the Ids, so let's stick with this solution (and if we face performance issues of too many object creations / deletions we think of another solution).

Merseleo (changes requested):
Only some minor changes and open discussions left, let's finish it ...

src/lib/controller/http-client.ts
Slartibartfass2: I'm not sure what your reasoning is by relating to this article because they all argument that a trailing slash doesn't make sense in most cases. In the [OpenAPI Spec](https://spec.openapis.org/oas/v3.1.1.html#paths-object) trailing slashes are omitted.

src/lib/controller/http-client.ts
Slartibartfass2: In the current implementation, the slash is added for a given path.

src/lib/backend-api.ts
Slartibartfass2: I changed it.

.github/workflows/code_quality_checks.yml
Slartibartfass2: Yes we can also do it now. I changed it.

Slartibartfass2 (commented):
Thanks again for taking the time to review this.

src/lib/controller/http-client.ts
Merseleo: My mistake.

Merseleo (approved):
I think every is clear now ...

PR #144, opened 2024-12-18
Added paper details to Paper View

### LINKED ISSUE #41: [Use Case]: View Paper Details

Project Member

### Describe the use case

As project member, I want to view the paper details in the paper view.

### Acceptance criteria

- show paper view page for specific paper
- show paper information
  - title
  - authors
  - year
  - publisher
  - abstract

### Anything else?

Related to #34 and #61

Slartibartfass2:
Closes #41 

## What I have made

- The main change of this PR is that `PaperDetailsCard.svelte` now displays all the information, planned in the design
  - initially only the paper title, authors, year and publisher are shown
  - the publication type, publication name and external ID can be shown, when pressing 'show more information'
  - by default all inputs are made read-only, but there are two edit mode buttons to toggle that property
- For that functionality, I added a `ToggleableInput` component, which is an auto-sizing textarea
  - it's a textarea because it can handle multiple lines
  - it's auto-sizing because that's what common textareas can't do (we use an action for that)
- I made most paper related components async because the data won't be there instantly
  - all of them use skeletons while loading
- For a long time, I had the problem that the abstract textarea won't stop expanding when it reaches the end of the card, after trying many solutions with no success, I implemented a semi-hard-coded solution that works
- This solution is a svelte action applied to the textarea
  - on creation, on update, when input is added to the textarea or when the screen size changes the action is executed
  - the action calculates how much space is left between the top of the textarea and the end of the card as it's supposed to be, this distance is added as max-height attribute
  - when exceeding this height, a scrollbar is added to the abstract

## Checklist

  - ~[ ] unit tests~ no logic that be tested by a unit test added
  - [x] integration tests
  - ~[ ] end-to-end tests~ not available yet


src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -7,15 +7,17 @@
     import Tooltip from "./Tooltip.svelte";
 
     interface Props {
-        paperId: number;
+        loadingPaperId: Promise<number>;
```
luca-schlecker: What's your reasoning for pulling that promise inside this component?
Couldn't this component be used as is when wrapped in svelte's `{#await ...}`?

This question arises multiple times throughout the code, one comment thread is enough though imo.

src/lib/components/composites/input/ToggleableInput.svelte
```diff
@@ -0,0 +1,49 @@
...
+This is used in `AbstractToggleableInput` and probably won't be needed elsewhere.
+
+Usage:
+```svelte
+    <ToggleableInput {isEditable} {value} />
+```
+-->
+<textarea
+    use:autosize
+    use:inputAction={inputActionProps}
```
luca-schlecker: Do you think this workaround is necessary? The abstract's `textarea` could instead fill the remaining space. It would reduce complexity with the caveat of always filling the remaining space even when provided with a short text (see image below). Visually there wouldn't be too much of a difference with that one exception.

image

src/lib/components/composites/paper-components/PaperInfo.svelte
```diff
@@ -1,27 +1,36 @@
...
-            >
-        {:else}
-            <span class="italic">unknown authors</span>
-        {/if}
+{:then paper}
+    <div class="grid grid-flow-row gap-0">
+        <div class="flex flex-row gap-1 items-center truncate">
+            {#if "id" in paper}
+                <div class="w-fit text-default-sb-nc text-neutral-500">#{paper.id}</div>
+            {/if}
```
luca-schlecker, line 23: Is there a reason the `id` could be missing?

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
```diff
@@ -0,0 +1,40 @@
...
+    import Skeleton from "$lib/components/primitives/skeleton/skeleton.svelte";
+    import type { Paper } from "$lib/model/backend";
+    import { cn } from "$lib/utils/shadcn-helper";
+    import type { WithElementRef } from "bits-ui";
+    import ToggleableInput from "../../input/ToggleableInput.svelte";
+    import type { HTMLAttributes } from "svelte/elements";
+
+    type Props = WithElementRef<HTMLAttributes<HTMLDivElement>> & {
+        key: string;
+        value: unknown;
```
luca-schlecker, line 11: As far as I can tell `value` is used for class names and content at the same time. This should probably be explained somewhere.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,11 +1,86 @@
...
+            basicInfos = {
+                Title: paper.title,
+                Authors: getNames(paper.authors),
+                Year: paper.year?.toString() ?? "N/A",
+                Publisher: "N/A",
+            };
+        })
+        .catch(() => {
+            basicInfos = { Title: "", Authors: "", Year: "", Publisher: "" };
+        });
```
luca-schlecker: I recognize a common pattern: Instead of relying on svelte's `{#await ...}` a state rune is used with a default value. In case you want to avoid `{#await ...}` (previous comment), I'd recommend simplifying that behavior similar to [what is shown here:](https://dev.to/jdgamble555/async-fetching-in-svelte-5-826)
```typescript
// resource.svelte.ts

export let resource = <T>(
    fn: () => Promise<T>,
    initialValue?: T
) => {

    const _rune = $state<{ value: T | undefined }>({
        value: initialValue
    });

    $effect(() => {
        fn().then((data) => {
            _rune.value = data;
        });
    });

    return _rune;
};
```

This could also be extended with a default error value. Maybe something like this:
```typescript
export let resource = <T>(
    fn: () => Promise<T>,
    initialValue?: T,
    onSuccess: (...) => T,
    onError?: (...) => T
) => { ... }
```

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -14,24 +89,86 @@ Paper Card for paper details in the Paper View component.
...
+                {/if}
+            </div>
+            {#await loadingPaper}
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[100%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[95%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[70%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[82%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[50%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[75%]" />
+                <Skeleton class="flex h-[1.625rem] rounded-full w-[90%]" />
```
luca-schlecker:
```suggestion
                {#each [100,95,70,82,50,75,90] as width}
                    <Skeleton class="flex h-[1.625rem] rounded-full w-[{width}%]" />
                {/each}
```

src/lib/components/composites/paper-components/paper-view/decision-buttons/DecisionButton.svelte
```diff
@@ -8,16 +8,29 @@
...
         buttonContent,
         tooltipContent,
+        loadingPaperId,
         onClick,
         class: className,
         ...restProps
     }: Props = $props();
+
+    let paperId = $state<number | undefined>(undefined);
+    loadingPaperId.then((id) => (paperId = id)).catch(() => (paperId = undefined));
```
luca-schlecker: Why not a nullable id?

tests/integration/paper-components/paper-view/paper-detail.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const spans = document.getElementsByTagName("span");
+        expect(spans.length).toEqual(1);
+        const keySpan = spans[0];
+
+        expect(keySpan.textContent).toEqual("Title");
+        expect(screen.queryByTestId("skeleton")).not.toBeNull();
+    });
+
+    test("When paper loading failed without error, then errot text is shown", async () => {
```
luca-schlecker:
```suggestion
    test("When paper loading failed without error, then error text is shown", async () => {
```

luca-schlecker (changes requested):
The page looks great, nice work. :+1: 

I stumbled upon issues with the integration tests though, I attached some details in the form of a log.
The pipeline doesn't seem to encounter the same problem.

Due to the upcoming gRPC work, I did not review changes regarding controllers or backend APIs.

[log.txt](https://github.com/user-attachments/files/18496177/log.txt)

src/lib/components/composites/PaperBookmarkButton.svelte
Slartibartfass2: I used this because the component does not heavily rely on the asynchronous loading, i.e., we can just display the component and as soon as the paper is loaded the functionality works as well.
Therefore, wrapping the component in an await block seemed too much to me.

Where else did you notice this?

src/lib/components/composites/input/ToggleableInput.svelte
Slartibartfass2: What did you use to produce that?

I tried plenty of approaches and none worked when the text overflows. The action was some sort of last resort.

src/lib/components/composites/paper-components/PaperInfo.svelte
Slartibartfass2: Yes, this is used when the data is only a `PaperSpec` with no ID. This is the case on the paper create page.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Slartibartfass2: Great catch! I added it to the component documentation.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Good catch. I added it.

src/lib/components/composites/paper-components/paper-view/decision-buttons/DecisionButton.svelte
Slartibartfass2: Please elaborate.

tests/integration/paper-components/paper-view/paper-detail.test.ts
Slartibartfass2: Good catch. Changed it.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Nice suggestion, I like it. It's way more clean and descriptive now.

Slartibartfass2 (commented):
Thank you for the review.


I couldn't reproduce the errors you got in that log. On which commit did you experience that?

src/lib/components/composites/input/ToggleableInput.svelte
luca-schlecker: The combination of
```css
flex-grow: 1;
flex-basis: 0;
overflow: scroll;
```
on the abstract.

src/lib/components/composites/paper-components/PaperInfo.svelte
luca-schlecker: I see, makes sense. 👍

src/lib/components/composites/PaperBookmarkButton.svelte
luca-schlecker: I guess the behavior you are describing is right. 👍 No need to change anything

src/lib/components/composites/paper-components/paper-view/decision-buttons/DecisionButton.svelte
luca-schlecker: Nevermind, I thought the promise could just take a nullable id but that is just nonsense semantically.

luca-schlecker:
> I couldn't reproduce the errors you got in that log. On which commit did you experience that?

Will have to test it on the machine that produced the errors again. This machine works through the tests nicely. Either fixed or flakey.

src/lib/components/composites/input/ToggleableInput.svelte
Slartibartfass2: I couldn't reproduce this.
Can you apply the changes to a new branch where I can try it for myself?

luca-schlecker (commented):
Looking great 👍. A rebase seems to be needed though.

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -7,15 +7,17 @@
...
     }
 
-    const { paperId, isBookmarkedDefault }: Props = $props();
+    const { loadingPaperId, isBookmarkedDefault }: Props = $props();
 
     let isBookmarked = $state(isBookmarkedDefault);
     let isHovered = $state(false);
     const tooltipText = $derived(isBookmarked ? "Remove from Reading List" : "Add to Reading List");
+    let paperId: string | undefined = $state(undefined);
+    loadingPaperId.then((id) => (paperId = id)).catch(() => (paperId = undefined));
```
luca-schlecker: This could be a call to `resource`.

src/lib/components/composites/paper-components/paper-view/decision-buttons/DecisionButton.svelte
```diff
@@ -8,16 +8,29 @@
...
         buttonContent,
         tooltipContent,
+        loadingPaperId,
         onClick,
         class: className,
         ...restProps
     }: Props = $props();
+
+    let paperId = $state<string | undefined>(undefined);
+    loadingPaperId.then((id) => (paperId = id)).catch(() => (paperId = undefined));
```
luca-schlecker: another call to `resource` here

luca-schlecker (changes requested):
Only these two minor things are left.

src/lib/components/composites/PaperBookmarkButton.svelte
Slartibartfass2: Great catch, I changed it.

src/lib/components/composites/paper-components/paper-view/decision-buttons/DecisionButton.svelte
Slartibartfass2: Great catch, I changed it.

luca-schlecker (approved):
Looking great! 🚀

PR #156, opened 2025-01-07
View open reviews on homepage

### LINKED ISSUE #121: [Use Case]: View Open Reviews

Merseleo:
Closes #121

## What I have made

- added list of open reviews on the _Homepage_ (making use of the _NamedList_ component)
  - a _PaperListEntry_ is shown, if the corresponding paper is either unreviewed or has the current status "Maybe" or has not enough reviews **and** belongs to a project, the user that is currently logged in, is member 
- added a skeleton component for a _PaperListEntry_

- added a new (project) controller method for requesting the current stage

## Checklist

  - [x] unit tests
  - [x] integration tests
  ~~- [ ] end-to-end tests~~ (not possible yet)

### DIFF
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -43,8 +43,8 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
         const allStageEntriesFromStageI: StageEntry[] = await projectController
             .stage(currentStage)
             .getPapers();
-        allUndecidedPapers.concat(
-            allStageEntriesFromStageI
+        allUndecidedPapers.push(
+            ...allStageEntriesFromStageI
                 .map((stageEntry) => ({
                     paper: stageEntry.paper,
                     projectId: project.id,
@@ -54,7 +54,6 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
                 .filter((paperInfo: PaperInfos) => isPaperUndecided(paperInfo.paper)),
         );
     }
-
     return allUndecidedPapers;
 }
 
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -7,7 +7,8 @@ import {
     type StageEntry,
     type User,
 } from "$lib/model/backend";
-import { calculateStageProgress, isPaperUndecided } from "$lib/utils/statistics-helper";
+import { calculateStageProgress } from "$lib/utils/statistics-helper";
+import { doesPaperNeedReview } from "$lib/utils/common-helper";
 
 interface PaperInfos {
     paper: Paper;
@@ -39,9 +40,7 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
     const latestStage: number = await projectController.getStageCount();
     const allUndecidedPapers: PaperInfos[] = [];
 
-    const numberOfRequiredReviews = await projectController
-        .get()
-        .then((project) => project.reviewDecisionMatrix.numberOfReviewers);
+    const numberOfRequiredReviews = project.reviewDecisionMatrix.numberOfReviewers;
 
     for (let i = currentStage; i <= latestStage; i++) {
         const allStageEntriesFromStageI: StageEntry[] = await projectController
@@ -55,11 +54,8 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
                     /// TODO: request this information, e.g. from local store
                     showReviewStatus: false,
                 }))
-                .filter(
-                    (paperInfo: PaperInfos) =>
-                        (isPaperUndecided(paperInfo.paper) ||
-                            paperInfo.paper.reviewData?.reviews.length) ??
-                        0 < numberOfRequiredReviews,
+                .filter((paperInfo) =>
+                    doesPaperNeedReview(paperInfo.paper, numberOfRequiredReviews),
                 ),
         );
     }
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -41,7 +41,7 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
 
     for (let i = currentStage; i <= latestStage; i++) {
         const allStageEntriesFromStageI: StageEntry[] = await projectController
-            .stage(currentStage)
+            .stage(i)
             .getPapers();
         allUndecidedPapers.push(
             ...allStageEntriesFromStageI
@@ -105,8 +105,6 @@ export const load: PageLoad = () => {
         });
 
     // attach noop-catch to handle promise rejection correctly (see https://svelte.dev/docs/kit/load#Streaming-with-promises)
-    projectsMetadata.catch(() => {});
-
     openReviews.catch(() => {});
 
     return { projectsMetadata, openReviews };
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -39,6 +39,10 @@ async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
     const latestStage: number = await projectController.getStageCount();
     const allUndecidedPapers: PaperInfos[] = [];
 
+    const numberOfRequiredReviews = await projectController
+        .get()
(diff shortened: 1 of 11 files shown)


Merseleo:
I just went through our meeting notes again and I read that the entries in this list should be grouped by project or that the corresponding project name should appear somewhere. What do you think about this at the moment? Should I still include this or should we put it at the end (and I'll open an issue for it) or should we ignore it completely?

tests/integration/paper-components/paper-list-entry-skeleton.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import ProjectListEntrySkeleton from "$lib/components/composites/paper-components/PaperListEntrySkeleton.svelte";
+import { render, screen } from "@testing-library/svelte";
+
+describe("PaperListEntrySkeletonComponent", () => {
+    test("When paper list entry skeleton is rendered, then no data are displayed but only skeletons.", () => {
+        const { unmount } = render(ProjectListEntrySkeleton);
+
+        // expect skeleton for paper id, paper title, authors
+        expect(screen.getAllByTestId("skeleton", { exact: false }).length).toBeGreaterThan(2);
+        unmount();
```
Slartibartfass2: Is there a reason why you didn't just split this up into two tests?

src/routes/+page.ts
```diff
@@ -58,5 +89,25 @@ export const load: PageLoad = () => {
...
+                throw new Error("Could not load open reviews.");
+            }
+        })
+        .then((openReviews) => openReviews.flat())
+        .catch(() => {
+            throw new Error("Could not load open reviews.");
+        });
+
+    // attach noop-catch to handle promise rejection correctly (see https://svelte.dev/docs/kit/load#Streaming-with-promises)
+    projectsMetadata.catch(() => {});
```
Slartibartfass2: This is duplicated in line 58/59

src/routes/+page.ts
```diff
@@ -27,25 +33,50 @@ async function requestProjectMetadata(project: Project): Promise<ProjectMetadata
...
 
+async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
+    const projectController = BackendController.getInstance().project(project.id);
+    const currentStage: number = await projectController.getCurrentStage();
+    const latestStage: number = await projectController.getStageCount();
+    const allUndecidedPapers: PaperInfos[] = [];
+
+    for (let i = currentStage; i <= latestStage; i++) {
+        const allStageEntriesFromStageI: StageEntry[] = await projectController
+            .stage(currentStage)
```
Slartibartfass2: That should be `i` instead of `currentStage`
```suggestion
            .stage(i)
```

Slartibartfass2 (changes requested):
Looks great. Just some smaller changes.

tests/integration/paper-components/paper-list-entry-skeleton.test.ts
Merseleo: As they imo belongs together and the PaperListEntrySkeleton has a random component in it, which I want to test by explicitly enforcing one random case. 
But I changed the first test case to explicitly enforcing the second case and so I use two components to test both cases of the random decision and as I said, imo this should be one test instead of two. Do you think differently?

src/routes/+page.ts
Merseleo: Thank you, deleted it.

src/routes/+page.ts
Merseleo: You are right, thank you!

Merseleo (commented):
Thank you for your feedback, hopefully everything is answered and done now.

tests/integration/paper-components/paper-list-entry-skeleton.test.ts
Slartibartfass2: I think you test two different things here: A skeleton with review and one without. IMO, that's two cases, but I'm also fine keeping it as is.

I didn't notice the randomness the first time, and we have to be careful with that so that there are no flaky tests. Good that you changed it.

Slartibartfass2 (dismissed):
Well done!

Just rebase and then we can merge it.

tests/integration/paper-components/paper-list-entry-skeleton.test.ts
Merseleo: Ok and you are right, at the beginning the test was "flaky", because I ignored this randomness, but I think now it should work. And I can now understand, why it could make sense to split these tests.

Merseleo:
I rebased and fixed the "bug" that papers, which are not undecided, but not have the required number of reviews are not listed in the open reviews. Please have a look at this change.

If everything is ok, you can merge.

src/routes/+page.ts
```diff
@@ -27,25 +33,59 @@ async function requestProjectMetadata(project: Project): Promise<ProjectMetadata
...
 
+async function requestUndecidedPapers(project: Project): Promise<PaperInfos[]> {
+    const projectController = BackendController.getInstance().project(project.id);
+    const currentStage: number = await projectController.getCurrentStage();
+    const latestStage: number = await projectController.getStageCount();
+    const allUndecidedPapers: PaperInfos[] = [];
+
+    const numberOfRequiredReviews = await projectController
+        .get()
+        .then((project) => project.reviewDecisionMatrix.numberOfReviewers);
```
Slartibartfass2: Here, we already have the whole project as an object, so we can just use the property directly
```suggestion
    const numberOfRequiredReviews = project.reviewDecisionMatrix.numberOfReviewers;
```

src/routes/+page.ts
```diff
@@ -27,25 +33,59 @@ async function requestProjectMetadata(project: Project): Promise<ProjectMetadata
...
+                    paper: stageEntry.paper,
+                    projectId: project.id,
+                    /// TODO: request this information, e.g. from local store
+                    showReviewStatus: false,
+                }))
+                .filter(
+                    (paperInfo: PaperInfos) =>
+                        (isPaperUndecided(paperInfo.paper) ||
+                            paperInfo.paper.reviewData?.reviews.length) ??
+                        0 < numberOfRequiredReviews,
```
Slartibartfass2: What are your thoughts on extracting this logic to a separate method, with an explanation of how you define an undecided paper?

Slartibartfass2: Maybe also add the mapping and the filtering function to the calls (`...getPapers().map(...)`)
I think this could help make this more descriptive.

Slartibartfass2 (changes requested):

src/routes/+page.ts
Merseleo: Sorry, missed it. Thank you, I will change it.

src/routes/+page.ts
Merseleo: So you want a separate method for determining, whether a paper needs further reviews? I can add it.

Regarding your second comment: I do not really understand what you want to say?

Slartibartfass2: Initially, I thought of something like the following
```ts
const allStageEntriesFromStageI: StageEntry[] = await projectController
            .stage(i)
            .getPapers()
            .map((stageEntry) => ({
                paper: stageEntry.paper,
                projectId: project.id,
                /// TODO: request this information, e.g. from local store
                showReviewStatus: false,
            }))
            .filter((paperInfo) =>
                doesPaperNeedReview(paperInfo.paper, numberOfRequiredReviews),
            );
 ```
 but then I noticed that the first part is async and this doesn't work this way, so nevermind.

Slartibartfass2 (approved):

src/routes/+page.ts
Merseleo: Ok

Issue #17, opened 2024-10-30
[Use Case]: Define Keywords For Review Mode


Slartibartfass2:
### Who is the primary actor in this use case?

Project Admin

### Describe the use case

As project admin, I want to define keywords, which are highlighted in the abstract of a paper in the review mode.

### Acceptance criteria

- the user can define multiple keywords
- the keywords must be unique
- the user can delete already existing keywords
- the maximum number of keywords is limited
- a keyword must be non-blank
- the maximum length of a keyword is limited
- a keyword can include spaces, i.e. it doesn't have to be just one word

### Anything else?

Part of #15

Related to #54

Issue #16, opened 2024-10-30
[Use Case]: Invite Users


Slartibartfass2:
### Who is the primary actor in this use case?

Project Admin

### Describe the use case

As project admin, I want to invite users to the project.

### Acceptance criteria

- a user, which isn't already a member of the project, can be invited
- a user, which is already a member of the project, can't be invited
- users with and without an account can be invited
- when a user gets invited, they receive an email to join the project
- a user who was invited to join a project can accept or decline the invitation
- a user who accepted an invitation is considered a project member
- an invitation has an expiry date, after which it is invalid
- a project admin can cancel the invitation, after which it is invalid

### Anything else?

Part of #15

PR #157, opened 2025-01-07
Create project

### LINKED ISSUE #14: [Use Case]: Create Project

User

### Describe the use case

As user, I want to create a project to conduct an SLR.

### Acceptance criteria

- the user has to provide a non-blank project name
- the user can invite other users as collaborators, but doesn't have to
- the newly created project uses the default values configured by the project creator in their user settings
- the user is considered a project admin and a project member after creation of a project

### Anything else?

_No response_

Merseleo:
Closes #14 

## What I have made

- added new `ChipsInput` component, which is inspired by the [Angular Chips component](https://v6.material.angular.io/components/chips/overview) (can be reused in #17)
  - this component is a collection of chips followed by an input to add new chips
  - chips can be added by ",", "+", "Enter", "Tab" and removed by "Backspace"
  - the user can navigate between the chips by "Arrow Left / Right"
  - the component can be extended by a function for suggestions which adds filtered suggestions beneath the component showing suggestions for the input, which can be navigated by "Arrow Up / Down"
  - the component can be extended by functions to validate the input and map the input to another input

- added button on _Homepage_ to create new project
- added dialog, in which the user can provide a project name and possible members to invite directly
  - create project with name on server
  - trigger a invitation by the server for every E-Mail provided in the second input
 - automatically navigate to the project dashboard after the creation, if the user wants to

- implemented the necessary controller methods for requesting users, creating a project and inviting members to a project

> **Hint**: To properly review this PR, you need to run the mock backend with the latest commit on the branch [feat/add-demo-projects](https://github.com/SE-UUlm/snowballr-mock-backend/tree/feat/add-demo-projects).

## Checklist

  - [x] unit tests
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (not possible yet)


package.json
```diff
@@ -59,5 +59,8 @@
         "vite": "^5.4.11",
         "vitest": "^2.1.5",
         "zod": "^3.23.8"
+    },
+    "dependencies": {
+        "fastest-levenshtein": "^1.0.16"
```
Slartibartfass2: This doesn't really matter, but we put all dependencies to the dev dependencies, so I think this one should be there too.

src/app.css
```diff
@@ -80,6 +80,7 @@
 /* width */
 ::-webkit-scrollbar {
     width: 8px;
+    height: 8px;
```
Slartibartfass2: What was the reason for this change?

src/routes/+page.svelte
```diff
@@ -15,6 +16,7 @@
 </svelte:head>
 <SimpleNavigationBar {user} title="SnowballR" />
 <main class="flex flex-row h-full w-full mb-10 gap-x-5 overflow-hidden">
+    <!-- TODO: exchange by the NamedList for the open reviews -->
```
Slartibartfass2: This TODO can be removed.
```suggestion
```

src/app.css
```diff
@@ -108,6 +109,10 @@
     }
 
     /* Custom Style Classes */
+    .text-h3 {
+        @apply text-lg font-semibold tracking-normal text-primary;
+    }
+
```
Slartibartfass2: I can't find where this is used. Is it an artifact from the development?
```suggestion
```

src/app.css
```diff
@@ -141,8 +146,7 @@
     .text-default-m-nc {
         @apply text-default-nc font-medium;
     }
-
-    .highlight-on-hover {
+    .text- .highlight-on-hover {
```
Slartibartfass2: Same here, `text-h3` isn't used.
```suggestion
    .highlight-on-hover {
```

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+
+    let {
+        items = $bindable(),
+        label,
+        labelPosition = "top",
+        searchSuggestions,
+        placeholder = "",
+    }: ChipsInputProps = $props();
+
+    /*
```
Slartibartfass2: Add another asterisk so that it counts as documentation for `selectedChipIndex`
```suggestion
    /**
```

Slartibartfass2: Also, \<length of items\> isn't displayed in VS Code. I recommend using square brackets.

Does it display the comment correctly in WebStorm?

Slartibartfass2: For both comments, it's the same for `selectedSuggestionIndex`

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Adds the input to the list of items and add a chip to the visualization,
+     * if the input is not a duplicate.
+     *
+     * Furthermore, it clears the input.
+     *
+     * @param item the item to add
+     */
+    function addItem(item: string): void {
+        if (item !== "" && !items.includes(item)) {
```
Slartibartfass2: Let's add `trim()` so it can't be blank either.
```suggestion
        if (item.trim() !== "" && !items.includes(item)) {
```

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+    function removeItem(index: number): void {
+        items = items.filter((_, i) => i !== index);
+    }
+
+    /**
+     * Handles a new (keyboard) input to this component.
+     * That means adding a new / removing an item or navigate between the items / suggestions.
+     *
+     * TODO: for reviewer: if you have a better idea for this handler, i am open for ideas, because
+     * i dont like it.
```
Slartibartfass2: In general, there's a lot of logic so it's challenging to make this a lot smaller, but there are some possibilities.

- sometimes you call `event.preventDefault()` and sometimes not, is there a reason for that?
- every time you call `document.getElementById("chips-input-" + label)?.focus();`
  - you can bind the div directly by using e.g., `bind:this={chipsInput}`, for an example look at `PasswordInput.svelte`
  - can it be called at the end of the whole handle method? Would there be a problem if the input is focussed, but none of these keys are pressed?
- same for the indices, maybe extract them to a reset method with a flag, which index has to be reset. This way, 2-3 lines can be reduced to one for each if block
- the navigating between the suggestions and chips is also very similar for all cases, at the same time it's difficult to read.
```ts
function nextIndex(step: number, current: number, max: number) {
  return (current + step + max) % max;
}
function nextSuggestion(step: number) {
  selectedSuggestionIndex = nextIndex(step, selectedSuggestionIndex, suggestions.length);
}
function nextChip(step: number) {
  selectedChipIndex = nextIndex(step, selectedChipIndex, items.length);
}
```

I think this way you can reduce the complexity of this handler a lot.

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+-->
+<div class="flex flex-col w-full gap-2">
+    <div class="flex {labelPosition === 'top' ? 'flex-col gap-2' : 'flex-row gap-4 items-center'}">
+        <Label for={"chips-input-" + label}>{label}</Label>
+        <div
+            id={"chips-input-" + label}
+            class="flex flex-wrap w-full items-center gap-2.5 px-4 {items.length === 0
+                ? 'py-2'
+                : 'py-1.5'}
+    text-default border-input-border-slate bg-background rounded-md border overflow-x-auto"
```
Slartibartfass2: The indentation here is a bit weird.
```suggestion
                text-default border-input-border-slate bg-background rounded-md border overflow-x-auto"
```

(I still prefer using `cn` for that, but that's up to you 😄)

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+                        : 'bg-slate-200'} rounded-full px-3 py-0.5 w-max"
+                >
+                    {item}
+                    <button
+                        class="ml-2 text-primary focus:outline-none"
+                        onclick={() => removeItem(index)}
+                    >
+                        &times;
+                    </button>
+                </div>
```
Slartibartfass2: Is there a reason why we don't use the [Badge component](https://next.shadcn-svelte.com/docs/components/badge) here?

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+            {/each}
+
+            <!-- input for next chip -->
+            <input
+                id={"chips-input-" + label}
```
Slartibartfass2: This input has the same ID as the div around it. AFAIK, that's a bad practice because IDs should be unique.

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,214 @@
...
+                    class="flex w-full justify-start {selectedSuggestionIndex === i
+                        ? 'bg-accent'
+                        : ''} text-default"
+                    onclick={() => addItem(suggestion)}
+                >
+                    {suggestion}
+                </Button>
+            {/each}
+        </ul>
+    {/if}
```
Slartibartfass2: I think it would look nice to put a border around that, so the suggestions don't float around.
Furthermore, there has to be maximum length otherwise it just grows endlessly

Here's an example:
image

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+    onMount(async () => {
+        try {
+            let possibleMembers: User[] = await BackendController.getInstance().getUsers();
+            // TODO: exchange by call to store or so, if login etc. is completely implemented
+            let thisUser: User = await BackendController.getInstance().thisUser().get();
+
+            possibleMemberEMails = possibleMembers
+                .filter((user) => user.id !== thisUser.id)
+                .map((user) => user.email)
+                .map((email) => email.toLowerCase());
```
Slartibartfass2: minor (optional)
```suggestion
                .map((user) => user.email.toLowerCase());
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+        } catch (error) {
+            // TODO: Question to reviewer: any better idea or further things to show?
+            // because the consequence is, no suggestions can be provided
+            console.log(`Could not get users from server (${error})`);
```
Slartibartfass2: I would use an alert to show that no suggestions could be loaded. I'm quite happy with how it turned out on the sign in/up pages.

You would store an `suggestionLoadError` as state, which would be set in the catch block. Further down, you can then add
```svelte
{#if suggestionLoadError}
    <Alert.Root variant="destructive">
        <CircleAlert class="size-4" />
        <Alert.Title>{suggestionLoadError}</Alert.Title>
        <Alert.Description>Please try again</Alert.Description>
    </Alert.Root>
{/if}
```

Also: use `console.error` to log errors.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+     *
+     * @param input the content of the input field, i.e. the search string
+     * @returns list of E-Mail (sorted) of users that can be invited
+     */
+    function filterPossibleMembers(input: string): string[] {
+        input = input.toLowerCase();
+
+        return possibleMemberEMails
+            .filter((item) => item.includes(input))
+            .sort((a, b) => distance(a, input) - distance(b, input));
```
Slartibartfass2: Can we make this more forgiving, e.g. by filtering with the Levenshtein distance with a maximum distance?

Small typos prevent the correct result from showing up, e.g. for "Eva" "eve.wilson@example.com" wasn't displayed.

As discussed, a fuzzy finder could be helpful. I tested [fzf](https://www.npmjs.com/package/fzf) and was happy with the result:
```ts
    function filterPossibleMembers(input: string): string[] {
        input = input.toLowerCase();

        const fzf = new Fzf(possibleMemberEMails);
        const results = fzf.find(input);

        return results.map(result => result.item);
    }
```

At this point we might also have to think about, if we also want to include the names of the users, in the case where the email doesn't match the name. Maybe a fuzzy find solution where the name and the email are concatenated would work.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+            errorOccurred = true;
+            console.log(`Could not create project (${error})`);
+            return;
+        }
+        // invite members (if necessary)
+        membersInput.forEach((email) =>
+            BackendController.getInstance().project(project.id).inviteUser(email),
+        );
+
+        await goto(`/project/${project.id}/dashboard`);
```
Slartibartfass2: Can we make this more interactive/provide additional feedback?

I think it would be great to see that the project is in creation and that the users are getting invited. When an error appears, an error message should be shown and only if everything succeeds, the page should navigate to the newly created project.

For the sign in we used this:
```ts
error = undefined;
loading = true;
await BackendController.getInstance()
    .signIn(email, password)
    .then(() => {
        goto("/");
    })
    .catch((signInError) => {
        error = signInError.message;
    });
loading = false;
```
```svelte
{#if loading}
    <Button type="submit" class="w-full" disabled>
        <LoaderCircle class="animate-spin" />
        Signing In
    </Button>
{:else}
    <Button type="submit" class="w-full">Sign In</Button>
{/if}
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+    }
+</script>
+
+<Dialog.Root bind:open>
+    <div class="px-5">
+        <!-- need to overwrite svg size in button, as the shadcn default button sets a default size
+                for possible icons -->
```
Slartibartfass2, line 184: Is this line break on purpose?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,151 @@
...
+    let open: boolean = $state(false);
+
+    let errorOccurred = $state(false);
+
+    let projectNameInput: Input;
+    let membersInput: string[] = $state([]);
+
+    // TODO: check, whether this solution scales as the backend contains hundreds / thousands of users
+    // list of possible members (represented by their E-Mail) that can be invited
+    let possibleMemberEMails: string[] = [];
```
Slartibartfass2: Otherwise, it would be `eMail` too
```suggestion
    let possibleMemberEmails: string[] = [];
```

Slartibartfass2 (changes requested):
Wow, I really like how this looks! I see that you put great effort into it.

Possible improvement can be seen in the suggestions and the process of creating the project in terms of feedback.

Some thoughts I still have:
- should be rather display the name of the users or maybe both?
- should we allow all sorts of inputs, even if it's not an email or a selected user?
- should we navigate to the new project right away?
  - I have an application in mind where you get asked first but I can't remember which

package.json
Merseleo: I moved it.

src/app.css
Merseleo: If the input is very long and especially longer than the width of the chips input, then a horizontal scrollbar is added. The default height of horizontal scrollbar is way bigger (on my system and browser I think 16px), as the width of a vertical scrollbar, so I set the width and height to a specific value which imo looks better. Do you want it bigger or smaller?

Bildschirmfoto 2025-01-22 um 10 09 27

src/app.css
Merseleo: You are right, it is an artifact from the development (the trigger button for the dialog for project creation used it, but now only use text-xl, which looks better imo). Thank you for the hint, I delete it.

src/app.css
Merseleo: I fixed this, but highlight-on-hover is used.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: 1) I didn't know about it until just now, but 2) it seems as the Badge component can only display things, but actually we want it to be clickable (for removing the chip) and highlightable. So I don't see the advantage of try to use this component.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: It was only for better readability, but I can delete it, if you want.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Not necessary, as I trim the input, before I pass it to this method, but I will change it, as your solution seems better to me.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: I will change that, didn't notice it, because Webstorm is smart enough to completely show the comment.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: > * sometimes you call `event.preventDefault()` and sometimes not, is there a reason for that?

Yes, the reason is, that sometimes (e. g. if I use the left or right arrow to navigate between the chips) I don't want to move the cursor I the current input as well) I want only my behavior and not the standard behavior in the input as well.

I consider your suggestions and try to reduce complexity. What do you think of the new key handler?

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: You are completely right, I will rename it.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: But what do you mean with the bind. I don't quite understand what the binding is supposed to be for? Can you please explain.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: I agree, I wasn't happy with the design too, thank you for your suggestion for the suggestions ;). I added
- a rounded border
- max height and vertical overflow, if more than 4 suggestions exist

What do you think of the new design?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: I agree, that a hint to the user would be better, I forgot this. I will add an alert. However, I don't like the solution to show the error object directly to the user, as this often contains an error message which is not understandable for the user. I think, we should add an error handler for such cases (see #173).

Thank you for the hint, I changed all `console.log` to `console.error`.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Of course we can, I didn't notice, that feedback could be needed, as this steps works on my machine instantly, but we don't correspond with a real server, but only a local one. So I agree, that we should consider this.

And (regarding your suggestion from the comment) I agree, that it would make sense to open another dialog after the project creation was successful, which asks the user, whether to open this project or not.

Merseleo (commented):
Thank you for the review and especially your ideas for improvement and better user experience. I look forward to your feedback to the new updates. 

Regarding your suggestion to show names and validate user input before invite new members, I try to find a solution.

Merseleo:
I extended the `ChipsInput` component by a validate function and a function for resolving "aliases", which allows to map input values to shown values. This means, no matter whether you type the name, autocomplete the suggestion or type the email, the name of the user is shown and saved (if it exist, which can be checked, with the validation function, which show a hint beneath the input field, if an error occurred). 
What do you think of my [solution](https://github.com/SE-UUlm/snowballr-frontend/pull/157/commits/92f856482779b69387ca0df47169fedcdf6ede1f)? Any ideas for improvement, is this how you imagined it?

src/lib/components/composites/input/ChipsInput.svelte
Slartibartfass2: Looks great now!

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Slartibartfass2: I agree with not showing the error message directly. This is fine now.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Slartibartfass2: I really like the workflow now.

src/lib/components/composites/input/ChipsInput.svelte
Slartibartfass2: In Svelte, you can bind a component to a variable. But this works only with Svelte components, not with HTML elements. This would look something like that:
```svelte
<script lang="ts">
  let input: Input;
  
  input.foo();
</script>

<Input bind:this={input} />
```
Initially, I thought it was possible to call focus on this, but you can only call exported functions.

TLDR; never mind


However, you renamed the ID of the input to `"input-" + label`.
When you want to focus, you call `document.getElementById("chips-input-" + label)?.focus();` still with the old ID.
I suggest storing the ID as a constant, so this won't happen again. The `Label` also refers to the old ID.

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,273 @@
...
+            if (resolveAlias) {
+                const resolvedAlias = resolveAlias(item);
+                if (resolvedAlias == undefined) {
+                    return;
+                }
+                item = resolvedAlias;
+            }
+            items = [...items, item];
+            inputText = "";
+        }
```
Slartibartfass2: I think, here's some potential to reduce indentation and make it more readable.
How about:
```suggestion
        const isValidNewItem = item.trim() !== "" && !items.includes(item);
        if (!isValidNewItem) {
            return;
        }

        const resolvedAlias = resolveAlias?.(item);
        if (resolvedAlias) {
            item = resolvedAlias;
        }

        items = [...items, item];
        inputText = "";
```
By negating the conditions to guard clauses, we can return early.

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,273 @@
...
+    import CircleAlert from "lucide-svelte/icons/circle-alert";
+
+    interface ChipsInputProps {
+        items: string[];
+        validate: (arg0: string) => boolean;
+        label?: string;
+        labelPosition?: "top" | "left";
+        searchSuggestions?: (arg0: string) => string[];
+        placeholder?: string;
+        resolveAlias?: (arg0: string) => string | undefined;
```
Slartibartfass2: What are your thoughts on renaming the argument names of these functions (`arg0`) to make them more descriptive?

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,273 @@
...
+        if (resetChipSelection) {
+            selectedChipIndex = -1;
+        }
+    }
+
+    /**
+     * Set the focus to the suggestions list and reset the {@link selectedChipIndex} to -1.
+     */
+    function focusSuggestionsList(): void {
+        document.getElementById("chips-suggestions-" + label)?.focus();
```
Slartibartfass2: Also, consider creating a variable for the ID and replacing it with this to keep it maintainable.

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,273 @@
...
+                    selectedChipIndex = items.length;
+                }
+                selectedChipIndex = nextIndex(
+                    event.key === "ArrowLeft" ? -1 : 1,
+                    selectedChipIndex,
+                    items.length,
+                );
+
+                focusInput();
+                break;
```
Slartibartfass2: As discussed, currently when the input field has text in it, we can't navigate the text, but instead the chips are navigated.

Instead, we want to navigate the chips only when the input text is empty; otherwise the text should be navigated.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,257 @@
...
+            .filter((item) => item.includes(input))
+            .sort((a, b) => distance(a, input) - distance(b, input));
+    }
+
+    /**
+     * Checks, whether a given input is a valid name or email of a registered user.
+     */
+    function validateInput(input: string): boolean {
+        return possibleNamesOrEmails.includes(input.trim());
+    }
```
Slartibartfass2: As discussed, we misunderstood each other. As initially planned in the design, we want to accept either existing users or email addresses of people that don't have an account yet.
The check for an email would be:
```ts
const emailSchema = zod.string().email();

...

const isValidEmail = emailSchema.safeParse(input).success;
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,257 @@
...
+                );
+            } else {
+                isErrorOnProjectCreation = true;
+                console.error("Could not create project (Project from server is undefined)");
+            }
+        } catch (error) {
+            isErrorOnProjectCreation = true;
+            console.error(`Could not create project (${error})`);
+            return;
+        }
```
Slartibartfass2: I'm not exactly sure why, but here when the project creation fails, it still continues to invite the users with then fails, obviously.

I solved this by using `then` and `catch`.
```suggestion
        await BackendController.getInstance()
            .createProject({
                name: projectNameInput.getValue(),
            })
            .then((project) => {
                Promise.all(
                    membersInput
                        .map(getEmailFromUserName)
                        .filter((email) => email !== undefined)
                        .map(BackendController.getInstance().project(project.id).inviteUser),
                )
                    .then(() => {
                        projectWasCreated = true;
                        open = false;
                    })
                    .catch((error) => {
                        isErrorOnProjectCreation = true;
                        console.error(`Could not invite users to project (${error})`);
                    });
            })
            .catch((error) => {
                isErrorOnProjectCreation = true;
                console.error(`Could not create project (${error})`);
            });
```

While testing, some questions appeared to me, which we should discuss:
- what should happen when the project creation succeeds but the inviting fails?
- what should happen when some invitations fail and others not?

What are your thoughts on that?

Slartibartfass2 (changes requested):
Great work there!

The workflow improved a lot.

A thing I also noticed that we might want to add: removing already added users from the suggestions, i.e., the user added a suggestion to the input and can't add the same user again.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Great suggestion, I will add it.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: I agree, that a better searching technique would be helpful. I will integrate *fzf*. But as I search in the possible members, I already consider both name and email. So does your comment relate to an old version?

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Thank you for the hint, you are right.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: I think failing invitations are not a problem of the project creation. We can create a project even without any invitations, so if any invitation fails, then we should display a hint in *Members* page in the settings or change the status from "invitation pending" to "invitation failed" or so? But I would add this for now and you can rebase on it and change it in your issue #16?!

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Already changed this, when creating a variable for the other ID.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: I can rename them, I thought this is convention, but of course, more readable code is always a good thing.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: I will change it.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Good catch, I updated the behavior and it is way better and more intuitive now.

Merseleo (commented):
Thank you for your ideas and the hints. I hope I didn't forgot any of your comments and everything works now.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Slartibartfass2: Yes, the comment was from the first review.

Same as for `fastest-levenshtein`, put `fzf` to the dev dependencies.
And now that you don't use `fastest-levenshtein` anymore, you can remove it.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Slartibartfass2: Fine for me

src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -0,0 +1,292 @@
...
+                <div
+                    class="flex items-center {index === selectedChipIndex
+                        ? 'bg-slate-300'
+                        : 'bg-slate-200'} rounded-full px-3 py-0.5 w-max"
+                >
+                    {#if displayItem !== undefined && displayItem(item) !== undefined}
+                        {displayItem(item)}
+                    {:else}
+                        {item}
+                    {/if}
```
Slartibartfass2: Minor:
If we provide an ID mapping as the default value for this method, we wouldn't have to do this condition.
```svelte
displayItem = (item) => item,
...
{displayItem(item)}
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,273 @@
...
+     *
+     * Furthermore, the filtered list is sorted by the score from the FZF algorithm, i.e.
+     * the members with the best matching name or email are at the beginning of the list (and will
+     * appear at the top of the suggestions list).
+     *
+     * @param input the content of the input field, i.e. the search string
+     * @returns list of "name \<email\>" (sorted) representations of users that can be invited
+     */
+    function filterPossibleMembers(input: string): string[] {
+        const fzf = new Fzf(possibleMembers.map((user) => `${getNames([user])} <${user.email}>`));
```
Slartibartfass2: We frequently call `getNames([user])`, what about `getNameAsString(user)`? And maybe we just rename that to `getName`.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,273 @@
...
+     * Checks, whether a given input is a valid name of a registered user or an email.
+     */
+    function validateInput(input: string): ValidationResult {
+        if (!Schema.email.safeParse(input.trim()).success) {
+            if (!possibleMembers.map((member) => getNames([member])).includes(input.trim())) {
+                return { success: false, error: "Please enter a valid name or email." };
+            }
+            if (
+                possibleMembers.filter((member) => getNames([member]) === input.trim()).length > 1
+            ) {
```
Slartibartfass2: Here, we basically run checks on the same result, so we can store the result before evaluating it.
```suggestion
            const matchingMembers = possibleMembers.filter(
                (member) => getNames([member]) === input.trim(),
            );
            if (matchingMembers.length === 0) {
                return { success: false, error: "Please enter a valid name or email." };
            }
            if (matchingMembers.length > 1) {
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,273 @@
...
+                return {
+                    success: false,
+                    error: "There are multiple users with this name. Please specify the user.",
+                };
+            }
+            return { success: true };
```
Slartibartfass2: Minor: This return statement is optional

Slartibartfass2 (changes requested):
As discussed, there were some minor misfunctions,

- when a project is created successfully, the chips input still has its values and the confirmation isn't shown anymore.
- also at the beginning when we type in a character, move in the suggestions and then just type an email adress, it won't be inserted as chip and an error is thrown
we could fix this by doing
```ts
            case "Tab":
            case "Enter":
            case ",":
            case "+":
                // add a new item
                event.preventDefault();
                if (suggestions.length > 0 && selectedSuggestionIndex !== -1) {
                    addItem(suggestions[selectedSuggestionIndex]);
                } else {
                    let validationResult = validate(inputText);
                    if (validationResult.success) {
                        addItem(inputText);
                    } else {
                        isInputValid = validationResult.success;
                        errorMessage = validationResult.error;
                    }
                }
                focusInput(true);
                break;

            case "ArrowDown":
            case "ArrowUp":
                if (suggestions.length === 0) {
                    break;
                }
                selectedSuggestionIndex = nextIndex(
                    event.key === "ArrowDown" ? 1 : -1,
                    selectedSuggestionIndex,
                    suggestions.length,
                );

                focusSuggestionsList();
                break;
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Good idea, I will integrate your suggestion.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: You are right, I will delete it.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Great idea, I will add it.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: You are right, we should use the function for only one object here. But I renamed to `getName`.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Yep, good catch.

Merseleo (commented):
Thank you for your suggestions and especially your attention and bugs to recognized. I hope I didn't forgot to fix anything and we are finished now. 

If so, I will rebase afterwards.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Get the name of a user identified by its email.
+     *
+     * @param input the email of the user
+     * @return the name of the user or undefined, if the no user with the given email was found
+     */
+    function mapEmailToName(input: string): string | undefined {
+        let name = getNames(initialPossibleMembers.filter((user) => input === user.email));
+        return name !== "" ? name : undefined;
```
Slartibartfass2: Here, we have to return the input, i.e., the email, when it couldn't be resolved to a name. This was introduced by setting the default value of `displayItem` to the ID function. Earlier we caught that case.
```suggestion
        return name !== "" ? name : input;
```

Slartibartfass2 (changes requested):
Just this small change.

Great work!

Slartibartfass2:
One other thing: Why did you rename the test directories? So far, they had the same names as the source directories.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: I will change it

Merseleo (commented):
I renamed them to make more clear, that we test components in this directories. You are right, that they are not called as its source directories anymore, but the source directories are subfolders of a `composite` / `components` folder.

Slartibartfass2 (approved):

Issue #38, opened 2024-10-31
[Use Case]: Filter Papers


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to filter the list of papers according to various properties.

### Acceptance criteria

- available filters are
  - stage
  - reviewer
  - publisher
  - year
  - decision
  - criterion
- the results are listed per stage
- papers that don't match the filter(s) are hidden
- if no paper matches the filter(s), a hint is shown that there are no results

### Anything else?

Related to #30

Extends #37

Issue #37, opened 2024-10-31
[Use Case]: Search For Papers


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to search for papers in the current project.

### Acceptance criteria

- provide search bar
- ability to search according to the paper title
- provide list of results per stage
- search results load while typing
- best matching results are shown on top of the results lists
- papers that do not match the search query are hidden
- if no paper matches the search query, a hint is shown that there are no results

### Anything else?

Related to #30

Is extended by #38 and #39

mowi12:
Removed highlighting of search result -> Will be handled in its own issue.

PR #158, opened 2025-01-11
View papers of each stage

### LINKED ISSUE #31: [Use Case]: View Papers Of Each Stage

Project Member

### Describe the use case

As project member, I want to view the papers of each stage in the project view.

### Acceptance criteria

- papers of different stages are separated
- papers of the same stage are grouped together
- the project member can hide all papers of a specific stage
- the project member can see of which stages the papers are hidden
- the project member can toggle the hidden status on and off
- for each paper, information about the review process and the final decision is shown

### Anything else?

Related to #30

Slartibartfass2:
Closes #31 

## What I have made

- first of all: don't worry 16 files are primitives and the customized select components are repetitive, so it's way less than you probably think :)
- added a reusable `Select` component with a generalized label and bindable selected values
- created a custom `Select` component for each filter
  - stages
  - reviewers
  - publishers
  - years
  - decisions
  - criteria
- all but the decision select take a promise that contains the available options 
- those filters are shown on the project papers page
- together with an accordion of stages with their respective papers
- this page is extended in various other issues, this is just the base

Use the following files to test it without the mock backend:
[mock-data.zip](https://github.com/user-attachments/files/18967665/mock-data.zip)
- put `+layout.ts` in `src/routes/project/[projectId]`
- put `+page.ts` in `src/routes/project/[projectId]/papers`

## Checklist

  - ~~[ ] unit tests~~ (no logic added that would require unit tests)
  - [x] integration tests


src/lib/components/primitives/select/select-trigger.svelte
```diff
@@ -0,0 +1,24 @@
...
+        ref = $bindable(null),
+        class: className,
+        children,
+        ...restProps
+    }: WithoutChild<SelectPrimitive.TriggerProps> = $props();
+</script>
+
+<SelectPrimitive.Trigger
+    class={cn(
+        "border-input bg-background ring-offset-background focus:ring-ring flex h-10 w-full items-center justify-between rounded-md border px-3 py-2 text-sm focus:outline-hidden focus:ring-2 focus:ring-offset-2 disabled:cursor-not-allowed disabled:opacity-50 [&>span]:line-clamp-1 gap-1.5",
```
Merseleo:
```suggestion
        "border-input bg-background flex h-10 w-full items-center justify-between rounded-md border px-3 py-2 text-sm focus:outline-hidden disabled:cursor-not-allowed disabled:opacity-50 [&>span]:line-clamp-1 gap-1.5",
```

src/lib/resource.svelte.ts
```diff
@@ -1,12 +1,51 @@
...
+ * @param options.onErrorValue - The value that is set when the promise rejects.
+ * @param options.ressourceName - The name of the resource that is used for error logging. Default is "resource".
+ * @returns The state that represents the loading resource.
+ */
 export const resource = <TPromise, TValue>(
     loadingResource: Promise<TPromise>,
     options: {
         initialValue: TValue;
-        onSuccess: (value: TPromise) => TValue;
+        onSuccess?: (value: TPromise) => TValue;
```
Merseleo, line 39: If `onSuccess` is optional you should update your comment, because this indicates that this field is mandatory.

src/lib/components/composites/select/Select.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+    export interface SelectOption {
+        value: string;
+        label: string;
+    }
+
+    interface Props {
+        options: SelectOption[];
+        categoryLabel?: string;
+        selectedValues?: string[];
```
Merseleo, line 12: Why it could be useful to pass selectedValues as properties? Is this for exposing the selected values?

src/lib/components/composites/select/Select.svelte
```diff
@@ -0,0 +1,59 @@
...
+        if (
+            !selectedValues ||
+            selectedValues.length === 0 ||
+            selectedValues.length === options.length
+        ) {
+            return `All ${categoryLabel} (${options.length})`;
+        }
+
+        return `${categoryLabel}: ${selectedValues.length} selected`;
+    }
```
Merseleo, line 28: I like the idea, especially when nothing or everything is selected. However in the Issue #38 we could discuss, whether it is possible and useful to display a preview of items selected instead of the number of selected items.

src/lib/components/composites/select/Select.svelte
```diff
@@ -0,0 +1,59 @@
...
+
+- when no options are available, a disabled item is displayed with the message "No [categoryLabel] available".
+- when all options are selected, the trigger displays "All [categoryLabel] ([options.length])".
+- when some options are selected, the trigger displays "[categoryLabel]: [selectedValues.length] selected".
+
+Usage:
+```svelte
+    <Select
+        {options}
+        allOptionsLabel="All Years"
```
Merseleo:
```suggestion
        categoryLabel="Years"
```

src/lib/components/composites/select/Select.svelte
Merseleo, line 1: Do you know, whether it is possible to remove the left padding / margin? 

Bildschirmfoto 2025-02-24 um 18 08 23

src/lib/components/composites/papers-view/CriteriaSelect.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+    import Select, { type SelectOption } from "$lib/components/composites/select/Select.svelte";
+    import { Criterion } from "$lib/model/api/criterion";
+    import { resource } from "$lib/resource.svelte";
+
+    interface Props {
+        loadingCriteria: Promise<Criterion[]>;
```
Merseleo, line 7: Actually we have the type `Criterion_List`. Do you think it makes sense to use it here?

src/lib/components/composites/papers-view/DecisionsSelect.svelte
```diff
@@ -0,0 +1,17 @@
...
+    interface Props {
+        selectedDecisions?: string[];
+    }
+
+    let { selectedDecisions = $bindable(undefined) }: Props = $props();
+
+    const options = $state<SelectOption[]>([
+        { value: "accepted", label: "Accepted" },
+        { value: "declined", label: "Declined" },
+        { value: "maybe", label: "Maybe" },
```
Merseleo: What do you think of the following suggestion: make the type of value a generic and use the enum values (I think from `PaperDecision`) here?

src/lib/components/composites/papers-view/YearsSelect.svelte
```diff
@@ -0,0 +1,40 @@
...
+        years.value
+            .slice()
+            .sort()
+            .map((year) => {
+                if (year === -1) {
+                    return {
+                        value: "-1",
+                        label: "N/A",
+                    };
+                }
```
Merseleo: How is this possible?

src/lib/components/composites/papers-view/StageEntry.svelte
```diff
@@ -0,0 +1,49 @@
...
+
+    let { projectId, stage, selectedPaper = $bindable(undefined) }: Props = $props();
+</script>
+
+<Accordion.Item value={`stage-${stage.stageIndex}`}>
+    <!-- TODO: Add amount of filtered/all e.g. (3/7) -->
+    <Accordion.Trigger data-testid="stage-entry-trigger">
+        <div class="flex flex-row w-full justify-between">
+            <span>Stage {stage.stageIndex}</span>
+            <span>({stage.papers.length} papers)</span>
```
Merseleo:
```suggestion
            <span>({stage.papers.length} {stage.papers.length > 1 ? "papers" : "paper"})</span>
```

Merseleo: Btw, great idea to show how many papers are in the stage.

src/lib/components/composites/papers-view/StageEntry.svelte
```diff
@@ -0,0 +1,49 @@
...
+        </div>
+    </Accordion.Trigger>
+    <Accordion.Content>
+        <div class="flex flex-col pl-5 gap-4">
+            {#each stage.papers as paper (paper.id)}
+                <PaperListEntry
+                    onClick={() => {
+                        selectedPaper = paper;
+                    }}
```
Merseleo, line 33: I think we want to deselect, if a paper is already selected and clicked again, what do you think? Because at the moment there is no possibility to close the overlay with paper information or did I miss something?

src/lib/components/composites/papers-view/StageEntry.svelte
```diff
@@ -0,0 +1,49 @@
...
+                    {paper}
+                    {projectId}
+                    showReviewStatus
```
Merseleo, line 36: Where this value is defined? Either you need to pass it or set it to a value?!

tests/integration/papers-view/criteria-select.test.ts
```diff
@@ -0,0 +1,69 @@
...
+                        description: "description1",
+                        category: CriterionCategory.EXCLUSION,
+                    },
+                    {
+                        id: "2",
+                        tag: "tag2",
+                        name: "name2",
+                        description: "description2",
+                        category: CriterionCategory.INCLUSION,
+                    },
```
Merseleo: I think you can make use of the `example-data.ts` (same applies to other custom select tests).

tests/integration/papers-view/criteria-select.test.ts
```diff
@@ -0,0 +1,69 @@
...
+                ]),
+                selectedCriteria: [],
+            },
+        });
+
+        await waitFor(() => {
+            const trigger = screen.getByText("All Criteria (2)");
+            expect(trigger).toBeInTheDocument();
```
Merseleo: Maybe also test for one text of one of the criteria (same applies to other custom select tests).

tests/integration/papers-view/decisions-select.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        render(DecisionsSelect, {
+            target: document.body,
+            props: {
+                selectedDecisions: [],
+            },
+        });
+
+        await waitFor(() => {
+            const trigger = screen.getByText("All Decisions (3)");
+            expect(trigger).toBeInTheDocument();
```
Merseleo: Maybe add expects for the three labels?

tests/integration/papers-view/stages-select.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        await waitFor(() => {
+            trigger = screen.getByText("All Stages (1)");
+            expect(trigger).toBeInTheDocument();
+        });
+        assert(trigger!);
+
+        await user.click(trigger);
+
+        const option = screen.getByText("Stage 0");
+        expect(option).toBeInTheDocument();
```
Merseleo, line 45: Maybe you can add an expect, that no empty hint is shown?

tests/integration/papers-view/stage-entry.test.ts
Merseleo, line 1: Maybe you should also add a test for selecting a paper (expected behavior: open / close the paper information on the right side)?

tests/integration/select/select.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        await user.click(trigger);
+
+        expect(trigger).toHaveAttribute("data-state", "open");
+        const option1 = screen.getByText("Option 1");
+        const option2 = screen.getByText("Option 2");
+        expect(option1).toBeInTheDocument();
+        expect(option2).toBeInTheDocument();
+    });
+
+    test("When option is clicked, then options are hidden and info is shown", async () => {
```
Merseleo: I dont really understand this test description.

src/routes/project/[projectId]/papers/+page.ts
```diff
@@ -0,0 +1,88 @@
...
+                .flatMap((projectPaper) => projectPaper.reviews)
+                .map((review) => review.userId);
+            const uniqueReviewerIds = new Set(allReviewerIds);
+            return members
+                .map((member) => member.user!)
+                .filter((user) => uniqueReviewerIds.has(user.id));
+        },
+    );
+    // attach noop-catch to handle promise rejection correctly (see https://svelte.dev/docs/kit/load#Streaming-with-promises)
+    loadingReviewers.catch(() => {});
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.ts]: Is this really necessary? Why not directly list all members as possible reviewers and if they never reviewed any paper, the list is empty, shouldn't be a problem?!

src/routes/project/[projectId]/papers/+page.ts
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @param papers - The papers of a project from different stages.
+ * @returns A list of stages with the papers of the respective stage.
+ */
+function organizePapersByStage(papers: Project_Paper[]): Stage[] {
+    const stages: Stage[] = [];
+    for (const paper of papers) {
+        let stage = stages.find((stage) => stage.stageIndex === paper.stage);
+        if (!stage) {
+            stage = { stageIndex: paper.stage, papers: [] };
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.ts:67]: Does this really work, if you add an empty stage here and add the paper to the newly created stage with `stage.papers.push(paper)` as stage is only locally defined, isn't it?

src/routes/project/[projectId]/papers/+page.ts
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.ts:1]: Same here, as in #195, I looks good to me, but I can't really test it.

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -1,17 +1,152 @@
...
+                        bind:selectedCriteria={papersFilters.criteria}
+                    />
+                    <Button
+                        onclick={() => {
+                            papersFilters = emptyFilters;
+                        }}
+                    >
+                        <Trash />
+                        Clear Filters
+                    </Button>
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.svelte]: I like the idea of clearing the filters (and maybe we should discuss these things when implementing the filter functionality), but what do you think of moving the clear button next to the filters button / search bar?

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -1,17 +1,152 @@
...
+                </div>
+            {/if}
+        </div>
+        <div class="w-full h-full">
+            {#await loadingStages}
+                <span class="text-hint">Loading stages...</span>
+            {:then stages}
+                <span class="text-hint">
+                    {`${stages.length} Stage${stages.length !== 1 ? "s" : ""}`}
+                </span>
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.svelte]: Do you think this is necessary or helpful? Because I doubt that there will be that many stages, that they dont fit on the screen.

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -1,17 +1,152 @@
...
+        </div>
+    </div>
+    {#if selectedPaper}
+        <Card.Root class="flex flex-col w-[60%] h-full gap-5 shadow-lg overflow-hidden">
+            <Card.Content>
+                <!-- TODO: replace with paper details card -->
+                <!-- This is done in https://github.com/SE-UUlm/snowballr-frontend/issues/219 -->
+                <pre>{JSON.stringify(selectedPaper.paper!, undefined, 2)}</pre>
+            </Card.Content>
+        </Card.Root>
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.svelte:153]: I think it would be cooler to animate this animate (with slide in and out), but you can also add this to the comment / issue and we try this when working on the this issue.

Merseleo (changes requested):
You did a great job, imo the papers page was well done and is nearly ready to be completed in #37, #38, and #219.

A general question: What do you think of moving the customized selects into the "select" folder?

src/lib/components/composites/papers-view/CriteriaSelect.svelte
Slartibartfass2: No, I deliberately destructured the `Criteria_List` type in the page load function so that I don't have to do that elsewhere.

src/lib/components/composites/papers-view/DecisionsSelect.svelte
Slartibartfass2: The value of the underlying `Select` component is always string.

src/lib/components/composites/papers-view/StageEntry.svelte
Slartibartfass2: Good catch. I added the `pluralize` method for that.

src/lib/components/composites/papers-view/StageEntry.svelte
Slartibartfass2: All this logic, I wanted to do in #219. Because we will also have to handle, when the user clicks anywhere else than the paper list entries, i.e., unselects them. Is it okay to do it there, or should I already do it here?

src/lib/components/composites/papers-view/StageEntry.svelte
Slartibartfass2: This is an abbreviation of `showReviewStatus={true}`. This is only possible for boolean values.
E.g., in native HTML, you can set `disabled` just like that.

src/lib/components/composites/papers-view/YearsSelect.svelte
Slartibartfass2: This was an artifact from the old API where the year could be undefined, and then I did set it to -1, so I can catch this case. Now, that this isn't the case anymore, I removed this check.

src/lib/components/composites/select/Select.svelte
Slartibartfass2: Yes, it's for binding the filters so that we can filter the papers.

src/lib/components/composites/select/Select.svelte
Slartibartfass2: I tried some options of what to display as a label and found this to be the best so far. What would be your idea to show a preview of the selected filters? How many should be shown, and what should be shown when there are too many?

src/lib/components/composites/select/Select.svelte
Slartibartfass2: Amazing!

src/lib/components/primitives/select/select-trigger.svelte
Slartibartfass2: Great suggestion, it looks better without the ring.

src/lib/resource.svelte.ts
Slartibartfass2: I'm not sure what you mean. It says "Default is the identity function."

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte]: When there are several hundred papers and the stages are uncollapsed, the user might not see all stages, so I thought this might be a small helpful hint.
Should we remove it?

src/routes/project/[projectId]/papers/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.ts:67]: The variable is declared before this block, so it's visible for `stage.papers.push`. In this line, we only add a value to the variable if it is undefined. As consequence, at `stage.papers.push` `stage` is definitely defined and can be used.

src/lib/components/composites/select/Select.svelte
Slartibartfass2: I added this change. Are you liking it now? We have to take care, that we don't want to use this somewhere else, especially with non-disabled items because then it would look weird.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte:153]: I added it to https://github.com/SE-UUlm/snowballr-frontend/issues/219

src/routes/project/[projectId]/papers/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.ts:1]: Can you test it with the mock data I provided?

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte]: I moved it next between the filter toggle button and the search bar. Looks better this way.

src/routes/project/[projectId]/papers/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.ts]: No, not really. I removed it.

tests/integration/papers-view/criteria-select.test.ts
Slartibartfass2: Good catch!

tests/integration/papers-view/criteria-select.test.ts
Slartibartfass2: Can you elaborate?

tests/integration/papers-view/decisions-select.test.ts
Slartibartfass2: Yes, I added those.

tests/integration/papers-view/stage-entry.test.ts
Slartibartfass2: I would like to handle this feature in https://github.com/SE-UUlm/snowballr-frontend/issues/219

tests/integration/papers-view/stages-select.test.ts
Slartibartfass2: Yes, added it.

tests/integration/select/select.test.ts
Slartibartfass2: Yes, it's a bit wonky. I changed it.

Slartibartfass2 (commented):
Thank you very much for this extensive review.
I updated the mock data.

src/lib/components/primitives/select/select-trigger.svelte
Merseleo: I think now we are done with importing all required shadcn components and dont need to delete rings anymore.

src/lib/resource.svelte.ts
Merseleo: Ah sorry, I missed this param comment. But I noticed something else: I think you always misspelled ‘resource’ there.

src/lib/components/composites/select/Select.svelte
Merseleo: Thank you, I read the documentation of svelte again and now it makes more sense.

src/lib/components/composites/select/Select.svelte
Merseleo: As I said, we could try to display a preview, so if the years "1998, 2000, 2014 and 2015" are selected, "1998, 2000, ...." could be displayed. But if you already tried several solution we can also leave it as it is. I would move this to the other issue anyway, just wanted to hear your opinion.

src/lib/components/composites/select/Select.svelte
Merseleo: Yes I like it way more now. And afaik we dont have any other place where we (want to) use these `Select` component.

src/lib/components/composites/papers-view/CriteriaSelect.svelte
Merseleo: Ok

src/lib/components/composites/papers-view/DecisionsSelect.svelte
Merseleo: Ah ok, I hope this will not cause any problems in the future.

src/lib/resource.svelte.ts
Merseleo: In general you "often" misspelled "ressource"?!

src/lib/components/composites/papers-view/YearsSelect.svelte
Merseleo: Ok.

src/lib/components/composites/papers-view/StageEntry.svelte
Merseleo: Great idea 👍

src/lib/components/composites/papers-view/StageEntry.svelte
Merseleo: No, it's ok. I read the issue description and all my thoughts are already listed in the acceptance criteria.

src/lib/components/composites/papers-view/StageEntry.svelte
Merseleo: Of course, sorry.

tests/integration/papers-view/criteria-select.test.ts
Merseleo: In this case:
```typescript
const criterion = screen.getByText("Relevance");
expect(criterion).toBeInTheDocument();
```

Or do you think these test cases are useless?

tests/integration/papers-view/stage-entry.test.ts
Merseleo: Agreed.

tests/integration/select/select.test.ts
Merseleo: Better, thank you.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte:153]: Perfect.

src/routes/project/[projectId]/papers/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.ts]: Ok let's test it this way. If we recognize, that your filtering would be advantageous, we can still reintegrate your commit.

src/routes/project/[projectId]/papers/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.ts:67]: Ok, luckily this wasn't a question in the Web Eng exam ;)

src/routes/project/[projectId]/papers/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.ts:1]: Yes, with your provided mock data everything works fine. So I clicked around like a stupid user and everything worked and I don't think any data was missing.

One thing I noticed now after adding more demo papers: The list becomes "very" long and is not scrollable inside the accordion, but the entire page becomes scrollable. Is this behavior we want?

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte]: I agree

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte]: No, i didnt think of your scenario and in this case it can make sense. So lets keep it for now.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -39,8 +40,10 @@
                 return "border-decline-red";
             case PaperDecision.UNDECIDED:
                 return numberOfReviews > 0 ? "border-maybe-yellow" : "border-unreviewed-gray";
-            default:
+            case PaperDecision.UNSPECIFIED:
                 return "border-unreviewed-gray";
+            default:
+                exhaustiveCheck(decisionStatus);
```
Merseleo, line 46: Do you want to add these checks in the other switch statements as well? For example in the Paper view?

src/lib/utils/common-helper.ts
```diff
@@ -59,10 +59,22 @@ function doesPaperNeedReview(paper: Project_Paper, numberOfRequiredReviews: numb
...
 }
 
-export { getName, getNames, isPaperUndecided, doesPaperNeedReview, exhaustiveCheck };
+/**
+ * Pluralize a word based on the count.
+ *
+ * @param count - the number of items
+ * @param singular - the singular form of the word
+ * @param plural - the plural form of the word
+ * @returns the pluralized word
```
Merseleo: Imo something like "the pluralized word, if count is not equal to one, singular otherwise" would be more correct?

Merseleo (changes requested):
Ok I think there only some discussions left and spelling mistakes, but the big part seems to me to be done. Great work!

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: I'm not sure where you mean. I searched for all switch statements and checked whether they should be exhaustive, and, if so, used this check. Can you tell me the statements I missed?

src/lib/components/composites/select/Select.svelte
Slartibartfass2: Yes, I had that in the first place, but the select button grows too wide when the user selects many options.

src/lib/resource.svelte.ts
Slartibartfass2: Ah, yes. I fixed it.

src/lib/utils/common-helper.ts
Slartibartfass2: Yes, I fixed it.

src/routes/project/[projectId]/papers/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.ts:1]: I think, yes. Otherwise, we would have two scroll bars, one for the stage papers and the other one for all the stages. In this case, I think it's better to have just one.

tests/integration/papers-view/criteria-select.test.ts
Slartibartfass2: That the options are visible when the trigger is clicked, is tested in `select.test.ts`, so I think it's fine to not test it again here.

src/lib/components/composites/papers-view/DecisionsSelect.svelte
Slartibartfass2: When filtering, we need to map it back to the enum value, but that shouldn't be a problem.

Slartibartfass2 (commented):
Thank you for always taking so much time to review my code.

src/lib/resource.svelte.ts
Merseleo: I thought you wanted to include a spell checker ;)?

src/lib/components/composites/select/Select.svelte
Merseleo: Ok, let's discuss this in #38. Maybe I need to try my idea to see that it doesn't work as I image. But good to know, that you already tried some things.

tests/integration/papers-view/criteria-select.test.ts
Merseleo: Ah, you are right. Forgot about these tests. So I agree, its fine.

src/routes/project/[projectId]/papers/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.ts:1]: Ok, wasn't really sure, whether this is normal when using accordions.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: I think in the `PaperDecisionBanner` component this check can be used.

Merseleo (changes requested):
We are nearly finished!

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: Yes, now after the rebase, we can do that.

Merseleo (approved):

PR #175, opened 2025-01-22
Integrate gRPC API into frontend

### LINKED ISSUE #174: [Feature]: Integration of the gRPC API

_No response_

At the moment, we use a REST API to communicate with the (mock) backend, which is heavily inspired by the old API. 
But as we agreed to use change to the new [gRPC API](https://github.com/SE-UUlm/snowballr-api), the controller for the REST API are obsolete 
and we need to create the new gRPC functions and integrate them into the current frontend.

We need to generate the gRPC functions from the API definition and integrate the new functions and types into the current frontend.
Therefore, we need to 
- delete the http controller
- exchange the types by the new types
- exchange old calls to the http controllers by the new corresponding function

_No response_

### Anything else?

_No response_

Merseleo:
Closes #174

## What I have made

- introduce new gRPC API and exchange all existing calls to the API by the new calls

If problems regarding the mock backend server occur (when testing the API integration), try to use other ports.
The following configuration worked for me without any problems:
- PUBLIC_API_BASE_URL=http://localhost:8080
- Frontend started on http://localhost:5173
- changed port (internal and exposed) of mock backend from 8080 (or 8081 for exposed port) to 9090 in `compose.yml`, `main.ts` and `envoy.yml`
- changed the name of cluster (and not address!) from *snowballr-mock-backend* to *snowballr-service* in the `envoy.yml`

Furthermore, you need to delete the `authInterceptor` in the main.ts, so no authentication is required, as this is currently not working in the frontend.

> **Hint**: The old REST API was not deleted, but is planned in #190

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (the API integration can not be tested directly)
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - ~~[ ] end-to-end tests~~


Merseleo:
One question to the reviewer: What is your opinion on renaming the `model/` directory that currently contains all types and interfaces used in the frontend to something better? Or is model a common name for this?

.github/workflows/setup/action.yml
```diff
@@ -15,3 +15,7 @@ runs:
         - name: Install dependencies
           shell: bash
           run: npm ci
+
+        - name: Generate gRPC web client code
+          shell: bash
+          run: npm run compile:proto
```
Slartibartfass2, line 21: The `README.md` should contain the following information:
- the user has to set up the git submodule
- the user has to run this command

Slartibartfass2: Similar to what I wrote [here](https://github.com/SE-UUlm/snowballr-mock-backend/pull/2#issuecomment-2642337425). Let us add a "How I get started" section.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "api"]
+	path = api
+	url = git@github.com:SE-UUlm/snowballr-api.git
+	branch = develop
```
Slartibartfass2: There's no `develop` branch anymore. Let's use the `v0.1.0` tag.

.prettierignore
```diff
@@ -2,3 +2,6 @@
 package-lock.json
 pnpm-lock.yaml
 yarn.lock
+
+# API
+api/
```
Slartibartfass2: This both matches the submodule and the directory where the generated code goes to. I'm not sure why.
Do you think it makes more sense to make it more explicit by configuring it like this:
```suggestion
api/**
src/lib/model/api/**
```

eslint.config.js
```diff
@@ -28,6 +28,13 @@ export default ts.config(
...
     },
     {
-        ignores: ["build/", ".svelte-kit/", "dist/", "coverage/", "node_modules/"],
+        ignores: [
+            "build/",
+            ".svelte-kit/",
+            "dist/",
+            "coverage/",
+            "node_modules/",
+            "src/lib/model/api/",
```
Slartibartfass2, line 37: While we currently don't have any files in `api/` to lint, we should add it for the future.
```suggestion
            "src/lib/model/api/",
            "api/",
```

package.json
```diff
@@ -20,10 +20,13 @@
         "test:e2e": "playwright test",
         "test:and-open-coverage": "npm run test && npm run show-coverage",
         "license-check": "license-checker --onlyAllow 'GPL-3.0-only;GPL-3.0-or-later;GPL-2.0-or-later;LGPL-3.0-only;LGPL-3.0-or-later;LGPL-2.1-only;LGPL-2.1-or-later;AGPL-3.0-only;AGPL-3.0-or-later;Apache-2.0;Artistic-2.0;ClArtistic;BSL-1.0;BSD-3-Clause;CC0-1.0;CECILL-2.0;BSD-3-Clause-Clear;ECL-2.0;EFL-2.0;EUDatagrid;MIT;BSD-2-Clause;FTL;HPND;iMatix;Imlib2;IJG;Intel;ISC;MPL-2.0;NCSA;OLDAP-2.7;Python-2.0;Ruby;SGI-B-2.0;SMLNJ;UPL-1.0;Unlicense;Vim;W3C;WTFPL;X11;XFree86-1.1;Zlib;ZPL-2.0;ZPL-2.1;CC-BY-4.0;BlueOak-1.0.0;0BSD' --excludePackages 'spdx-exceptions@2.5.0' --summary",
-        "lighthouse": "lighthouse --config-path .github/lighthouse.config.js --chrome-flags=\"--headless\" --output-path=./lighthouse-reports/report.html --view"
+        "lighthouse": "lighthouse --config-path .github/lighthouse.config.js --chrome-flags=\"--headless\" --output-path=./lighthouse-reports/report.html --view",
+        "compile:proto": "npx protoc --ts_out src/lib/model/api --ts_opt generate_dependencies --proto_path api/proto/ api/proto/main.proto"
```
Slartibartfass2: Minor: `npx` is not necessary here, as it already calls the command in the node environment.
```suggestion
        "compile:proto": "protoc --ts_out src/lib/model/api --ts_opt generate_dependencies --proto_path api/proto/ api/proto/main.proto"
```

package.json
```diff
@@ -20,10 +20,13 @@
...
     },
     "devDependencies": {
         "@playwright/test": "^1.49.1",
+        "@protobuf-ts/grpcweb-transport": "^2.9.4",
+        "@protobuf-ts/plugin": "^2.9.4",
```
Slartibartfass2, line 29: All `@protobuf-ts` dependencies have the same version. Let's add a Dependabot group for that.
For that, add
```yaml
          protobuf-ts:
              patterns:
                  - "@protobuf-ts/*"
```
under `groups` in `.github/dependabot.yaml`

src/lib/grpc-api.ts
```diff
@@ -0,0 +1,9 @@
+import { GrpcWebFetchTransport } from "@protobuf-ts/grpcweb-transport";
+import { SnowballRClient } from "./model/api/main.client";
+import { PUBLIC_API_BASE_URL } from "$env/static/public";
+
+const transport = new GrpcWebFetchTransport({
+    baseUrl: PUBLIC_API_BASE_URL,
+});
+
+export const BACKEND = new SnowballRClient(transport);
```
Slartibartfass2: We talked about naming constants. Shouldn't this be called `backend` as it's not a deeply immutable constant?

src/routes/(auth)/signup/+page.svelte
```diff
@@ -23,17 +26,33 @@
...
+            .then((user) => {
+                const { accessToken, refreshToken } = user.response;
+                // TODO: Login with accessToken, store both tokens and redirect to the home page,
+                // will be completed in #134
+                console.log(
+                    `New user was registered with the following token: ${accessToken} (refreshToken: ${refreshToken})`,
+                );
+            })
+            .catch((error) => {
+                if (error.code == "ALREADY_EXISTS") {
```
Slartibartfass2: Are these codes defined somewhere, or should we do that ourselves?

src/routes/(auth)/signup/+page.svelte
```diff
@@ -23,17 +26,33 @@
...
+                    registrationError = {
+                        errorTitle: "An account with this email address already exists.",
+                        errorDetails: "Try logging in or resetting your password",
+                    };
+                    return;
+                }
+                registrationError = { errorTitle: "Something went wrong while registration." };
+                console.error(error);
```
Slartibartfass2: Shouldn't we log the other error as well?
```suggestion
                    registrationError = {
                        errorTitle: "An account with this email address already exists.",
                        errorDetails: "Try logging in or resetting your password",
                    };
                } else {
                    registrationError = { errorTitle: "Something went wrong while registration." };
                }
                console.error(error);
```

src/routes/(auth)/signup/+page.svelte
```diff
@@ -84,6 +103,12 @@
             />
             <PasswordInput class="w-full" bind:this={passwordInput} />
             <Button type="submit" class="w-full">Create an account</Button>
+            {#if registrationError}
+                <ErrorAlert
+                    errorTitle={registrationError.errorTitle}
+                    errorDetails={registrationError.errorDetails}
+                />
+            {/if}
```
Slartibartfass2, line 113: We often use the same pattern as this. What do you think about only passing the error object to the component and when it is defined, we display the alert?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -148,16 +147,20 @@
...
 
                 return Promise.all(
-                    membersInput.map((member) =>
-                        BackendController.getInstance().project(projectId!).inviteUser(member),
+                    membersInput.map(
+                        (member) =>
+                            BACKEND.inviteUserToProject({
+                                projectId: projectId!,
+                                userEmail: member,
+                            }).response,
```
Slartibartfass2: `member` is a bit misleading, let's give it a more descriptive name, e.g., `memberEmail` or just `email`.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -77,14 +97,16 @@ Usage:
...
+            ? `border-l-4 ${getReviewDecisionColor(projectPaper.decision, projectPaper.reviews.length)}`
             : ''} rounded-md px-3 py-1.5"
     >
-        <PaperInfo {paper} />
+        <PaperInfo paper={projectPaper.paper!} />
     </div>
-    {#if paper.reviewData !== undefined && showReviewStatus}
-        {#each paper.reviewData.reviews as review}
-            <UserAvatar user={review.user} reviewDecision={review.decision} />
+    {#if projectPaper.reviews.length !== 0 && showReviewStatus}
```
Slartibartfass2: `projectPaper.reviews.length !== 0` is not necessary because the `#each` block doesn't get executed when there are no reviews
```suggestion
    {#if showReviewStatus}
```

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -27,30 +21,37 @@ Furthermore this component is clickable and navigates to the corresponding proje
 
 Usage:
 ```svelte
-    <ProjectListEntry project={demoProject} members={memberUserSpecArray} stage={3} stageProgress={70} />
+    <ProjectListEntry project={demoProject} members={memberUserSpecArray} statistics={{projectProgress: 0.3}} />
```
Slartibartfass2: This name changed
```suggestion
    <ProjectListEntry project={demoProject} membersList={memberUserSpecArray} statistics={{projectProgress: 0.3}} />
```

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -27,30 +21,37 @@ Furthermore this component is clickable and navigates to the corresponding proje
...
-        {#if members.length > 0}
-            <span class="truncate text-hint">{getNames(members)}</span>
+        {#if membersList.members.length > 0}
+            <span class="truncate text-hint"
+                >{getNames(
+                    membersList.members
+                        .map((member) => member.user)
+                        .filter((user) => user !== undefined),
+                )}</span
+            >
```
Slartibartfass2: Here, the automatic formatting looks weird again.
```suggestion
            <span class="truncate text-hint">
                {getNames(
                    membersList.members
                        .map((member) => member.user)
                        .filter((user) => user !== undefined),
                )}
            </span>
```

We could possibly destruct the `membersList` object to get the `members` directly.

src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -30,12 +32,12 @@ in the PaperEntry component.
...
-    <UserAvatar user={{
-        ...
+    <UserAvatar userId={{
+        id: "1",
+        email: "john@doe.com",
         firstName: "John",
         lastName: "Doe",
-        ...
-    }} reviewDecision: ReviewDecision.Accept />
+    }} reviewDecision={ReviewDecision.ACCEPT} />
```
Slartibartfass2:
```suggestion
    }} reviewDecision={ReviewDecision.ACCEPTED} />
```

src/lib/model/general.ts
```diff
@@ -1,3 +1,28 @@
...
+    Project_Member_List,
+    Project_Paper,
+    Project_Statistics,
+} from "$lib/model/api/project";
+
 type ValidationResult = { success: true } | { success: false; error: string };
 
-export type { ValidationResult };
+// TODO: find better name
+interface ApiLoadError {
```
Slartibartfass2: `ApiError`, `BackendError`, `RequestError`?

Slartibartfass2 (changes requested):
This looks great so far. I see that a lot of code is now shorter.

There are some types that are unnecessarily doubly nested, e.g., the members list. But it seems like we have to get used to it.

Slartibartfass2:
> One question to the reviewer: What is your opinion on renaming the `model/` directory that currently contains all types and interfaces used in the frontend to something better? Or is model a common name for this?

I think it doesn't fit into `model` as it also contains logic. Maybe let's just put it under `lib` next to `model`?

Merseleo:
> > One question to the reviewer: What is your opinion on renaming the `model/` directory that currently contains all types and interfaces used in the frontend to something better? Or is model a common name for this?
> 
> I think it doesn't fit into `model` as it also contains logic. Maybe let's just put it under `lib` next to `model`?

What kind of logic does it contain? So I do not really your suggestion: `src/lib/lib/` and `src/lib/model`? And what should contain which files?

.github/workflows/setup/action.yml
Merseleo: I can't quite believe this is happening to me right now: I get the hint that something is not sufficiently documented ^^. But of course you're right, I'll add it.

.gitmodules
Merseleo: What do you think of the main branch instead? Because as soon as we add new things (and I need to very soon) we need to update the version and then we have to update the tag number as well in this file, which could be a bit annoying. And the main branch should contain the latest, stable API anyway?!

Slartibartfass2:
> > > One question to the reviewer: What is your opinion on renaming the `model/` directory that currently contains all types and interfaces used in the frontend to something better? Or is model a common name for this?
> > 
> > 
> > I think it doesn't fit into `model` as it also contains logic. Maybe let's just put it under `lib` next to `model`?
> 
> What kind of logic does it contain? So I do not really your suggestion: `src/lib/lib/` and `src/lib/model`? And what should contain which files?

Sorry, my answer was confusing. I meant putting the `api` directory under `lib` because it contains not only model types but logic as well and `model` doesn't fit, in my opinion.

.github/workflows/setup/action.yml
Merseleo: What do you think of the new README.md version?

.prettierignore
Merseleo: Yes, probably it makes more sense to explicitly state, that both paths should be ignored, I add it.

eslint.config.js
Merseleo: I agree.

package.json
Merseleo: Ok

package.json
Merseleo: Ok, i am not sure, whether these versions really correlate or whether it is coincidence, but I can add it.

src/lib/grpc-api.ts
Merseleo: To what extent can it be changed? For me, such a client is an example of an immutable variable?!

src/routes/(auth)/signup/+page.svelte
Merseleo: Actually I thought we can the types (so `GrpcStatusCode.ALREADY_EXISTS` or something similar), but it seems that error.code returns a string. 
But back to your question: Yes the possible error codes are defined, for example [here](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

src/routes/(auth)/signup/+page.svelte
Merseleo: Probably we could, yep. I didn't add it, as it will be displayed, but yes we should log every error as well.

src/routes/(auth)/signup/+page.svelte
Merseleo: I saw this pattern in your code and we can also do it this way (and I also tried it), but imo it makes the code more readable this way, if you explicitly provide the error title or details. But if you doesn't like this way, I will change it.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Good idea.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: You are right, I think I only updated the old code, where this check was still necessary. Thank you for this hint.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: Good catch.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: I don't think it is the effort worth.

src/lib/model/general.ts
Merseleo: ApiError it is

.gitmodules
Slartibartfass2: While this is all true, it will cause problems when the branch updates and forces the frontend to react to the changes.
Imagine you work on a branch, push your changes and the CI will test your code. When the main branch of the API is updated at the same time, your CI run will probably fail, and you might wonder what happened.
Therefore, I think we should make controlled updates to prevent such cases. This makes the whole code base more stable because a single commit in the code base will always work with a specific API version.

Using always, the main branch can lead to unwanted behavior because the submodule can and will change independently.
It's comparable to using `latest` as version tag in a docker image, which is a big NO NO.

package.json
Slartibartfass2: They do, I checked out the repo that maintains those packages.
See https://github.com/timostamm/protobuf-ts/tree/main/packages

src/lib/grpc-api.ts
Slartibartfass2: There are some mutable properties. Nonetheless, I think a class is never really a constant, but I think we can argue about that.

src/routes/(auth)/signup/+page.svelte
Slartibartfass2: That should be a problem. we could define an enum similar to [this](https://github.com/flowr-analysis/flowr/blob/05431cdb498928fe65d87297acb7be0664c90130/src/r-bridge/lang-4.x/ast/model/type.ts#L159) and then check `error.code === GrpcStatusCode.ALREADY_EXISTS` (or `AlreadyExists`).

.github/workflows/setup/action.yml
Slartibartfass2: It looks great. I think later we should move the testing and general contribution guidlines to the wiki.

Is there a reason why you made "SnowballR's use cases" h3 and not h2 as all the other sections?

src/routes/(auth)/signup/+page.svelte
Slartibartfass2: I agree with explicitly providing the error title. For me, it's just the `{#if ...}` block that feels repetitive. IMO, this should be inside the component.

Slartibartfass2 (changes requested):
Basically done, just some open things to discuss.

What's your decision on the location of the `api` directory?

.github/workflows/setup/action.yml
Merseleo: I agree, especially as we must add E2E tests soon, we should document the testing procedure etc. in the wiki.

Nope, my mistake.

src/lib/grpc-api.ts
Merseleo: No, the argument with the mutable properties is a good one, I will refactor it.

src/routes/(auth)/signup/+page.svelte
Merseleo: I can add this.

src/routes/(auth)/signup/+page.svelte
Merseleo: So that the component is always called and depending on the error, it displays something or nothing? 
That seems very confusing to me, as it looks like you always provide an error alert imo.

.gitmodules
Merseleo: And still often used to prevent to old code and make permanent updates, which have their advantages. But I see the point, that it cause problems and we can also handle this by explicitly setting a new version tag, when we introduce a new feature in the frontend. We will go with your solution, thank you for the explanation!

Merseleo: However, we can not set this automatically, as it seems impossible to specify a tag in the .gitmodules file

Merseleo (commented):
Now I understand your comment on the model and api stuff and what you suggest. But imo this would be even more confusing to have to api folders relatively close to each other. And yes it contains 'logic', but not really and mainly types.
In the meantime, I don't think the place is so bad after all. Are you ok with this?

src/lib/model/api/error-codes.ts
Slartibartfass2, line 1: This looks good. Now that we put it in the `api` directory, it won't get checked by our formatter or linter. We either would've to include this back again or just move the file out of this directory and keep `api` for generated files only.

src/routes/(auth)/signup/+page.svelte
Slartibartfass2: I see your point. Let's keep it like that.

.gitmodules
Slartibartfass2: I think you can do it like [that](https://stackoverflow.com/questions/1777854/how-can-i-specify-a-branch-tag-when-adding-a-git-submodule/1778247#1778247)

Slartibartfass2 (changes requested):
NEARLY DONE

src/lib/model/api/error-codes.ts
Merseleo: You are right, I moved it one folder up. I think it is better to only store the generated code in the `/api` directory.

.gitmodules
Merseleo: If you prefer this solution, I can also use this solution.

Slartibartfass2 (approved):

PR #183, opened 2025-02-01
Use reusable setup action

### LINKED ISSUE #182: Use npm ci in the CI
Currently, we use `npm install` in the CI to fetch the dependencies.
According to [this](https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci) `npm ci` is more suited for the CI as it checks whether the package-lock file matches the package file and throws an error if that's not the case.

Slartibartfass2:
Closes #182 
## What I have made

- Replaced `npm install` with `npm ci` (see issue description)
- Add reusable setup action
  - Note that checkout needs to be separate so that the workflow file can be found

## Checklist

  - ~~[ ] unit tests~~ (only CI changes)
  - ~~[ ] integration tests~~ (only CI changes)
  - ~~[ ] end-to-end tests~~ (only CI changes)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
                   node-version: ${{ env.node_version }}
 
             - name: Install dependencies
-              run: npm install
+              run: npm ci
 
             - name: Build application
               run: npm run build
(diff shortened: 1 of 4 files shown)


Merseleo (approved):
LGTM.
I will add the command for generating the gRPC front-end code in PR.

Issue #43, opened 2024-10-31
[Use Case]: Manually Add References And Citations


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to manually add references and citations to a specific paper.

### Acceptance criteria

- only papers that are not already in the references can be added to the references
- only papers that are not already in the citations can be added to the citations
- the papers can be added by using their paper ID

### Anything else?

Related to #34 and #42

PR #184, opened 2025-02-02
Add paper references and citations to paper view

### LINKED ISSUE #42: [Use Case]: View Paper References And Citations

Project Member

### Describe the use case

As project member, I want to view the references and citations of a specific paper.

### Acceptance criteria

- list the papers, which reference the current paper
- list the papers, which are cited by the current paper
- both lists should be separated and labeled

### Anything else?

Related to #34 and #43

Is extended by #44

Slartibartfass2:
Closes #42 

## What I have made

- `PaperListEntry` can now also take just a `Paper` for when review data isn't shown
- `NamedList` can now take a `preListContent` snippet, which is displayed below the title but above the list itself
  - this is used for the search bar, but may also be used for descriptions or other information
- added `ReferencesAndCitationsCardContent` that displays references and citations in the paper view
  - both lists are searchable and scrollable in their respective half of the card

## Checklist

  - [x] integration tests


src/lib/components/composites/list/NamedList.svelte
```diff
@@ -49,14 +51,18 @@ If the option showNumberOfListItems is set to true (default: false),
...
 or an optional hint (provided with 'emptyHint') that the list is empty.
 Otherwise the error message is shown.
+
+You can render additional content between the title and the list by providing `preListContent`.
+This can be e.g. a search bar.
 -->
 <div class="flex flex-col h-full w-full px-5 gap-y-5 overflow-hidden">
     {#await items}
         <h2>{listName}</h2>
+        {@render preListContent?.()}
```
Merseleo, line 67: Is it clever to always render the preListContent, even while loading the list data or in the error case? What do you think of the idea of only rendering this content, if the list could be successfully be loaded? Only want to think about this idea, but we can also drop it.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -11,12 +11,19 @@
...
 
-    const navigateToPaperView = () => goto(`/project/${projectId}/paper/${projectPaper.id}`);
+    const navigateToPaperView = () => {
+        const paperId = isProjectPaper(paper) ? paper.paper!.id : paper.id;
+        const paperLink =
+            projectId !== undefined
+                ? `/project/${projectId}/paper/${paperId}`
+                : `/paper/${paperId}`;
+        goto(paperLink);
+    };
```
Merseleo: Very thoughtful thanks, I hadn't even thought of that.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -11,12 +11,19 @@
         onClick?: () => void;
     };
 
-    const navigateToPaperView = () => goto(`/project/${projectId}/paper/${projectPaper.id}`);
+    const navigateToPaperView = () => {
+        const paperId = isProjectPaper(paper) ? paper.paper!.id : paper.id;
```
Merseleo: One idea (could be reused when creating the PaperInfo as well): What do you think of converting `isProjectPaper` to `asPaper` which gets a paper or project paper as now and return the paper either directly or extract it from the project paper? Maybe this could make the code more readable?

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -11,12 +11,19 @@
...
+                : `/paper/${paperId}`;
+        goto(paperLink);
+    };
 
     const {
-        projectPaper,
+        paper,
         projectId,
-        showReviewStatus = false,
+        showReviewStatus,
```
Merseleo, line 26: Why did you delete this? Is the default automatically `false`?

src/lib/model/general.ts
```diff
@@ -18,10 +19,21 @@ interface ProjectListEntryInterface {
     statistics: Project_Statistics;
 }
 
-interface PaperListEntryInterface {
-    projectPaper: Project_Paper;
-    projectId: string;
-    showReviewStatus?: boolean;
+type PaperListEntryInterface =
```
Merseleo, line 29: Maybe it could be helpful to have a comment, why this type has two different interfaces ? 
Something like "...the interface for either a paper or a project paper?". Or do you think this would be confusing?

Merseleo: But I like the renaming to paper (instead of my projectPaper)

src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -5,13 +5,13 @@
     interface Props {
         placeholderText?: string;
         onSearch: (searchText: string) => void;
+        timeoutInMs?: number;
     }
-    const { placeholderText = "Search", onSearch }: Props = $props();
+
+    const { placeholderText = "Search", onSearch, timeoutInMs = 500 }: Props = $props();
```
Merseleo, line 11: If you make it a prop (good idea btw), then it would be helpful to add in the usage example below.

tests/integration/paper-components/paper-list-entry.test.ts
```diff
@@ -91,14 +86,8 @@ describe("PaperListEntryComponent", () => {
...
-                    id: "0",
-                    paper: Papers.demoPaper1,
-                    stage: 0n,
-                    decision: PaperDecision.UNDECIDED,
-                    reviews: [Reviews.demoReview1],
-                },
-                projectId: "0",
+                paper: Papers.demoPaper1,
+                projectId: undefined,
                 showReviewStatus: false,
```
Merseleo, line 91: Actually this test should test, that no review information are shown although it is a project paper and information exist. So I think you should actually not provide a paper, but a project paper.

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        const paper1 = screen.getByText(Papers.demoPaper1.title, { exact: false });
+        expect(paper1).toBeInTheDocument();
+        const paper2 = screen.getByText(Papers.demoPaper2.title, { exact: false });
+        expect(paper2).toBeInTheDocument();
+
+        const searchBars = screen.getAllByTestId("search-bar-input");
+        expect(searchBars).toHaveLength(2);
+    });
+
+    test("When references are searched, then only search results are shown", async () => {
```
Merseleo:
```suggestion
    test("When references are searched, then only matching reference papers are shown", async () => {
```

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        const user = userEvent.setup();
+        render(ReferencesAndCitationsCardContent, {
+            target: document.body,
+            props: {
+                backwardReferencedPapers: Promise.resolve([Papers.demoPaper1, Papers.demoPaper2]),
+                forwardReferencedPapers: Promise.resolve([]),
+            },
+        });
+
+        expect(Papers.demoPaper1.title).not.toEqual(Papers.demoPaper2.title);
```
Merseleo, line 44: This is a use case for assert, isn't it?

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        await user.type(screen.getAllByTestId("search-bar-input")[0], Papers.demoPaper1.title);
+
+        const paper1 = screen.getByText(Papers.demoPaper1.title, { exact: false });
+        expect(paper1).toBeInTheDocument();
+        const paper2 = screen.queryByText(Papers.demoPaper2.title, { exact: false });
+        expect(paper2).not.toBeInTheDocument();
+    });
+
+    test("When citations are searched, then only search results are shown", async () => {
```
Merseleo:
```suggestion
    test("When citations are searched, then only matching citation papers are shown", async () => {
```

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+    const tabs = [
+        { value: "1", label: "Review Criteria" },
+        { value: "2", label: "References/Citations" },
+    ];
+</script>
+
+<!--
+@component
+Paper Card for paper review information in the Paper View component.
```
Merseleo:
```suggestion
Card for review information of a paper in the Paper View component.
```

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
```diff
@@ -0,0 +1,41 @@
...
+    } from "./ReferencesAndCitationsCardContent.svelte";
+
+    let {
+        backwardReferencedPapers,
+        forwardReferencedPapers,
+    }: ReferencesAndCitationsCardContentProps = $props();
+
+    const tabs = [
+        { value: "1", label: "Review Criteria" },
+        { value: "2", label: "References/Citations" },
```
Merseleo: Imo it is not so user friendly, that the tabs are not only completely renamed ("Review Citeratia" to "Review Information" makes totally sense, but why "References/Citations" and "Forward/Backward References") but also the order of the tabs is changed?! My suggestion is to only rename "criteria" to "information" and change the default tab, but keep the order. What do you think?

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Merseleo, line 1: I do not really understand, why you have this and the PaperReaearchContextCard wrapper components? They seem to be very very similar and the advantage of this is not quite clear to me yet. Why don't have on component and change one tab name and the default tab depending on the review mode (which will be stored in a store soon and can be easily accessed) and the change the child component depending on the review mode? Because all the cards make it a bit confusing imo.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+    import SearchBar from "$lib/components/composites/search-bar/SearchBar.svelte";
+    import { Fzf } from "fzf";
+    import PaperListEntry from "../../PaperListEntry.svelte";
+    import PaperListEntrySkeleton from "../../PaperListEntrySkeleton.svelte";
+    import type { Paper } from "$lib/model/api/paper";
+
+    export interface ReferencesAndCitationsCardContentProps {
+        backwardReferencedPapers: Promise<Paper[]>;
+        forwardReferencedPapers: Promise<Paper[]>;
+    }
```
Merseleo, line 12: Why don't you move this interface in the general.ts, if you want it to use somewhere else?

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+        backwardReferencedPapers: allBackwardReferencedPapers,
+        forwardReferencedPapers: allForwardReferencedPapers,
+    }: ReferencesAndCitationsCardContentProps = $props();
+
+    let backwardReferencedPapers = $state<Promise<Paper[]>>(allBackwardReferencedPapers);
+    let forwardReferencedPapers = $state<Promise<Paper[]>>(allForwardReferencedPapers);
+
+    function filterPapers(allPapers: Paper[], searchText: string) {
+        const fzf = new Fzf(allPapers, {
+            selector: (paper) => `#${paper!.id} ${paper!.title}`,
```
Merseleo: What do you think of searching by more keywords than only the id and paper title? So by authors or external id or so too? And maybe we could outsource the different search functions as we need to reuse this function in the general paper search too later.

Merseleo: And if you outsource this function, imo it would be helpful to have a comment for this function which describes which keywords or content of the paper is used for search.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+
+    function filterForwardReferencedPapers(searchText: string) {
+        forwardReferencedPapers = allForwardReferencedPapers.then((allPapers) =>
+            filterPapers(allPapers, searchText),
+        );
+    }
+</script>
+
+<section class="flex flex-col h-full gap-5">
+    <div class="flex h-full overflow-hidden">
```
Merseleo: Why did you encapsulated the named list in these divs? I delete them and it seems as they make no difference?

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+        <NamedList
+            listName="References"
+            items={backwardReferencedPapers}
+            showNumberOfListItems
+            numberOfSkeletons={3}
+        >
+            {#snippet preListContent()}
+                <SearchBar onSearch={filterBackwardReferencedPapers} timeoutInMs={0} />
```
Merseleo, line 55: Why did you override the timeout to 0. I thought we tested it and agreed, that 500ms is ok?

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+            {/snippet}
+            {#snippet listItemComponent(paper)}
+                <PaperListEntry {paper} projectId={undefined} showReviewStatus={false} />
```
Merseleo, line 58: This seems a bit weird to me: why you always set the projectId to undefined and the review status to false? I think there should be the option to have a list of project papers in there?

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+    }
+</script>
+
+<section class="flex flex-col h-full gap-5">
+    <div class="flex h-full overflow-hidden">
+        <NamedList
+            listName="References"
+            items={backwardReferencedPapers}
+            showNumberOfListItems
+            numberOfSkeletons={3}
```
Merseleo, line 50:
```suggestion
            listName="References"
            items={backwardReferencedPapers}
            showNumberOfListItems
            numberOfSkeletons={3}
            emptyHint="No references found."
```

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -0,0 +1,80 @@
...
+                <PaperListEntrySkeleton />
+            {/snippet}
+        </NamedList>
+    </div>
+    <div class="flex h-full overflow-hidden">
+        <NamedList
+            listName="Citations"
+            items={forwardReferencedPapers}
+            showNumberOfListItems
+            numberOfSkeletons={3}
```
Merseleo, line 70:
```suggestion
            listName="Citations"
            items={forwardReferencedPapers}
            showNumberOfListItems
            numberOfSkeletons={3}
            emptyHint="No citations found.
```

src/routes/paper/[paperId]/+page.ts
Comment (Merseleo) [inline src/routes/paper/[paperId]/+page.ts:1]: I think you forgot the error handling (so your catch cases). Because if the paper (id) was not found or the server is not running and anything could be loaded, the plain grpc errors are shown to the user and no custom messages which could be helpful or are understandable.
Bildschirmfoto 2025-02-15 um 17 03 27
Bildschirmfoto 2025-02-15 um 17 52 30

So I think you need to catch the errors and forward better error messages as on the homepage.

Merseleo (changes requested):
It seems as if we are making great strides towards a finished paper page. 
I think, on the whole, the two list work greatly. I have also tested it and if there is data, it works as intended, only if there is no data I don't see such good error messages as a user. 
Another thing that confused me: in the mockups we have a button to add a new paper in the normal view after the list or if there were no search results. Did you leave this out on purpose and if so, then it would be good to create an issue for it or link it and record it somewhere in the code.

src/lib/components/composites/list/NamedList.svelte
Slartibartfass2: I thought it would make sense to keep it visually consistent, and the search bar shouldn't pop out of nowhere when the list is finally loaded. IMO, we can hide it in the error case, but I would keep it while loading.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: Great catch, didn't see that.

I didn't find a way to use [extension methods](https://www.geeksforgeeks.org/how-to-use-extension-methods-in-typescript/) because there exists not `prototype` property on the generated types. So we have to pass the objects as arguments :(.

How would you reuse that in `PaperInfo`?

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: Having a project independent paper view further complicates everything.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: No, this is due to the union type `PaperListEntryInterface`. Now, that `showReviewStatus` is either `boolean` or `false`, it doesn't work with default values, and we have explicitly set the values to have the type safety.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: This is the same answer as for https://github.com/SE-UUlm/snowballr-frontend/pull/184/files#r1957159395.

When using a union type, we have to explicitly set the values so that the type can be inferred correctly.
When we request the backward and forward referenced papers, we only receive `Paper` object, which is why we can only pass them as `Paper` object and not as `Project_Paper` objects here.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: I agree, they differ mostly on the review criteria tab, which was the reason for me to do it separately.
We can merge them.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: Probably, forgot to rename "References/Citations" in the `PaperReviewCard`, I changed that.
We decided to set this order, so that in the review mode the review information is seen directly and the other mode not, so the user has to actively decide to see the review information, which might bias them.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: Changed it.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: I think it makes sense to keep it with the component because if it wouldn't be in the same file, one would have to go to another file to see the props, which seems wrong/confusing to me. This way, we see the props directly and can use it in components that use this component.
Furthermore, changes can be directly applied in here and not somewhere else.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: I outsourced the search function and we now also check the authors. I wouldn't add more because the user only sees these properties. Otherwise, it would be confusing if you type the publication name or something but don't see that it matches it.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: The divs are important to limit the overflow of the lists.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: We did that for asynchronous calls, but in this case, we have all the data on the client (maybe we should change that later) and therefore can update the list immediately.

src/lib/components/composites/search-bar/SearchBar.svelte
Slartibartfass2: I agree, I added it.

src/lib/model/general.ts
Slartibartfass2: Good catch. I added an explanation.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: Added it.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: Added it.

src/routes/paper/[paperId]/+page.ts
Comment (Slartibartfass2) [reply src/routes/paper/[paperId]/+page.ts:1]: Yes, that was missing in `NamedList`. I changed it.

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
Slartibartfass2: It's the same, isn't it?

Slartibartfass2 (commented):
Thanks for the review.

tests/integration/paper-components/paper-list-entry.test.ts
Slartibartfass2: I'm a bit confused, the test you reference tests a custom `onClick` function. Which one, do you mean?

Slartibartfass2:
> It seems as if we are making great strides towards a finished paper page. I think, on the whole, the two list work greatly. I have also tested it and if there is data, it works as intended, only if there is no data I don't see such good error messages as a user. Another thing that confused me: in the mockups we have a button to add a new paper in the normal view after the list or if there were no search results. Did you leave this out on purpose and if so, then it would be good to create an issue for it or link it and record it somewhere in the code.

For the feature you ask for, [this issue](https://github.com/SE-UUlm/snowballr-frontend/issues/43) already exists.

src/lib/components/composites/list/NamedList.svelte
Merseleo: Ok, i understand your reasoning and your suggestions seems to be a good idea: keep it while loading and hide it, when an error occurred, so that only the error is visible. 
Are you implementing this in this PR?

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: I agree, but especially for the reading list etc. it makes sense, so imo we have the time to deal with this independent view and it could be helpful, so we should take on this additional effort.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: Ok.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: I would reuse this method, as you already did when creating the `PaperInfo` in the `PaperListEntry` component.

And imo it is not a problem to pass the objects as arguments.

src/lib/model/general.ts
```diff
@@ -18,10 +19,32 @@ interface ProjectListEntryInterface {
...
+function isProjectPaper(paper: Project_Paper | Paper): paper is Project_Paper {
+    return "paper" in paper;
+}
+
+function asPaper(paper: Project_Paper | Paper): Paper {
+    return isProjectPaper(paper) ? paper.paper! : paper;
 }
 
 export type { ValidationResult, ApiError, ProjectListEntryInterface, PaperListEntryInterface };
+export { isProjectPaper, asPaper };
```
Merseleo: But now can keep the isProjectPaper method locally as you already export the asPaper method and don't use the isProjectPaper method anymore?

src/lib/components/composites/search-bar/SearchBar.svelte
```diff
@@ -54,9 +54,12 @@ and the search method is triggered with an empty string,
 so if no further handling exist, all elements are retrieved
 by the search.
 
+To limit the amount of calls (if used with backend calls), the search only starts after a
+specified timout in which no other events occur. The default for this is 500 ms.
```
Merseleo:
```suggestion
To limit the amount of calls (if used with backend calls), the search only starts after a
specified timeout in which no other events occur. The default timeout is 500 ms.
```

tests/integration/paper-components/paper-list-entry.test.ts
Merseleo: Ops, sorry. I was confused by the way Github showed me the diffed and it seems as you deleted code in other tests. My mistake ...

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
Merseleo: It is, but if I remember correctly we once discussed about where using assert could make sense (especially in case of readability). Yes afaik it is the same and you dont need to change anything.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Merseleo: I completely understand your point, but why not change the default tab instead of the order of the tabs. So that always "References" is the first tab und "Review information" the second, but when I enter the review mode, the second tab "review information" is automatically selected?

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Merseleo: Ok, thank you.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Merseleo: I think you forgot to merge this update as well? Because the comment in `PaperResearchContextCard` is still the old one and not correct anymore for the new component.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: Ok, if it works this way, I am conform with this (because WebStorm show errors because of the export of the interface and the way you did causes problems in WebStorm, but nothing important).

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: Makes sense, I am fine with your solution.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: I tested it again and combined with your new `flex-[1_1_0]` it seems to work now, because before your change, the divs didn't make a difference, when I tested it, but now. Thank you for fixing it.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: Ah ok, that makes sense. I didn't consider, that all data are already in the client.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: Ok, in my head I was still thinking you solved it with optional parameters, but I had another look at your interface and now it makes sense to me.

src/routes/paper/[paperId]/+page.ts
Comment (Merseleo) [reply src/routes/paper/[paperId]/+page.ts:1]: Perfect. I tested it and it works 👍.
Would you mind testing the whole behaviour (i.e. whether the hint or error is shown when the map is rendered with an empty list or an error). I know it doesn't increase the coverage, but it feels a bit more complete or do you want to leave it out because the list itself is already tested (which I would totally understand)

src/lib/components/composites/list/NamedList.svelte
```diff
@@ -12,6 +12,8 @@
         showNumberOfListItems?: boolean;
         numberOfItems?: number;
         emptyHint?: string;
+        errorHint?: string;
```
Merseleo, line 15: I already solved this problem actually by catching the errors in the page load and rename the error message, so dont need to think of setting an error hint, especially as your way only allows to display one error message and not multiple, which is not ideal imo.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -11,12 +11,19 @@
...
     };
 
-    const navigateToPaperView = () => goto(`/project/${projectId}/paper/${projectPaper.id}`);
+    const navigateToPaperView = () => {
+        const paperId = asPaper(paper).id;
+        const paperLink =
+            projectId !== undefined
+                ? `/project/${projectId}/paper/${paperId}`
+                : `/paper/${paperId}`;
+        goto(paperLink);
```
Merseleo, line 20: I found another problem with this goto (didn't recognize this earlier, sorry): Try to add two papers as references and click on one reference, which should navigate you to this new page. It does happen, but the title in the navigation bar and everything else will not be reloaded and the old paper information are displayed.

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
```diff
@@ -1,37 +1,73 @@
...
+        : [
+              { value: "1", label: "Forward/Backward References" },
+              { value: "2", label: "Review Information" },
+          ];
 </script>
 
 <!--
 @component
-Paper Card for paper research context in the Paper View component.
+Paper Card for review information of a paper in the Paper View component.
```
Merseleo:
```suggestion
Card for displaying the review information and forward / backward references of a paper in the Paper View component.
```

src/lib/filters.ts
Merseleo, line 1: Did you test these methods or it is ok to not explicitly test them?
And second question: do you think we should move this file in the `utils` folder?

src/lib/filters.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+function filter<T>(items: T[], selector: (item: T) => string, searchText: string) {
+    // Here we case the items to string[] and selector to Selector<string> because we can't just use a selector
+    // with a generic type. This is weird but it's the only way to make it work.
+    const fzf = new Fzf(items as string[], {
+        selector: selector as Selector<string>,
+        casing: "case-insensitive",
+    });
+    return fzf.find(searchText).map((result) => result.item) as T[];
+}
```
Merseleo: Nice typescript generics, very useful for my exam preparation ;) (close it by yourself)

Merseleo (changes requested):
Great work, I think we will finish it in this round.

src/routes/paper/[paperId]/+page.ts
Comment (Slartibartfass2) [reply src/routes/paper/[paperId]/+page.ts:1]: No, I think that's fine. I added them.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: Yes, I would've just preferred to call `projectPaper.asPaper()` instead of `asPaper(projectPaper)` but that's just syntactic sugar.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Slartibartfass2: Yes, I already noticed this too, and I'm convicted this is some svelte behavior that we don't know because the URL changes, but the page stays the same. I would like to handle this in a separate issue, e.g., when we face navigating papers or in an entirely new one.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Slartibartfass2: Yes, maybe later we want to do the filtering on the server side, and then we would use a delay instead again.

src/lib/components/composites/list/NamedList.svelte
Slartibartfass2: Yes, I added it.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: As discussed, we keep it this way for now and discuss it later when we gather some user feedback. I mentioned it [here](https://github.com/SE-UUlm/snowballr-frontend/issues/211).

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: Currently, it's "Paper Card for review information of a paper in the Paper View component." (I kept the "Paper ")
Is that alright?

Slartibartfass2 (commented):
This should be all. Thanks for the review.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Merseleo: No it's not, as this is not only the card only for review information but also for the references, isn't it?

src/routes/paper/[paperId]/+page.ts
Comment (Merseleo) [reply src/routes/paper/[paperId]/+page.ts:1]: Thanks.

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: Ok, then I write a bug issue for this problem.

Merseleo (changes requested):
I think you missed some comments.

src/lib/components/composites/list/NamedList.svelte
Slartibartfass2: If the error hint is not set, it displays the content of the error object, so both ways work.

src/lib/components/composites/paper-components/paper-view/cards/PaperReviewCard.svelte
Slartibartfass2: I improved the documentation.

src/lib/model/general.ts
Slartibartfass2: Yes, I kept it exported if we may ever want to use it. But if so, we can export it again.

src/lib/components/composites/search-bar/SearchBar.svelte
Slartibartfass2: Definitely need spell checking in VS Code. I changed it.

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
Slartibartfass2: Changed it.

src/lib/filters.ts
Slartibartfass2: I am not sure if we should, actually. I tested it through the component because this way I see if it works for the component. The method itself is just FZF, so IDK....

Slartibartfass2 (commented):
This time, it should be all answered.

src/lib/components/composites/list/NamedList.svelte
Merseleo: Ok I am also fine with that.

src/lib/model/general.ts
Merseleo: Basically yes, but I just think that you should export as little as possible and thus expose, especially because, as you say, if you then need it, you can simply export it quickly.

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: A great idea ;).

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
Merseleo: Better now imo.

src/lib/filters.ts
Merseleo: I agree, I think it is ok to "only" implicitly test them and fzf should work correctly...

Merseleo (approved):
I think all problems are solved, everything well tested and documented and the functionality should be added to our develop version ...

PR #191, opened 2025-02-04
Added ability to remove project members

### LINKED ISSUE #19: [Use Case]: Remove Project Members

Project Admin

### Describe the use case

As project admin, I want to remove project members.

### Acceptance criteria

- a project member, which was removed from a project, isn't considered a project member anymore
- a project admin, cannot be removed from a project by another project admin

### Anything else?

Part of #15

Slartibartfass2:
Closes #19 

## What I have made

- I noticed that the tab list of the navigation bar wraps its tabs (buttons) in an \<a\> tag, therefore the button is never actually used (click events are handled by the \<a\> tag)
  - I changed the tabs trigger to directly be an \<a\> tag when `href` is provided
  - the change can be seen in https://github.com/SE-UUlm/snowballr-frontend/pull/191/commits/24f28708338b2acbac53e913c29b214eff4c807e
- added a reusable alert dialog, which is used in the create project dialog and as remove project member dialog
- added the remove project member dialog
- added e2e tests for invite users workflow
- added e2e tests for remove project members workflow

## Checklist

  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -4,18 +4,22 @@ import type { TestOptions } from "./tests/e2e/fixtures/general-fixture";
 export default defineConfig<TestOptions>({
     testDir: "tests/e2e",
     forbidOnly: !!process.env.CI,
-    retries: process.env.CI ? 2 : 3,
-    // Opt out of parallel tests on CI.
+    retries: 2,
     failOnFlakyTests: !!process.env.CI,
+    // Opt out of parallel tests on CI.
     workers: process.env.CI ? 1 : undefined,
     reporter: [
         [process.env.GITHUB_ACTIONS ? "github" : "list"],
         ["html", { outputFolder: "e2e-report" }],
     ],
+    expect: {
+        timeout: 2_500,
+    },
+    timeout: 15_000,
     use: {
         baseURL: "http://localhost:4173",
         screenshot: "on",
-        trace: "retain-on-failure",
+        trace: "on",
     },
     webServer: {
         command: "npm run build && npm run preview",
@@ -72,7 +76,6 @@ export default defineConfig<TestOptions>({
             },
             dependencies: ["firefox setup"],
         },
-
         {
             name: "webkit",
             use: {
--- a/src/lib/components/composites/dialog/AlertDialog.svelte
+++ b/src/lib/components/composites/dialog/AlertDialog.svelte
@@ -6,6 +6,7 @@
     import type { DialogTriggerProps } from "bits-ui";
     import type { AlertDialogCancelProps } from "$lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte";
     import ErrorAlert from "$lib/components/composites/utils/ErrorAlert.svelte";
+    import { wrapLongWords } from "$lib/utils/common-helper";
 
     interface Props {
         triggerProps?: DialogTriggerProps;
@@ -115,7 +116,8 @@ Usage:
     {/if}
     <AlertDialog.Content>
         <AlertDialog.Header>
-            <AlertDialog.Title><h1>{title}</h1></AlertDialog.Title>
+            <!-- eslint-disable-next-line svelte/no-at-html-tags -->
+            <AlertDialog.Title><h1>{@html wrapLongWords(title, 30)}</h1></AlertDialog.Title>
             <AlertDialog.Description class="text-default-m" data-testid="alert-dialog-description">
                 {@render description?.()}
             </AlertDialog.Description>
--- a/src/lib/components/composites/dialog/AlertDialog.svelte
+++ b/src/lib/components/composites/dialog/AlertDialog.svelte
@@ -1,0 +1,142 @@
+<script lang="ts">
+    import type { AlertDialogActionProps } from "$lib/components/primitives/alert-dialog/alert-dialog-action.svelte";
+    import * as AlertDialog from "$lib/components/primitives/alert-dialog/index.js";
+    import type { Snippet } from "svelte";
+    import LoaderCircle from "lucide-svelte/icons/loader-circle";
+    import type { DialogTriggerProps } from "bits-ui";
+    import type { AlertDialogCancelProps } from "$lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte";
+    import ErrorAlert from "$lib/components/composites/utils/ErrorAlert.svelte";
+
+    interface Props {
+        triggerProps?: DialogTriggerProps;
+        trigger?: Snippet;
+        title?: string;
+        description?: Snippet;
+        cancelProps?: AlertDialogCancelProps;
+        cancelButtonText?: string;
+        actionProps?: AlertDialogActionProps;
+        actionButtonText?: string;
+        errorText?: string;
+        loading?: boolean;
+        error?: unknown;
+        open?: boolean;
+    }
+
+    let {
+        triggerProps = {},
+        trigger = undefined,
+        title = "Are you absolutely sure?",
+        description = undefined,
+        cancelProps = {},
+        cancelButtonText = "Cancel",
+        actionProps = {},
+        actionButtonText = "Confirm",
+        errorText = "An error occurred",
+        loading = $bindable(false),
+        error = $bindable(undefined),
+        open = $bindable(false),
+    }: Props = $props();
+</script>
+
+<!--
+@component
+Highly customizable AlertDialog Element.
+
+Bindings:
+- `loading`: Whether the action button is in a loading state.
+- `error`: The error object e.g. from a `catch` block, when defined, an error alert is shown.
+
+With those bindings you can easily handle the loading state and error state of the dialog.
+Example:
+```svelte
+    <script lang="ts">
+        ...
+
+        let loading = $state(false);
+        let error = $state<unknown>(undefined);
+        let open = $state(false);
+
+        function dialogOnClick() {
+            error = undefined;
+            loading = true;
+            await someAsyncFunction()
+                .then(() => {
+                    console.log("Success");
+                    open = false;
+                })
+                .catch((error) => {
+                    console.error("Error", error);
+                    error = error;
+                });
+            loading = false;
+        }
+    </script>
+
+    ...
+
+    <AlertDialog
+        ...
+        actionProps={{
+            onclick: dialogOnClick,
+        }}
+        bind:loading
+        bind:error
+        bind:open
+    >
+        ...
+    </AlertDialog>
+```
+
+Usage:
+```svelte
+    <AlertDialog
+        title="This is the title of the dialog"
+        actionProps={{
+            variant: "destructive",
+            onclick: () => console.log("Action button clicked"),
+        }}
+        actionButtonText="Execute Destructive Action"
+        errorText="Couldn't execute action"
+    >
+        {#snippet trigger()}
+            Open Dialog
+        {/snippet}
+        {#snippet description()}
+            This is a description
+        {/snippet}
+    </AlertDialog>
+```
+-->
+<AlertDialog.Root bind:open>
+    {#if trigger}
+        <AlertDialog.Trigger {...triggerProps} data-testid="alert-dialog-trigger">
+            {@render trigger()}
+        </AlertDialog.Trigger>
+    {/if}
+    <AlertDialog.Content>
+        <AlertDialog.Header>
+            <AlertDialog.Title><h1>{title}</h1></AlertDialog.Title>
+            <AlertDialog.Description class="text-default-m" data-testid="alert-dialog-description">
+                {@render description?.()}
+            </AlertDialog.Description>
+        </AlertDialog.Header>
+        {#if error}
+            <ErrorAlert errorTitle={errorText} />
+        {/if}
+        <AlertDialog.Footer>
+            <AlertDialog.Cancel {...cancelProps} data-testid="alert-dialog-cancel">
+                {cancelButtonText}
+            </AlertDialog.Cancel>
+            <AlertDialog.Action
+                {...actionProps}
+                data-testid="alert-dialog-action"
+                disabled={loading}
+            >
+                {#if loading}
+                    <LoaderCircle class="animate-spin" />
+                {/if}
+                {actionButtonText}
+            </AlertDialog.Action>
+        </AlertDialog.Footer>
+    </AlertDialog.Content>
+</AlertDialog.Root>
--- a/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
@@ -75,5 +75,6 @@ Usage:
     {#snippet description()}
         Once removed, <span class="font-bold">{memberName}</span> will no longer have access to this
         project. You can always re-invite them later.
+        <!-- TODO: Clarify what happens with deleted users reviews -->
     {/snippet}
 </AlertDialog>
--- a/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
@@ -63,6 +63,7 @@ Usage:
     triggerProps={{
         disabled: isDisabled,
         class: buttonVariants({ variant: "destructive-subtle", size: "icon" }),
+        "aria-label": `Remove member ${member.user!.email}`,
     }}
     bind:loading
     bind:error
--- a/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
@@ -55,14 +55,14 @@ Usage:
 <AlertDialog
     actionButtonText={`Remove ${memberName} From This Project`}
     actionProps={{
-        variant: "destructive-subtle",
+        variant: "destructiveSubtle",
         onclick: removeMember,
     }}
     errorText="Couldn't remove member"
     title={`Remove ${memberName} From This Project`}
     triggerProps={{
         disabled: isDisabled,
-        class: buttonVariants({ variant: "destructive-subtle", size: "icon" }),
+        class: buttonVariants({ variant: "destructiveSubtle", size: "icon" }),
         "aria-label": `Remove member ${member.user!.email}`,
     }}
     bind:loading
--- a/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
@@ -53,7 +53,7 @@ Usage:
 ```
 -->
 <AlertDialog
-    actionButtonText={`Remove ${memberName} From This Project`}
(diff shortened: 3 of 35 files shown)


src/lib/components/composites/dialog/AlertDialog.svelte
```diff
@@ -0,0 +1,142 @@
...
+        }}
+        bind:loading
+        bind:error
+        bind:open
+    >
+        ...
+    </AlertDialog>
+```
+
+Usage:
```
mowi12, line 91: Isn't this usage useless if you provide an example before?

src/lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte
mowi12, line 1: Why did you change the type? This doesn't change anything.

playwright.config.ts
```diff
@@ -4,18 +4,22 @@ import type { TestOptions } from "./tests/e2e/fixtures/general-fixture";
...
+    retries: 2,
     failOnFlakyTests: !!process.env.CI,
+    // Opt out of parallel tests on CI.
     workers: process.env.CI ? 1 : undefined,
     reporter: [
         [process.env.GITHUB_ACTIONS ? "github" : "list"],
         ["html", { outputFolder: "e2e-report" }],
     ],
+    expect: {
+        timeout: 2_500,
```
mowi12, line 16: 2.5 seconds seem a bit short. Did you test it with shorter times, and they worked as well, and 2.5 seconds is already with puffer?

tests/integration/navigation-bar/navigation-bar.test.ts
```diff
@@ -105,7 +104,7 @@ describe("NavigationBar", () => {
             },
         });
 
-        const linkTags = screen.getAllByRole("link");
+        const linkTags = screen.queryAllByRole("link");
```
mowi12, line 107: What's the difference between `get` and `query`?

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -84,5 +84,5 @@ describe("KeywordSettings", () => {
         await userEvent.keyboard("{Enter}");
         expect(screen.queryByText("Maximum number of keywords reached!")).toBeInTheDocument();
         expect(screen.queryByText(`New Tag ${maximumAmountOfTags}`)).not.toBeInTheDocument();
-    });
+    }, 10000);
```
mowi12: Does this actually reduce the amount of flaky tests you receive?

src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
```diff
@@ -0,0 +1,79 @@
...
+    bind:loading
+    bind:error
+    bind:open
+>
+    {#snippet trigger()}
+        <Trash class="size-6!" />
+    {/snippet}
+    {#snippet description()}
+        Once removed, <span class="font-bold">{memberName}</span> will no longer have access to this
+        project. You can always re-invite them later.
```
mowi12: What about the users reviews?
Should this also be stated in here?

src/lib/components/primitives/button/button.svelte
```diff
@@ -9,6 +9,8 @@
             variant: {
                 default: "bg-primary text-primary-foreground hover:bg-primary/90",
                 destructive: "bg-destructive text-destructive-foreground hover:bg-destructive/90",
+                "destructive-subtle":
```
mowi12:
```suggestion
                destructiveSubtle:
```

mowi12: Don't forget to update all occurences.

src/routes/project/[projectId]/settings/members/+page.svelte
```diff
@@ -27,27 +28,43 @@
         },
     );
 
-    async function reloadMembers() {
+    async function reloadMembers(errorMessage: string) {
         // First fetch the members again and only then replace them, so that no loading state is shown
         await loadMembers({ id: projectId })
             .then((members) => {
                 loadingMembersLocal = Promise.resolve(members);
```
Comment (mowi12) [inline src/routes/project/[projectId]/settings/members/+page.svelte:35]: As i should have already commented in the main review comment, I think you need to invalidate something in order to be able to invite the last removed user directly without refreshing the site.

tests/integration/settings/project-settings/members/remove-member-dialog.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        expect(removeButton).toBeInTheDocument();
+
+        const description = screen.getByTestId("alert-dialog-description");
+        expect(description).toBeInTheDocument();
+        expect(description).toHaveTextContent(
+            `Once removed, ${memberName} will no longer have access to this project. You can always re-invite them later.`,
+        );
+    });
+
+    test("When the member is an admin, then the trigger is disabled", async () => {
```
mowi12, line 63: I assume you mean the member to remove?

tests/integration/settings/project-settings/members/remove-member-dialog.test.ts
```diff
@@ -0,0 +1,138 @@
...
+                member: adminMember,
+                isCurrentUser: false,
+            },
+        });
+
+        const trigger = screen.getByTestId("alert-dialog-trigger");
+        expect(trigger).toBeDisabled();
+    });
+
+    test("When the member is the current user, then the trigger is disabled", async () => {
```
mowi12, line 77: Same here.

tests/e2e/fixtures/general-fixture.ts
```diff
@@ -60,4 +63,19 @@ export const mockBackendServiceTest = base.extend<
...
+    email: "alice.smith@example.com",
+    password: "12abAB!?",
+};
+
+/**
+ * Extends the default test fixture with a test user.
+ * This test user is used is registered before each test.
+ */
+export const userTest = base.extend<{ user: TestUser }>({
+    user: [testUser, { option: true }],
```
mowi12: What is the `option` field doing?

tests/e2e/fixtures/project-members-settings-fixtures.ts
```diff
@@ -0,0 +1,25 @@
...
+        ).response;
+        await page.goto(`/project/${project.id}/dashboard`);
+        await page.getByRole("tab", { name: "Settings" }).click();
+        await page.getByRole("link", { name: "Members" }).click();
+        await expect(page.getByRole("heading", { name: "Demo Project", level: 2 })).toBeVisible();
+        const projectMembersSettingsPage = new DevProjectMemberSettingsPage(page);
+
+        await use(projectMembersSettingsPage);
+
+        await mockBackendService.softDeleteProject({ id: project.id });
```
mowi12, line 23: Why do you already delete this project here?

tests/e2e/helper.ts
```diff
@@ -14,3 +14,25 @@ export async function reloadWait(page: Page, element: Locator) {
...
+ * This function assumes that the user is logged in and the user menu is visible.
+ * It opens the user menu, retrieves the user's name, and closes the menu again.
+ *
+ * **Note**: This function is only necessary because the user name is changed in other tests and they interfere with each other.
+ *
+ * @param page - the current page
+ * @returns the first and last name of the user
+ */
+export async function getCurrentUser(page: Page): Promise<{ firstName: string; lastName: string }> {
+    await page.getByRole("button", { name: /[A-Z]{2}/ }).click();
```
mowi12:
```suggestion
    await page.getByRole("button", { name: /^[A-Z]{2}$/ }).click();
```

If there are any two capital letters in the project which are on a button, they would also match, now they need to be at the start and end of a line making it more rare to match.
-> `home-page-model.ts`

tests/e2e/helper.ts
```diff
@@ -14,3 +14,25 @@ export async function reloadWait(page: Page, element: Locator) {
...
+ *
+ * This function assumes that the user is logged in and the user menu is visible.
+ * It opens the user menu, retrieves the user's name, and closes the menu again.
+ *
+ * **Note**: This function is only necessary because the user name is changed in other tests and they interfere with each other.
+ *
+ * @param page - the current page
+ * @returns the first and last name of the user
+ */
+export async function getCurrentUser(page: Page): Promise<{ firstName: string; lastName: string }> {
```
mowi12: Imo this function name is a little bit misleading, isn't it?

tests/e2e/invite-users.test.ts
```diff
@@ -0,0 +1,112 @@
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/project-members-settings-fixtures";
+import { getName } from "$lib/utils/common-helper";
+import { getCurrentUser } from "./helper";
+
+test.describe("Inviting users to a project", () => {
+    /**
+     * Here we test for project members. In the fixture we create a project with one admin and one invitee.
+     */
```
mowi12, line 9: Is there a reason why this comment is inside `test.describe` and not above?

tests/e2e/pom/project-member-settings-page-model.ts
```diff
@@ -0,0 +1,135 @@
...
+            await this.inviteUsersDialogInput.press("ArrowDown");
+            await this.inviteUsersDialogInput.press("Enter");
+        } else {
+            await this.inviteUsersDialogInput.press("Tab");
+        }
+    }
+
+    /**
+     * Checks whether the user with the passed userId is listed in the project members.
+     */
```
mowi12, line 77: Missing `@param`.

mowi12 (changes requested):
Wow, this looks really good. I don't see any logical error or similar. I only have some minor adjustments that we need to talk about and two problems I faced.

1. After removing an user from the project, this user is not in the list of users you can add when you click on the 'Invite Users' button. Only after reloading the page, this users is in this list.

2. Screenshot 2025-05-14 at 17 15 08

This happens when the user name is the email adress.

src/routes/project/[projectId]/settings/members/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/settings/members/+page.svelte:35]: I fixed it. The only problem was that in `InviteUsersDialog` the list of members is not updated.

src/lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte
Slartibartfass2: Not really no, as we do not change the type. I mainly did it for exporting, but I realized that I could've used the type directly. However, this appears more clean to me.

src/lib/components/composites/dialog/AlertDialog.svelte
Slartibartfass2: I see your point that it feels duplicated. The first example is about binding the properties and the second one is about the general usage. I think more documentation is better than less, even if there's a bit duplicated.

src/lib/components/primitives/button/button.svelte
Slartibartfass2: Done.

tests/e2e/fixtures/general-fixture.ts
Slartibartfass2: `option` makes the parameter available in `playwright-config`. But as we do not use this, this can be removed. Good catch!

tests/e2e/fixtures/project-members-settings-fixtures.ts
Slartibartfass2: This is after the test is executed.
`use` executes the test, i.e., the fixture is a wrapper of the test.
Everything before `use` is the before each hook, and everything after `use` is the after each hook.

tests/e2e/helper.ts
Slartibartfass2: Good catch

tests/e2e/helper.ts
Slartibartfass2: What is misleading you?

tests/e2e/invite-users.test.ts
Slartibartfass2: No, not really. Seemed like the best place to me.

tests/e2e/pom/project-member-settings-page-model.ts
Slartibartfass2: Great catch. I added more documentation.

tests/integration/navigation-bar/navigation-bar.test.ts
Slartibartfass2: `get` throws if there's nothing, `query` just returns a list, which can also be empty.
Now that the `<a>` tags have the `tab` role, they don't have the `link` role anymore, which is why this wouldn't have worked with `get`. While there are many ways to test this, this was the one way with the least difference.

tests/integration/settings/project-settings/members/remove-member-dialog.test.ts
Slartibartfass2: Yes, that's correct. But this what the component is about. Do you think we require a more detailed description?

tests/integration/settings/project-settings/members/remove-member-dialog.test.ts
Slartibartfass2: See above.

tests/integration/settings/project-settings/review/keyword-settings.test.ts
Slartibartfass2: Yes, the test timed out on some occasions, which is why increased the timeout. I added a comment for clarification.

src/routes/project/[projectId]/settings/members/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/settings/members/+page.svelte:35]: I added a test that ensures that this works.

src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
Slartibartfass2: Do we have a defined procedure for that?

playwright.config.ts
Slartibartfass2: I checked this again and the longest I've got was 1.3s, so I think 2.5s is fine. Initially, I chose this because it's half of the default.

As I use a slow ass machine, this should be fine. Did you experience any errors with that?

Slartibartfass2 (commented):
Thank you very much for the fast review of these many changes.

Next to changing what you've noticed, I also changed the signature of `pluralize` to also accept an array, as this is used frequently. Additionally, I removed the member name from the action button of the dialog as this often caused an overflow.

src/lib/components/composites/dialog/AlertDialog.svelte
mowi12: Probably true. Then we keep it.

src/lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte
mowi12: I understand. But if you want to keep it clean and consistent don't you need to update all the files, e.g. `alert-dialog-description.svelte`. Else, I don't see the reason to keep this modification.

playwright.config.ts
mowi12: No, I experienced no errors, I just wanted to make sure you tested it and didn't just choose some value.

tests/integration/navigation-bar/navigation-bar.test.ts
mowi12: Makes sense.

tests/integration/settings/project-settings/review/keyword-settings.test.ts
mowi12: Okay, thanks!

src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
mowi12: If we have one, you should know it, as it would have been decided early on in the project, or not?
But I would be fine if we omit this information or just add a comment with 'TODO: Clarify what happens with deleted users reviews'. What do you think?

src/routes/project/[projectId]/settings/members/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/members/+page.svelte:35]: Perfect!

tests/integration/settings/project-settings/members/remove-member-dialog.test.ts
mowi12: No, you're right. It now makes sense. It was getting pretty late as I reviewed this.

tests/e2e/fixtures/project-members-settings-fixtures.ts
mowi12: Ahh, good to know.

tests/e2e/helper.ts
mowi12: I mean, you kind of get the current user but only the name. If someone else wants to write a test and sees this, wouldn't he assume he gets the full user object? Yes, the documentation states that this only returns the first and last name.
But as I do not have a better solution, I think we can keep this as is.

Resolve on your own if you also want to keep it.

tests/e2e/invite-users.test.ts
mowi12: I guess we can keep it here as it is no documentation of the method, you're right.

mowi12 (changes requested):
Nice, you fixed all the larger problems. There are only some minor discussions we need to finish, but this should handled very fast!

mowi12 (changes requested):
I know, no one has an e-mail adress this long, but my own e-mail adress was pretty close. Is there any way to handle this problem? I don't have any good idea how do handle this case. Splitting the address feels wrong, but may be the only solution.

But in every case, the buttons should not shift that much to the right.

Screenshot 2025-05-15 at 17 26 35

src/lib/components/composites/settings/project-settings/members/RemoveMemberDialog.svelte
Slartibartfass2: Sounds great!

tests/e2e/helper.ts
Slartibartfass2: I renamed it.

Slartibartfass2 (commented):
Thank you for the fast feedback.

I found a way to wrap long emails, what are your thoughts on this?

mowi12 (changes requested):
Nicely done! Looks great. Now only one open conversation is left. Then we are good to go!

src/lib/components/primitives/alert-dialog/alert-dialog-cancel.svelte
Slartibartfass2: But we don't export these props. We only do this for the action and cancel button, which is why I made the modification there.

mowi12: Well, I guess you're right. Since these are no major files and this isn't really important, I think we can agree to keep the current state.

mowi12 (dismissed):
Nice. We really got some work done.

While testing, I found a [bug](https://github.com/SE-UUlm/snowballr-frontend/issues/377), but I don't think this is part of this issue, since it has nothing to do with your logic.

mowi12 (approved):

PR #192, opened 2025-02-04
Added ability to promote project member to project admin

### LINKED ISSUE #24: [Use Case]: Promote Project Member to Project Admin

Project Admin

### Describe the use case

As project admin, I want to promote a project member to project admin.

### Acceptance criteria

- a member who is already an admin cannot be promoted
- after a member is promoted, they are considered a project admin (and still project member)

### Anything else?

Part of #15 

Related to #56

### LINKED ISSUE #405: [Refactor]: Replace all toast with a specific variant

_No response_

In order to achieve a clean and consistent look. All toasts should use a specific variant.

Use a specific variant for each toast. Even default if non is selected.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #24 
Closes #405 

## What I have made

- This is pretty straightforward
- This PR implements the promote members button on the members settings page
- The testing is very similar to the one where members are removed
- Currently, we cannot test whether invited project members can be promoted using the e2e tests
- Please verify this manually
- We need to discuss whether we should implement this in the API, but this requires some refactoring

## Checklist

  - [x] integration tests

### DIFF
--- a/src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
@@ -12,6 +12,7 @@
         isAdminView: boolean;
         onMemberRemoved?: (member: Project_Member) => void;
         onMemberPromoted?: (member: Project_Member) => void;
+        disabled?: boolean;
     }
 
     let {
@@ -21,6 +22,7 @@
         isAdminView,
         onMemberRemoved = undefined,
         onMemberPromoted = undefined,
+        disabled = false,
     }: Props = $props();
 </script>
 
@@ -63,6 +65,7 @@ Usage:
             <span class="text-hint">Invitation Pending ...</span>
         {/if}
         <PromoteMemberDialog
+            {disabled}
             {isAdminView}
             {isCurrentUser}
             {member}
@@ -70,7 +73,7 @@ Usage:
             {projectId}
         />
         {#if isAdminView}
-            <RemoveMemberDialog {isCurrentUser} {member} {onMemberRemoved} {projectId} />
+            <RemoveMemberDialog {disabled} {isCurrentUser} {member} {onMemberRemoved} {projectId} />
         {/if}
     </div>
 </li>
--- a/src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
@@ -1,8 +1,7 @@
 <script lang="ts">
-    import Button from "$lib/components/primitives/button/button.svelte";
-    import { MemberRole, type Project_Member } from "$lib/model/api/project";
+    import { type Project_Member } from "$lib/model/api/project";
     import { getName } from "$lib/utils/common-helper";
-    import { cn } from "$lib/utils/shadcn-helper";
+    import PromoteMemberDialog from "./PromoteMemberDialog.svelte";
     import RemoveMemberDialog from "./RemoveMemberDialog.svelte";
 
     interface Props {
@@ -12,6 +11,7 @@
         isInvitationPending: boolean;
         isAdminView: boolean;
         onMemberRemoved?: (member: Project_Member) => void;
+        onMemberPromoted?: (member: Project_Member) => void;
     }
 
     let {
@@ -21,10 +21,8 @@
         isInvitationPending,
         isAdminView,
         onMemberRemoved = undefined,
+        onMemberPromoted = undefined,
     }: Props = $props();
-
-    const role = member.role === MemberRole.ADMIN ? "Admin" : "Member";
-    const isRoleReadonly = $derived(isCurrentUser || !isAdminView);
 </script>
 
 <!--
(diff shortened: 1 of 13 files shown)


src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
domienderle, line 1: Please update the `Usage` of this component in the according comment. Also add the missing comments for the remove member functionality.

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
```diff
@@ -0,0 +1,90 @@
...
+        projectId: string;
+        member: Project_Member;
+        isCurrentUser: boolean;
+        isAdminView: boolean;
+        onMemberPromoted?: (member: Project_Member) => void;
+    }
+
+    let { projectId, member, isCurrentUser, isAdminView, onMemberPromoted }: Props = $props();
+
+    const memberName = getName(member.user!);
```
domienderle, line 28: In which cases can `member.user` be `undefined` that we need the `!`? I get that we need it, because `member.user` can be `undefined`, but I was wondering if this implementation makes sense at all.

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
```diff
@@ -0,0 +1,90 @@
...
+    }}
+    errorText="Couldn't promote member"
+    title={`Promote ${memberName} to a Project Admin?`}
+    triggerProps={{
+        class: cn(
+            "w-[7.7rem] disabled:opacity-100! select-auto",
+            buttonVariants({ variant: isRoleReadonly ? "ghost" : "destructiveSubtle" }),
+            isRoleReadonly ? "hover:bg-transparent hover:cursor-default" : "text-primary",
+        ),
+        disabled: isRoleReadonly,
```
domienderle: Wouldn't it make more sense to disable the option to promote a member if the role is readonly per default and not in the `AlertDialog`?

tests/e2e/pom/promote-member-dialog-model.ts
```diff
@@ -0,0 +1,56 @@
...
+            displayName = getName({ firstName: user.firstName, lastName: user.lastName });
+        } else {
+            displayName = user.email;
+        }
+
+        this.page = page;
+        this.dialog = page.getByRole("alertdialog", {
+            name: `Promote ${displayName} to a Project Admin`,
+        });
+        this.openButton = page.locator(`button[aria-label="Promote member ${user.email}"]`);
```
domienderle, line 23: Why do you use a `locator` here instead of `getByRole`?

domienderle (changes requested):
Imo an invited user should not be promtoable. But I would remove the role promotion button for invited members and just show the text `Invitation Pending`. Let me know your opinion about this.

src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
Slartibartfass2: Good catch. I updated the usage. What do you mean with the missing comments for the remove member functionality?

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
Slartibartfass2: There are no cases where this is actually true, but the grpc type defines all objects as undefined. Do you think we should check whether the user really exists before using it like that?

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
Slartibartfass2: I'm not sure what you mean by that. With this, we disable the trigger for the dialog.

tests/e2e/pom/promote-member-dialog-model.ts
Slartibartfass2: Because all buttons look the same, i.e., we cannot check for the name of the button, but with the locator, we can check the `aria-label`, which stores information about which member to promote.

Slartibartfass2 (commented):
Thank you for the fast review.
I also changed the toasts so they customized according to the event, e.g., `toast.success` for successful events.

src/lib/components/composites/settings/project-settings/members/ProjectMemberListEntry.svelte
domienderle: Don't mind. You alread added it.

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
domienderle: You implemented it just right. Looks good.

src/lib/components/composites/settings/project-settings/members/PromoteMemberDialog.svelte
domienderle: No I think this is not necessary, since it cannot be `undefined`. I was just wondering, if this makes any sense, because we already know that `user` cannot be `undefined` in an existing `member`. But since this cannot be changed easily, I would let it as it is.

tests/e2e/pom/promote-member-dialog-model.ts
domienderle: I get it. Isn't possible otherwise. Thanks.

domienderle (changes requested):
- Good call with toast. Looks good.
- If the backend is slow and you try to promote a member, there is a loading spinner for the promoting in the `AlertDialog`. But as soon as the `AlertDialog` is closed, the role button stays as it is, until the response of the backend is received. I would expect a loading spinner for this button as well, to not be able to click the button again, if the promotion was already executed.

Slartibartfass2 (commented):
What are your thoughts about this solution:
image

As we use the reloading at several spots, I thought this might be the best solution.

domienderle (changes requested):
The `Promote member` button is still not disabled as long as the members are reloaded. After that we should be go to go.

domienderle (dismissed):

domienderle (approved):

Issue #52, opened 2024-10-31
[Use Case]: Decide On Paper


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to decide on a paper, i.e. review it.

### Acceptance criteria

- provide buttons for the decision options
- the project member can choose between "Yes", "No" and "Maybe" (if "Maybe" is allowed in the project settings)
- when the project member presses one of the button (or its shortcut) the decision is stored and the view navigates to the next paper

### Anything else?

Related to #34

PR #193, opened 2025-02-06
Open project view

### LINKED ISSUE #30: [Use Case]: Open Project View

Project Member

### Describe the use case

As project member, I want to open the project view for a specific project.

### Acceptance criteria

- show project view page

### Anything else?

Related to #31, #32, #33, #34, #35, #36, #37, #38, #39, #40, #50 and #51

Merseleo:
Closes #30

## What I have made

- added first version of project dashboard page including:
  - the open reviews list of the project
  - a section for the project information as a "long" text, whereas the project and stage start are statically set and not loaded from the server, as neither the api nor the mock backend provide this opportunity (see [here](https://github.com/SE-UUlm/snowballr-api/issues/7) for more information)
  - a section for the progress of the current stage using a donut chart and a legend displaying all the data (please add a review comment if further animation or hover information or ... for the donut chart are wanted)
- the third section for the project activity was not developed as this section require a lot of work, especially on the API and (mock backend), but is addressed in #240
- adapt layout of named list to the layout from the mockup 
- dynamically load current user (especially in the top layer `layout.ts`) to work with data from mock backend
- moved some components in separate directories, so the components are more organized

## Checklist

  - [x] unit tests
  - [x] integration tests


src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -1,9 +1,10 @@
 <script lang="ts">
     import { Progress } from "$lib/components/primitives/progress";
     import { getNames } from "$lib/utils/common-helper";
-    import type { ProjectListEntryInterface } from "$lib/model/general";
     import { ProjectStatus } from "$lib/model/api/project";
 
+    import type { ProjectListEntryInterface } from "$lib/model/component-interfaces";
```
Slartibartfass2: Minor
```suggestion
    import { ProjectStatus } from "$lib/model/api/project";
    import type { ProjectListEntryInterface } from "$lib/model/component-interfaces";
```

src/lib/components/composites/select/StageEntry.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
     import PaperListEntry from "$lib/components/composites/paper-components/PaperListEntry.svelte";
-    import * as Accordion from "$lib/components/primitives/accordion/index.js";
-    import Button from "$lib/components/primitives/button/button.svelte";
+    import * as Accordion from "$lib/components/primitives/accordion";
+    import Button from "../../primitives/button/button.svelte";
```
Slartibartfass2: What are the reasons for these changes? Did you/your IDE had trouble with these imports?

tests/integration/utils/paper-bookmark-button.test.ts
Slartibartfass2, line 1: `PaperBookmarkButton.svelte` is not in the utils directory. Either move the component in the same directory or this test back to the parent directory.

vite.config.ts
```diff
@@ -19,7 +19,7 @@ export default defineConfig({
                 "src/routes/**/+*.{svelte,ts}",
                 "src/lib/components/primitives/**",
                 "**/*.d.ts",
-                "src/lib/model/api/**",
+                "src/lib/model/**",
```
Slartibartfass2, line 22: I think this fine for the types, but maybe we should move the functions somewhere else.

src/lib/utils/common-helper.ts
```diff
@@ -71,10 +72,57 @@ function exhaustiveCheck(x: never): never {
  * @param count - the number of items
  * @param singular - the singular form of the word
  * @param plural - the plural form of the word
- * @returns the singular form if count is 1, otherwise the plural form
+ * @return the singular form if count is 1, otherwise the plural form
```
Slartibartfass2: `@return` is not a valid TSDoc tag. We should consider adding the ESLint TSDoc plugin to get rid of such uncertainties.

src/lib/model/general.ts
```diff
@@ -1,52 +1,20 @@
...
-    type Project_Member_List,
-    type Project_Statistics,
-} from "$lib/model/api/project";
+import { Project_Paper } from "$lib/model/api/project";
 import type { Criterion } from "./api/criterion";
 import type { Paper } from "./api/paper";
 import type { Review } from "./api/review";
 
+/*
+Contains general types that are used in several components or pages but cannot be categorised anywhere else.
```
Slartibartfass2: Now that I have my own spell checker:
It's "categorized" not "categorised". The latter is British English.

Slartibartfass2: These "region" comments look a bit malformed. Are there any alternatives?

src/lib/model/error-codes.ts
```diff
@@ -1,4 +1,6 @@
-// gRPC status (= error) codes, see https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+/*
+ gRPC status (= error) codes, see https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+ */
```
Slartibartfass2: Use the correct format to apply the comment to the enum type.
```suggestion
/**
 * gRPC status (= error) codes, see https://grpc.github.io/grpc/core/md_doc_statuscodes.html
 */
```

src/lib/components/composites/utils/Tooltip.svelte
```diff
@@ -3,8 +3,8 @@
         buttonVariants,
         type ButtonSize,
         type ButtonVariant,
-    } from "$lib/components/primitives/button/index.js";
-    import * as Tooltip from "$lib/components/primitives/tooltip/index.js";
+    } from "$lib/components/primitives/button";
+    import * as Tooltip from "$lib/components/primitives/tooltip";
```
Slartibartfass2: Same here as above, are these changes in any way necessary?

src/routes/+layout.ts
```diff
@@ -1,16 +1,18 @@
...
+    try {
+        const user: User = await backendService.getCurrentUser(Nothing).response;
+        return {
+            user,
+        };
+    } catch (err) {
+        console.error(`Current user could not be loaded ${err}`);
+        /// TODO: find better error code
+        error(500, "Could not load current user!");
+    }
```
Slartibartfass2: This just shows an error when a user enters our page. What's the purpose of that?

Didn't we agree on fixing this in the sign in PR, as soon as the mock backend ready to use?

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -0,0 +1,110 @@
...
+async function groupProjectPapersInStageByDecision(
+    project: Project,
+): Promise<{ [key: string]: Project_Paper[] }> {
+    const papersOfStage = await backendService
+        .getAllProjectPapersForProject({ id: project.id })
+        .response.then((papers) =>
+            papers.projectPapers.filter(
+                (projectPaper) => projectPaper.stage === project.currentStage,
+            ),
+        );
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts:20]: Add a catch to this request, so the server doesn't shut down when an error appears.

Comment (Slartibartfass2) [reply src/routes/project/[projectId]/dashboard/+page.ts:20]: I'm actually not sure if this is the only point. Run the frontend without the mock backend and navigate to this page. The server will immediately shut down.

src/lib/components/composites/list/NamedList.svelte
```diff
@@ -69,7 +69,7 @@ Otherwise the error message is shown.
 You can render additional content between the title and the list by providing `preListContent`.
 This can be e.g. a search bar.
 -->
-<div class="flex h-full w-full flex-col gap-y-5 overflow-hidden px-5">
+<div class="flex h-full w-full flex-col gap-y-5 overflow-hidden p-5">
```
Slartibartfass2: The vertical padding leads to a misalignment on the homepage. Are there alternatives on how we can fix this?

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -0,0 +1,110 @@
...
+        stage: project.currentStage,
+        decisions: decisions,
+    };
+}
+
+/**
+ * Loads project information and open reviews for the currently opened project.
+ */
+export const load: PageLoad = async ({ params, parent }) => {
+    const { loadingProject } = await parent();
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]: What about
```suggestion
    const loadingProject = parent().then((project) => project.loadingProject);
```
Otherwise, we could risk waterfalls as the documentation of this function states.

src/routes/+page.ts
```diff
@@ -38,7 +42,7 @@ async function requestUndecidedPapers(project: Project): Promise<PaperListEntryI
  * - open reviews from the project
  */
 export const load: PageLoad = async () => {
-    const thisUserId = "0";
+    const thisUserId = (await backendService.getCurrentUser(Nothing).response).id;
     const allUserProjects = backendService.getAllProjectsForUser({ id: thisUserId }).response;
```
Slartibartfass2: Make this asynchronous to not block the page loading.
```suggestion
    const allUserProjects = backendService
        .getCurrentUser(Nothing)
        .response.then((user) => backendService.getAllProjectsForUser({ id: user.id }).response);
```

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -0,0 +1,110 @@
...
+            case PaperDecision.ACCEPTED:
+                return "Accepted";
+            case PaperDecision.DECLINED:
+                return "Declined";
+            case PaperDecision.UNDECIDED:
+            case PaperDecision.UNSPECIFIED:
+                return projectPaper.reviews.length > 0 ? "Undecided" : "Not reviewed";
+            default:
+                exhaustiveCheck(projectPaper.decision);
+        }
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]: There's a duplicate of this in `src/lib/components/composites/criteria/PaperDecisionBanner.svelte`. Let's extract this to a globally accessible method.

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -0,0 +1,110 @@
...
+    });
+}
+
+async function requestProjectInformation(project: Project): Promise<ProjectInformationInterface> {
+    // TODO: add a real call to the backend (extend API + extend mock backend)
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]: Is there an issue for this? If so, refer it in here.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/dashboard/+page.ts]: At the same moment, we can create an enum for the resulting text.

src/lib/utils/common-helper.ts
```diff
@@ -71,10 +72,57 @@ function exhaustiveCheck(x: never): never {
...
+ * @param list the list to be grouped
+ * @param keySelector function that map a (list) item to a certain key
+ * @return the grouped list as an object with the association key: \<list of items associated to key\>
+ */
+function groupBy<T>(list: T[], keySelector: (arg0: T) => string): { [key: string]: T[] } {
+    return list.reduce((result: { [key: string]: T[] }, item: T) => {
+        (result[keySelector(item)] ??= []).push(item);
+        return result;
+    }, {});
+}
```
Slartibartfass2: Use the `Record` type for such cases.
```suggestion
function groupBy<T>(list: T[], keySelector: (arg0: T) => string): Record<string, T[]> {
    return list.reduce((result: Record<string, T[]>, item: T) => {
        (result[keySelector(item)] ??= []).push(item);
        return result;
    }, {});
}
```

src/lib/utils/common-helper.ts
```diff
@@ -71,10 +72,57 @@ function exhaustiveCheck(x: never): never {
...
+        case "Declined":
+            return "text-decline-red";
+        case "Undecided":
+            return "text-maybe-yellow";
+        case "Not reviewed":
+            return "text-unreviewed-gray";
+        default:
+            exhaustiveCheck(status);
+    }
+}
```
Slartibartfass2: This is also very similar to the code in the `PaperDecisionBanner`. Do you think we can find a solution that covers both cases?

src/lib/components/composites/statistics-components/ProjectInformation.svelte
```diff
@@ -0,0 +1,56 @@
...
+-->
+<div class="w-full px-5 py-2 break-words" data-testid="project-information">
+    {#await projectInformation}
+        <div class="flex flex-col gap-y-3">
+            {#each { length: 4 }}
+                <Skeleton class="flex h-6 rounded-full" />
+            {/each}
+        </div>
+    {:then information}
+        The project <span class="text-emphasized">{information.projectName}</span> started at
```
Slartibartfass2: We display the project name in the navigation bar and in the tab title. Do you think we need to display it here again?

src/lib/components/composites/statistics-components/ProjectInformation.svelte
```diff
@@ -0,0 +1,56 @@
...
+        <span class="text-emphasized">
+            {information.projectStart.toLocaleString().split(",")[0]}
```
Slartibartfass2: Is this what you're looking for?
```suggestion
            {information.projectStart.toLocaleDateString()}
```

src/lib/components/composites/statistics-components/ProjectInformation.svelte
```diff
@@ -0,0 +1,56 @@
...
+        </span>
+        and you are currently in stage
+        <span class="front-medium relative top-1 text-3xl">{information.projectStage}</span>. You
```
Slartibartfass2: This looks a bit too big in my opinion. Isn't the `text-emphasized` enough?

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Slartibartfass2:
```suggestion
        The project <span class="text-emphasized">{information.projectName}</span> started on
```

src/lib/components/composites/statistics-components/ProjectInformation.svelte
```diff
@@ -0,0 +1,56 @@
...
+        are working
+        <span class="text-emphasized">{information.daysInStage}</span>
+        {pluralize(information.daysInStage, "day", "days")} in this stage and reviewed
+        <span class="text-emphasized">{information.reviewedPapersInStage}</span>
+        /
+        <span class="text-emphasized">{information.totalPapersInStage}</span>
+        {pluralize(information.totalPapersInStage, "paper", "papers")} so far, so your estimated remaining
+        time {pluralize(information.estimatedRemainingDays, "is", "are")}
```
Slartibartfass2: time is singular
```suggestion
        time is
```

src/lib/components/composites/statistics-components/ProjectInformation.svelte
```diff
@@ -0,0 +1,56 @@
...
+        <span class="text-emphasized">{information.estimatedRemainingDays}</span>
+        {pluralize(information.estimatedRemainingDays, "day", "days")}.
```
Slartibartfass2: What about:
"You are currently working 2 days in this stage and have reviewed 7 / 14 papers so far. Based on your progress, your estimated remaining time will be 2 days."

src/lib/components/composites/statistics-components/StageProgress.svelte
```diff
@@ -0,0 +1,57 @@
...
+Usage:
+```svelte
+    <StageProgress
+        {stageProgress}
+    />
+```
+-->
+<div class="flex flex-row items-center gap-x-4 p-5" data-testid="stage-progress">
+    {#await stageProgress}
+        <Skeleton class="h-[150px] w-[150px] rounded-full" />
```
Slartibartfass2: Minor suggestion:
```suggestion
        <Skeleton class="size-37 rounded-full" />
```

src/lib/components/composites/statistics-components/StageProgress.svelte
```diff
@@ -0,0 +1,57 @@
...
+        <StageProgressChart {decisions} {stage} />
+        <ul class="mx-10 space-y-3">
+            {#each Object.entries(decisions) as [decision, number] (decision)}
+                <li class="flex flex-row items-center">
+                    <Circle
+                        class="mr-2 {getStatusColor(decision as PaperStatus)}"
+                        size={14}
+                        strokeWidth={4}
+                    />
+                    <span class="text-default-sb min-w-[150px]">{decision}:</span>
```
Slartibartfass2: Is it important that it's exactly 150px?
```suggestion
                    <span class="text-default-sb min-w-38">{decision}:</span>
```

src/lib/components/composites/statistics-components/StageProgress.svelte
```diff
@@ -0,0 +1,57 @@
...
+    {:then { stage, decisions }}
+        <StageProgressChart {decisions} {stage} />
+        <ul class="mx-10 space-y-3">
+            {#each Object.entries(decisions) as [decision, number] (decision)}
+                <li class="flex flex-row items-center">
+                    <Circle
+                        class="mr-2 {getStatusColor(decision as PaperStatus)}"
+                        size={14}
+                        strokeWidth={4}
+                    />
```
Slartibartfass2: What do you think about also filling the circle?
image

src/lib/components/composites/statistics-components/StageProgressChart.svelte
```diff
@@ -0,0 +1,65 @@
...
+    const { stage, decisions }: StageProgressInterface = $props();
+
+    const totalNumberOfDecisions = Object.values(decisions).reduce((acc, cur) => acc + cur, 0);
+    const segments: Segment[] = Object.entries(decisions).map(([key, numOfDecisions]) => ({
+        decision: key as PaperStatus,
+        value: numOfDecisions / totalNumberOfDecisions,
+    }));
+
+    const WIDTH = 150;
+    const HEIGHT = 150;
```
Slartibartfass2: What about just `SIZE`? It should always be a circle, doesn't it?

src/lib/components/composites/statistics-components/StageProgressChart.svelte
```diff
@@ -0,0 +1,65 @@
...
+    <text
+        class="text-default"
+        dominant-baseline="middle"
+        dy="-0.75em"
+        text-anchor="middle"
+        x="50%"
+        y="50%"
+    >
+        Stage
+        <tspan class="text-3xl" dy="1.25em" x="50%">{stage}</tspan>
```
Slartibartfass2: What about
```suggestion
        <tspan class="text-3xl" dy="1em" x="50%">{stage}</tspan>
```

tests/unit/helpers/common-helper.test.ts
```diff
@@ -92,3 +93,38 @@ describe("Pluralize a word based on the count", () => {
         expect(pluralize(0, "item", "items")).toBe("items");
     });
 });
+
+describe("Group items of a list by a key (function)", () => {
+    it("When list is empty,then no items are grouped", () => {
```
Slartibartfass2:
```suggestion
    it("When list is empty, then no items are grouped", () => {
```

tests/unit/helpers/common-helper.test.ts
```diff
@@ -92,3 +93,38 @@ describe("Pluralize a word based on the count", () => {
...
+        ];
+        expect(groupBy(list, (i) => i.type)).toStrictEqual({
+            a: [
+                { type: "a", value: 1 },
+                { type: "a", value: 2 },
+                { type: "a", value: 1 },
+            ],
+            b: [{ type: "b", value: 3 }],
+        });
+        expect(groupBy(list, (i) => "" + i.value)).toStrictEqual({
```
Slartibartfass2: `"" + i.value` looks like there should be something inside "".
What about
```suggestion
        expect(groupBy(list, (i) => `${i.value}`)).toStrictEqual({
```
or `i.value.toString()`

tests/integration/statistics-components/project-information.test.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test, describe } from "vitest";
+import { render, screen } from "@testing-library/svelte";
+import ProjectInformation from "$lib/components/composites/statistics-components/ProjectInformation.svelte";
+import { waitForComponentLoading } from "../test-helper";
+import { loading } from "../../model-builder";
+
+describe("ProjectInformationComponent", () => {
```
Slartibartfass2: Until now, we only used the component name as test suite name, is there a reason why you treat it differently now?
```suggestion
describe("ProjectInformation", () => {
```

Slartibartfass2: Same in `stage-progress.test.ts`

tests/integration/statistics-components/project-information.test.ts
Slartibartfass2, line 1: Is there a reason why you named the directory `statistics-components`?
We have some directories with the `-components` suffix and some without.

I still think the suffix is unnecessary :D

Slartibartfass2 (changes requested):
Looks great.
Do you think we need to adjust the page, when it's smaller, i.e., <`md`?

The most important part is to fix the page load so that the server doesn't crash.
I think the lighthouse job fails because the home page returns 500, but I already commented on that part.

src/lib/components/composites/project-components/ProjectListEntry.svelte
Merseleo: I fixed it.

src/lib/components/composites/select/StageEntry.svelte
Merseleo: No it had no troubles but I also can't remember to changed this manually, so I am bit confused too. But I changed it back.

tests/integration/utils/paper-bookmark-button.test.ts
Merseleo: I accidentally moved the test, imo the `PaperBookmarkButton` is not a helper or utility component and shouldn't be located in the utils directory. I moved the test back to the parent directory.

src/lib/utils/common-helper.ts
Merseleo: Ah, I always thought it is exactly the other way around and return is not valid. I will fix this.

src/lib/model/general.ts
Merseleo: I can make them TS-Doc comments, what do you think of it now?

src/lib/model/error-codes.ts
Merseleo: Fixed it.

src/lib/components/composites/utils/Tooltip.svelte
Merseleo: Not and also not intended.

Merseleo: But I think WebStorm automatically shortens these imports as the "/index.js" is not necessary. However, I will change it back.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: Not directly, there is an issue in the API repo to extend the API, but no extra issue in the mock backend repo and the frontend repo. I would add these issues, if somebody is working on this issue and we are sure, that we only introduce this single call and do not combine with something similar.

src/lib/utils/common-helper.ts
Merseleo: Good idea, didn't think of it.

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: I've just taken the text from the mockup because we added it there. But I agree with you that you could also leave it out. Should I delete it (which I would be totally fine with)?

Merseleo: I fixed this typo (luckily there are no more english grammar grades for us)

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: Same as above ... I fix it.

tests/unit/helpers/common-helper.test.ts
Merseleo: Changed it.

tests/unit/helpers/common-helper.test.ts
Merseleo: Yes I agree, that this cast isn't the best solution, I changed it to your first suggestion.

tests/integration/statistics-components/project-information.test.ts
Merseleo: No, that wasn't intentional, I'm changing it so that it's consistent.

tests/integration/statistics-components/project-information.test.ts
Merseleo: I agree as the parent directory already is called components, I will delete the suffix, ok?

vite.config.ts
Merseleo: What do you suggest? Because the two tiny functions that are located in the model directory aren't tested anyway?!

src/routes/+layout.ts
Merseleo: I am not sure, whether we agreed on this, I only added this to test the different project dashboards for different users (I changed the authenticated user in the mock backend). 

But now I understand your comment. Actually we need a redirect to the login page when the user could not be loaded, correct? 
Still, I would prefer to load the current user from the backend this was useful for testing, but I would add a comment, that the error should be exchanged with a redirect. Are you ok with that?

src/lib/components/composites/list/NamedList.svelte
Merseleo: What misalignment do you mean, actually I only added the padding as this was done in the mockup as well?!

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts:20]: No it is not the only point. The thing with the dynamic user loading was not ideal, as I throw a 500 internal error, if no user could be loaded and then everything stops working.

Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts:20]: However, I will add the catch

src/routes/+layout.ts
Merseleo: What do you think of the updated version? I request the user from the mock backend and if there no mock backend running but fallback to this default user. This should prevent the 500

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: You are correct, didnt consider this.

src/routes/+page.ts
Merseleo: And add the default catch after it, so the page does not crash ... oh oh oh, I really was not concentrated.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: I agree to the first point, but the second? Why make a enum for this text, especially when we only use it in this method?

Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: I have seen that I have even already made a literal type for it.

src/lib/utils/common-helper.ts
Merseleo: What do you think of the current solution? Or do want to return only the color and manually prefix the color instead of replacing text with the type we want? Imo the new solution is good.

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: Yes, thank you!

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: I just implemented the mockup, but I agree, that it is big and I am totally fine to use `text-emphasized`. Looks good too, imo.

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: Sounds better, great idea.

src/lib/components/composites/statistics-components/StageProgress.svelte
Merseleo: Actually we can use size as well, correctly, as it is always a circle.

src/lib/components/composites/statistics-components/StageProgress.svelte
Merseleo: No same here, I only reused the widths etc. from the mockup, but I agree that it seems cleaner to use w-38. I will change it.

src/lib/components/composites/statistics-components/StageProgress.svelte
Merseleo: A great idea, again. I like it a bit more than the design before. I will change it.

src/lib/components/composites/statistics-components/StageProgressChart.svelte
Merseleo: I changed it.

src/lib/components/composites/statistics-components/StageProgressChart.svelte
Merseleo: A question of interest. What browser do you use for testing and can you provide a image, how it looks on your machine, because `1em` looks very "squeezed" when I test it.

Merseleo (commented):
Thank you for the quick and careful feedback. I hope I didn't forgot to answer something. 
Regarding your question with the responsive design for smaller screen widths. Do you have any idea? I would try to layout things beneath instead of next to other and provide it with the next review round, but this is the only idea I have.

Merseleo:
> Thank you for the quick and careful feedback. I hope I didn't forgot to answer something. Regarding your question with the responsive design for smaller screen widths. Do you have any idea? I would try to layout things beneath instead of next to other and provide it with the next review round, but this is the only idea I have.

And I am not very happy with this design. Maybe I have to found another solution for the spacing so everything "flows" more when expanding or shrinking the page ...

tests/integration/statistics-components/project-information.test.ts
Slartibartfass2: Thanks :))

src/lib/components/composites/statistics-components/StageProgressChart.svelte
Slartibartfass2: As discussed, it looks way different.
```svelte
<text class="text-3xl" dominant-baseline="middle" text-anchor="middle" x="50%" dy="3em">
    {stage}
</text>
```
This worked for me.

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Slartibartfass2: Yes, I think "The project started ..." is totally fine.

src/lib/utils/common-helper.ts
Slartibartfass2: I think a prefix parameter would be great. Replacing the prefix seems ugly.

src/lib/components/composites/list/NamedList.svelte
Slartibartfass2: You can see it, on the home page. The end of the list on the left and the create project button are not aligned. You can also see it in the references and citations tab on the paper view, where the padding is humongous.

Maybe just wrap it in the project dashboard page.

src/routes/+layout.ts
Slartibartfass2: This is fine now.

vite.config.ts
Slartibartfass2: It's a general question, whether we want to keep functions inside the model directory. We may want to move them in the future, or we shouldn't any more functions there, as it doesn't fit into a model directory.

Slartibartfass2 (changes requested):
We've come a long way. Only some smaller discussions left.

src/lib/components/composites/criteria/PaperDecisionBanner.svelte
```diff
@@ -6,8 +6,8 @@
     import UserAvatar from "../user-avatar/UserAvatar.svelte";
     import { Skeleton } from "$lib/components/primitives/skeleton";
```
Merseleo, line 7: gre

vite.config.ts
Merseleo: In my opinion there should be no functions in the model directory as this directory should only contain files that contain types and interfaces (so the model of data) and helper functions should be in the utils directory. 
Are you ok with moving the two functions and keep the config for the coverage?

src/lib/components/composites/list/NamedList.svelte
Merseleo: Yes, I see it now, good catch. I will implement your suggestion.

src/lib/components/composites/statistics-components/ProjectInformation.svelte
Merseleo: Ok.

src/lib/components/composites/statistics-components/StageProgressChart.svelte
Merseleo: I will integrate your suggestion.

src/lib/utils/common-helper.ts
Merseleo: I agree, I will try something.

Merseleo (commented):
Did you check my responsive design? And what is your opinion on this change?

vite.config.ts
Slartibartfass2: Yes, that's what I would've suggested too.

src/lib/utils/common-helper.ts
Slartibartfass2: Now it looks like this:
image

I already experienced something similar, and I think I finally know why this happens: Tailwind searches the source files for all class names that are used and only exports them to keep the CSS as small as possible.

When dynamically creating class names, the names are not exported. They might be already exported from other places, but we cannot be sure. AFAIK, we didn't consider this yet, but should.
You can read more [here](https://tailwindcss.com/docs/detecting-classes-in-source-files).

Slartibartfass2 (commented):
One thing to fix left. Also, don't forget to rebase.

src/lib/utils/common-helper.ts
Merseleo: Yes, I have noticed this problem before, but I overlooked it at the time, very thoughtful of you.

src/lib/utils/common-helper.ts
```diff
@@ -77,4 +78,90 @@ function pluralize(count: number, singular: string, plural: string): string {
...
+        fill: "fill-maybe-yellow",
+        bg: "bg-maybe-yellow",
+    },
+    "Not reviewed": {
+        "": "unreviewed-gray",
+        text: "text-unreviewed-gray",
+        fill: "fill-unreviewed-gray",
+        bg: "bg-unreviewed-gray",
+    },
+};
```
Slartibartfass2: There's no class without any prefix
```suggestion
type colorPrefix = "text" | "bg" | "fill";
const statusColors: Record<PaperStatus, Record<colorPrefix, string>> = {
    Accepted: {
        text: "text-accept-green",
        fill: "fill-accept-green",
        bg: "bg-accept-green",
    },
    Declined: {
        text: "text-decline-red",
        fill: "fill-decline-red",
        bg: "bg-decline-red",
    },
    Undecided: {
        text: "text-maybe-yellow",
        fill: "fill-maybe-yellow",
        bg: "bg-maybe-yellow",
    },
    "Not reviewed": {
        text: "text-unreviewed-gray",
        fill: "fill-unreviewed-gray",
        bg: "bg-unreviewed-gray",
    },
};
```

Slartibartfass2: In `PaperListEntry` we use a very similar logic in `getReviewDecisionColor`. Please replace it with `getStatusColor(getStatusText(paper), "border")`.
We probably could also use this for the decision buttons, but changing these would go too far.

Slartibartfass2 (changes requested):
Just this small change.

src/lib/utils/common-helper.ts
Merseleo: Good catch.

And regarding the decision buttons. You are right, maybe this could be reused, but we can also do this when implementing #52.

Slartibartfass2 (approved):
🚀

PR #204, opened 2025-02-10
Added backend api mocks

### LINKED ISSUE #203: [Feature]: Add test mocks for grpc backend

_No response_

Currently, when a gRPC call is used in a test, the error is caught and printed to `stderr` (as it should be) but sometimes the result is necessary for testing. Therefore, we need to mock those calls to test components that use those calls to the API inside them.

Use vitest to mock those calls.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #203 

## What I have made

- added mocks for backend API calls used by the integration tests

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only changes in test setup)
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - ~~[ ] end-to-end tests~~

### DIFF
--- a/tests/integration/paper-components/paper-list-entry.test.ts
+++ b/tests/integration/paper-components/paper-list-entry.test.ts
@@ -81,8 +81,9 @@ describe("PaperListEntryComponent", () => {
         expect(screen.getByText("An Analysis of TypeScript Performance")).toBeInTheDocument();
         expect(screen.getByText("John Doe, Bob Johnson")).toBeInTheDocument();
 
+        expect(screen.getByRole("button").childElementCount).toBe(2);
         expect(screen.getByRole("button").children[0]).toHaveClass("border-l-4 border-decline-red");
-        expect(screen.getByRole("button").childElementCount).toBe(1);
+        expect(screen.getByRole("button").children[1]).toHaveTextContent("JD");
     });
 
     test("When the user provides a custom onclick function, then it is executed on a single click (and not on double click)", async () => {
--- a/tests/setupTest.ts
+++ b/tests/setupTest.ts
@@ -91,7 +91,7 @@ vi.mock("$app/stores", (): typeof stores => {
     };
 });
 
-// This type takes the SnowballRClient interface and returns a new type where each function returns a MockUnaryCall i.e. a mocked backend API
+// This type takes the SnowballRClient interface and returns a new "interface" type where each api function returns a MockUnaryCall, i.e. a mocked API call, instead of a normal UnaryCall.
 type MockReturnType<T> = {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     [K in keyof T]: T[K] extends (...args: infer A) => UnaryCall<infer _I, infer R>
@@ -104,7 +104,7 @@ interface MockUnaryCall<T> {
 }
 
 /**
- * Mock a backend API call
+ * Mocks a backend API call
  *
  * Usage:
  * ```ts
@@ -112,7 +112,7 @@ interface MockUnaryCall<T> {
  * getAllUsers: mock(Object.values(Users)),
  * ```
  *
- * @param fn - Either a function that takes an argument and returns a value or a constant value
+ * @param fn - Either a function that takes an argument and returns a value or a constant value that is returned
  * @returns A mocked backend API call
  */
 function mock<T, R>(fn: ((arg: T) => R) | R): Mock<(input: T) => MockUnaryCall<R>> {
@@ -145,7 +145,7 @@ vi.mock("$lib/grpc-api", () => {
             getAllUsers: mock({ users: Object.values(Users) }),
             getCurrentUser: mock(Users.johnDoe),
             getUserById: mock(({ id }) => createUser({ id })),
-            getUserByEmail: mock(({ id }) => createUser({ email: id })),
+            getUserByEmail: mock(({ id: email }) => createUser({ email: email })),
             updateUser: vi.fn(),
             softDeleteUser: vi.fn(),
             softUndeleteUser: vi.fn(),
--- a/tests/setupTest.ts
+++ b/tests/setupTest.ts
@@ -1,11 +1,22 @@
 /* Test setup file, inspired by https://github.com/wd-David/svelte-component-test-recipes?tab=readme-ov-file#setuptestts */
 import * as matchers from "@testing-library/jest-dom/matchers";
-import { expect, vi } from "vitest";
+import { expect, vi, type Mock } from "vitest";
 import type { Navigation, Page } from "@sveltejs/kit";
 import { readable } from "svelte/store";
 import * as environment from "$app/environment";
 import * as navigation from "$app/navigation";
 import * as stores from "$app/stores";
+import {
+    createCriterion,
+    createPaper,
+    createProject,
+    createProjectPaper,
+    createReview,
+    createUser,
+} from "./model-builder";
+import { Criteria, Members, Papers, ProjectPapers, Projects, Reviews, Users } from "./example-data";
+import type { ISnowballRClient } from "$lib/model/api/main.client";
+import type { UnaryCall } from "@protobuf-ts/runtime-rpc";
 
 // Add custom jest matchers
 expect.extend(matchers);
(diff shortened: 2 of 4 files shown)


tests/integration/paper-components/paper-list-entry.test.ts
```diff
@@ -81,8 +81,9 @@ describe("PaperListEntryComponent", () => {
         expect(screen.getByText("An Analysis of TypeScript Performance")).toBeInTheDocument();
         expect(screen.getByText("John Doe, Bob Johnson")).toBeInTheDocument();
 
+        expect(screen.getByRole("button").childElementCount).toBe(2);
         expect(screen.getByRole("button").children[0]).toHaveClass("border-l-4 border-decline-red");
-        expect(screen.getByRole("button").childElementCount).toBe(1);
+        expect(screen.getByRole("button").children[1]).toHaveTextContent("JD");
```
Merseleo, line 86: Good idea, thank you (close this conversation on your own)

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
         updated,
     };
 });
+
+// This type takes the SnowballRClient interface and returns a new type where each function returns a MockUnaryCall i.e. a mocked backend API
```
Merseleo:
```suggestion
// This type takes the SnowballRClient interface and returns a new "interface" type where each api function returns a MockUnaryCall, i.e. a mocked API call, instead of a normal UnaryCall.
```

I needed some time to understand this comment (although I knew, what this type is used for. What do you think of my suggestion? Or do you have other ideas?

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+ *
+ * Usage:
+ * ```ts
+ * getUserById: mock(({ id }) => createUser({ id })),
+ * getAllUsers: mock(Object.values(Users)),
+ * ```
+ *
+ * @param fn - Either a function that takes an argument and returns a value or a constant value
+ * @returns A mocked backend API call
+ */
```
Merseleo: This comment was very helpful, great 👍

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+        ? Mock<(...args: A) => MockUnaryCall<R>>
+        : T[K];
+};
+type MockApi = MockReturnType<ISnowballRClient>;
+interface MockUnaryCall<T> {
+    response: Promise<T>;
+}
+
+/**
+ * Mock a backend API call
```
Merseleo:
```suggestion
 * Mocks a backend API call
```

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+ *
+ * Usage:
+ * ```ts
+ * getUserById: mock(({ id }) => createUser({ id })),
+ * getAllUsers: mock(Object.values(Users)),
+ * ```
+ *
+ * @param fn - Either a function that takes an argument and returns a value or a constant value
```
Merseleo:
```suggestion
 * @param fn - Either a function that takes an argument and returns a value or a constant value that is returned
```

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            };
+        } else {
+            return {
+                response: Promise.resolve(fn),
+            };
+        }
+    });
+}
+
+// Mock Backend API
```
Merseleo, line 133: Maybe it could be helpful to inform the reader about the usage? So that this is a mock of the API used only for integration tests and not for E2E tests, wherefore we have additionally a more complex mocked backend with API? Or do you think this isn't necessary as this is / should be already documented in the README of this project?

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+vi.mock("$lib/grpc-api", () => {
+    const mockBackend: { backendService: MockApi } = {
+        backendService: {
+            getAvailableFetcherApis: mock({ fetcherApis: [] }),
```
Merseleo: Why no return already a list with some dummy fetcher apis?

tests/setupTest.ts
Merseleo, line 1: One general question: you implemented some mocked api calls, others not. Do you think we should outsource this work and do it later?

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            logout: vi.fn(),
+            isAuthenticated: mock({ value: true }),
+            renewSession: vi.fn(),
+            requestPasswordReset: vi.fn(),
+            resetPassword: vi.fn(),
+            changePassword: vi.fn(),
+            getAllUsers: mock({ users: Object.values(Users) }),
+            getCurrentUser: mock(Users.johnDoe),
+            getUserById: mock(({ id }) => createUser({ id })),
+            getUserByEmail: mock(({ id }) => createUser({ email: id })),
```
Merseleo:
```suggestion
            getUserByEmail: mock(({ id: email }) => createUser({ email: email })),
```

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            updateUser: vi.fn(),
+            softDeleteUser: vi.fn(),
+            softUndeleteUser: vi.fn(),
+            getAllPapersToReview: mock({ projectPapers: Object.values(ProjectPapers) }),
+            getPapersToReviewForProject: mock({ projectPapers: Object.values(ProjectPapers) }),
```
Merseleo: Maybe you could easily filter these values here with the ID

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            resetPassword: vi.fn(),
+            changePassword: vi.fn(),
+            getAllUsers: mock({ users: Object.values(Users) }),
+            getCurrentUser: mock(Users.johnDoe),
+            getUserById: mock(({ id }) => createUser({ id })),
+            getUserByEmail: mock(({ id }) => createUser({ email: id })),
+            updateUser: vi.fn(),
+            softDeleteUser: vi.fn(),
+            softUndeleteUser: vi.fn(),
+            getAllPapersToReview: mock({ projectPapers: Object.values(ProjectPapers) }),
```
Merseleo: Maybe you could filter here for PaperDecision.UNDECIDED?

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            getPapersToReviewForProject: mock({ projectPapers: Object.values(ProjectPapers) }),
+            getUserSettings: vi.fn(),
+            updateUserSettings: vi.fn(),
+            getReadingList: mock({ papers: Object.values(Papers) }),
+            isPaperOnReadingList: vi.fn(),
+            addPaperToReadingList: vi.fn(),
+            removePaperFromReadingList: vi.fn(),
+            getPendingInvitationsForUser: mock({ projects: Object.values(Projects) }),
+            inviteUserToProject: vi.fn(),
+            getPendingInvitationsForProject: mock({ users: Object.values(Users) }),
```
Merseleo: Maybe only return on user and project?

tests/setupTest.ts
```diff
@@ -79,3 +90,114 @@ vi.mock("$app/stores", (): typeof stores => {
...
+            getProjectMembers: mock({ members: Object.values(Members) }),
+            removeProjectMember: vi.fn(),
+            getAllProjects: mock({ projects: Object.values(Projects) }),
+            getAllDeletedProjects: mock({ projects: Object.values(Projects) }),
+            getAllDeletedProjectsForUser: mock({ projects: Object.values(Projects) }),
+            getAllArchivedProjects: mock({ projects: Object.values(Projects) }),
+            getAllProjectsForUser: mock({ projects: Object.values(Projects) }),
+            getAllArchivedProjectsForUser: mock({ projects: Object.values(Projects) }),
```
Merseleo: Same here, don't we want to filter them a bit more?

Merseleo (changes requested):
Very good idea to mock the API calls at this level, especially for the integration tests. I should have done this before for the ProjectCreationDialog, which needed this. 
So maybe we can extend the tests for this component afterwards.

tests/setupTest.ts
Slartibartfass2: Yes, I focussed on the get requests. I think this is fine as is for now. Probably mocking other requests makes more sense on a component test level, i.e., writing a setup method, which mocks all API calls that are used by a component. This way, the mock can behave to the specific needs of the component/test suite.

What's your opinion on that?

tests/setupTest.ts
Slartibartfass2: Sounds better, thank you!

tests/setupTest.ts
Slartibartfass2: Changed it.

tests/setupTest.ts
Slartibartfass2: Changed it.

tests/setupTest.ts
Slartibartfass2: I'm glad it helped to understand this.

tests/setupTest.ts
Slartibartfass2: I added more explanation, please provide feedback on this. A detailed explanation of how our test setup works should go into the wiki.

tests/setupTest.ts
Slartibartfass2: I added some.

tests/setupTest.ts
Slartibartfass2: Yeah, no fan of the generic `id` argument.

tests/setupTest.ts
Slartibartfass2: Right, I changed it.

tests/setupTest.ts
Slartibartfass2: `ProjectPaper` doesn't store the project ID, which makes that impossible. I applied the filter for `PaperDecision.UNDECIDED`.

tests/setupTest.ts
Slartibartfass2: Changed it.

tests/setupTest.ts
Slartibartfass2: I changed it a bit, but as mentioned above, a specific mock should be taken on the component test level.

Slartibartfass2 (commented):
Thanks for the review.

tests/setupTest.ts
Merseleo: I like.

tests/setupTest.ts
Merseleo: I was not very familiar with mocking in vitest, but I read the documentation and now I agree with you. It seems to be better to locally mock the calls and the functionality repeats to add a global default mock. So you ignore this comment.

tests/setupTest.ts
Merseleo: Great, thank you. Because I think this is an example where it makes sense to add a general result as this could always be the same and we save the effort of implementing it every time

tests/setupTest.ts
Merseleo: Right, but the filtering for undecided papers is more important anyway, thank you.

tests/setupTest.ts
Merseleo: You are right, but sometimes I think you want to extend on a good default implementation, don't you? 
In my opinion, it is therefore a good idea to at least have sensible and good default results for these Get requests that you can build on. But I agree with you, we can also simply expand and change this later if we see exactly what we need in the application.

But as we don't have any deleted projects, I think you should not return a list containing an archived project, when requesting deleted projects?!

Merseleo (changes requested):
I think we are almost finished.

tests/setupTest.ts
Slartibartfass2: As discussed, for the deleted projects, we return an empty list for now.

Merseleo (approved):

PR #171, opened 2025-01-20
Added ability to invite users

Slartibartfass2:
Closes #16 
Closes #33 

## What I have made

- I know this might look like a big PR again :/, but I swear it isn't. Many changes are due to refactoring.
- I created a reusable `SettingsLayout`, which can be used for the project and user settings
- for the project settings, I created a reusable `ProjectSettingsLayout`
- We often use some sort of `Tab` type. Previously, all these tab types were spread across the codebase, now I declared them in `tabs.ts` so that they can be used everywhere
- I added a reusable `Dialog` component, which is now used in `CreateProjectDialog` and `InviteUsersDialog`
- Because `CreateProjectDialog` and `InviteUsersDialog` share the same invite users input element, I extracted it to a separate component `InviteUsersInput`, which uses the `ChipsInput` and manages the user<->chip logic, which was previously in `CreateProjectDialog`
- Lastly, I added the members list of the members settings page
  - the promotion and removal button don't work yet, we have other PRs for that
- To test components that use API calls inside, I added a `mockApiCall` and `mockFailedApiCall`, which enable mocking single API calls

Use the following files to test it without the mock backend:
[mock-data.zip](https://github.com/user-attachments/files/18913052/mock-data.zip)
- put `+layout.ts` in `src/routes/project/[projectId]`
- put `+page.ts` in `src/routes/project/[projectId]/settings/members`

### Note
- currently, when running the tests, you'll see a lot of `invalid_ownership_binding` error messages
- this is because we bind `membersInput` in `CreateProjectDialog` and `InviteUsersDialog` to `InviteUsersInput`
- when it isn't wrapped in a dialog, the error doesn't appear
- I spent hours in figuring this out and trying to fix this, but didn't succeed.

## Checklist

Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does not apply.

- [x] I have updated the documentation accordingly and commented my code
- [x] I have added tests that prove my fix is effective or that my feature works
  - ~~[ ] unit tests~~ (no logic added that needs unit tests)
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (not possible yet)
- [ ] (for reviewer) I have checked the implementation against the requirements

PR #218, opened 2025-02-18
Use `#each` blocks with keys and add svelte-eslint-plugin rules

### LINKED ISSUE #202: Use keyed each blocks
Currently, we don't use keys for svelte each blocks. According to the [docs](https://svelte.dev/docs/svelte/each#Keyed-each-blocks), this is good because svelte can then diff the elements and reload the items correctly.
There's an ESLint rule for that: https://sveltejs.github.io/eslint-plugin-svelte/rules/require-each-key/
We should investigate why the rule doesn't apply and enforce this behavior.

Slartibartfass2:
Closes #202

## What I have made

- I recommend reviewing this commit-by-commit to see which changes belong to which ESLint rule
- in #171  I noticed a problem with the ESLint plugin, I described in https://github.com/SE-UUlm/snowballr-frontend/commit/13f60ca6408ae9dd51a7208335d63c7707291c88
- I fixed it
- Also, I added some rules to improve further code quality, tell me your opinion
- I am considering adding the [`sort-attributes`](https://sveltejs.github.io/eslint-plugin-svelte/rules/sort-attributes/) rule
  - It could be a lot, but also it could improve readability and I find myself often thinking about how the order should be
  - What's your opinion on that?

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no logic or similar was added, only linting stuff)

### DIFF
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -32,7 +32,8 @@
     }: NamedListProps = $props();
 </script>
 
-<!--@component
+<!--
+@component
 Named list with a header and custom list items of type `YourComponent`.
 
 This list component can be used to uniformly format named lists.
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -97,7 +97,7 @@ This can be e.g. a search bar.
         {console.error(`Could not load items: ${error}`)}
         <h2>{listName}</h2>
         <div class="flex flex-row items-center gap-x-2 p-4">
-            <CircleAlert size={20} class="text-neutral-500" />
+            <CircleAlert class="text-neutral-500" size={20} />
             <span class="text-error">{errorHint ? errorHint : error}</span>
         </div>
     {/await}
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -14,6 +14,7 @@
         emptyHint?: string;
         errorHint?: string;
         preListContent?: Snippet;
+        keySelector: (item: T) => string | number;
     }
 
     const {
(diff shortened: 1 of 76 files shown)


src/lib/components/composites/list/NamedList.svelte
```diff
@@ -14,6 +14,7 @@
         emptyHint?: string;
         errorHint?: string;
         preListContent?: Snippet;
+        keySelector: (item: T) => string | number;
```
Merseleo, line 21: Maybe explain the usage of this function in the component comment, so that it should map the items to a unique id.

Merseleo (changes requested):
I think can add the "sort attributes" rule.

What do you think of the following rules: 
- [svelte/no-ignored-unsubscribe](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-ignored-unsubscribe/)
- [svelte/no-unused-class-name](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-unused-class-name/)


And is there any argument for splitting the lint and lint:fix commands instead of automatically fixing?

src/lib/components/composites/list/NamedList.svelte
Slartibartfass2: I added it as a comment directly for the property.

Slartibartfass2:
> I think can add the "sort attributes" rule.
> 
> What do you think of the following rules:
> 
>     * [svelte/no-ignored-unsubscribe](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-ignored-unsubscribe/)
> 
>     * [svelte/no-unused-class-name](https://sveltejs.github.io/eslint-plugin-svelte/rules/no-unused-class-name/)
> 
> 
> And is there any argument for splitting the lint and lint:fix commands instead of automatically fixing?

I added `svelte/no-ignored-unsubscribe` but we cannot use `svelte/no-unused-class-name` as it doesn't recognize the tailwindcss classes.

I think, when using `lint` we just want to receive feedback whether the code complies to our standards. That's important for the CI, where we don't want to automatically fix it.

Merseleo (approved):
You are right, I didn't consider the possible problems caused a fix in the CI. So we are good to go ...

PR #220, opened 2025-02-21
gRPC authentication

### LINKED ISSUE #134: [Feature]: Complete sign up logic

#2

Currently, the sign-up endpoint doesn't work properly and needs to be fixed. Until this is fixed, we cannot implement the correct sign-up behavior.

Wait for https://github.com/SE-UUlm/snowballr-backend/issues/267 and then complete the sign-up logic.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #134

## What I have made

Http-Only Authentication with the gRPC backend api.
See SE-UUlm/snowballr-api#14, SE-UUlm/snowballr-mock-backend#12

To test this out:
- Use this mock-backend branch: SE-UUlm/snowballr-mock-backend#12
- Use this branch for the frontend
- Open the frontend in the browser
- You should be navigated to `/signin` automatically.
- Navigate to `/signup` (as the `signin` is currently not implemented)
- Register as usual
- You should be redirected to `/` automatically, but now logged in.

E2E tests are updated to spin up mock backends for each test/worker, see the two new fixtures for documentation.

To validate E2E tests, the used mock backend version has to be updated to (PR not merged yet):
- Use this mock-backend branch: SE-UUlm/snowballr-mock-backend#12
- Build the docker container: `docker build . -t mock-backend:latest`
- Checkout this PR in the frontend
- Execute this in the root folder: `sed -i "s/ghcr.io\/se-uulm\/snowballr-mock-backend:main/mock-backend:latest/" tests/e2e/mock-backend.ts` or manually change the image name in `mock-backend.ts`
- Run the e2e tests

## Checklist

  - [ ] unit tests
  - [ ] integration tests
  - [x] end-to-end tests


src/lib/grpc-api.ts
```diff
@@ -9,8 +9,23 @@ if (!env.PUBLIC_API_BASE_URL) {
...
+    fetch(input, init);
+
+export function setFetch(newFetch: typeof fetch) {
+    fetch = newFetch;
+}
+
 const transport = new GrpcWebFetchTransport({
     baseUrl: env.PUBLIC_API_BASE_URL,
+    fetchInit: {
+        credentials: "include", // same-origin?
```
mowi12: Since frontend and backend do not need to run on the same domain, `// same-origin?` can be removed, or not?
Or do you want to keep this comment, in order to remember, that this can be set?
If yes, then maybe use a more descriptive comment.

src/routes/(auth)/signup/+page.svelte
```diff
@@ -34,12 +35,11 @@
             password: passwordInput.getValue(),
         };
 
-        backendService
+        await backendService
             .register(userData)
             .then(() => {
-                // TODO: Login with accessToken, store both tokens and redirect to the home page,
-                // will be completed in !220
                 console.log("New user was registered.");
```
mowi12: I think we can remove this `console.log()` since it is not providing any useful information. If it was successful, we will see it, because we will be redirected to the homepage.
And, if I remember correctly, no other `then()` call prints, that it was successful.

src/lib/grpc-api.ts
```diff
@@ -9,8 +9,23 @@ if (!env.PUBLIC_API_BASE_URL) {
     process.exit(1);
 }
 
+let fetch = globalThis.fetch;
+const customFetch = (input: URL | RequestInfo, init?: RequestInit | undefined): Promise<Response> =>
+    fetch(input, init);
+
+export function setFetch(newFetch: typeof fetch) {
+    fetch = newFetch;
+}
```
mowi12, line 21: What exactly is this custom fetch needed, can you explain?

src/routes/+layout.ts
```diff
@@ -1,27 +1,45 @@
...
+    const authStatus = (await backendService.getAuthenticationStatus(Nothing).response)
+        .authenticationStatus;
+
+    if (authStatus === AuthenticationStatus.ACCESS_TOKEN_EXPIRED) {
+        await backendService.renewSession({});
+    } else if (authStatus !== AuthenticationStatus.AUTHENTICATED && !onUncheckedPath) {
+        redirect(307, "/signin");
+    }
+
     /// TODO: exchange this logic in !220 or !124 with better handling of the case, the user could not be loaded
```
mowi12: Did you implement this comment or not?
If yes, then please remove it, if not, remove !220 from the comment as it then needs to be implemented in !124

src/routes/+layout.ts
```diff
@@ -1,27 +1,45 @@
...
+
+    const uncheckedPaths = ["/signin", "/signup", "/resetpassword"];
+    const onUncheckedPath = uncheckedPaths.includes(url.pathname);
+    if (onUncheckedPath) {
+        return {
+            user: undefinedUser,
+        };
+    }
+
+    const authStatus = (await backendService.getAuthenticationStatus(Nothing).response)
```
mowi12: What about the the case if this is an error?
Maybe use a try catch around the whole auth status check just in case.

wiki/Testing.md
```diff
@@ -134,27 +134,14 @@ end-to-end tests and getting feedback on them more quickly.
...
-| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT`   | 3004                                 | The port that the mock backend will listen on for E2E testing of the webkit browser.   |
+The E2E tests will automatically create these mock backends using docker,
+requiring it to be installed and running.
 
 ### Best Practices
 
 1. **Use the test fixture**:
    Instead of importing the `test` fixture directly from "@playwright/test" directly,
-   use the `test` variable exported from [`fixtures/general-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/tests/e2e/fixtures/general-fixture.ts).
+   use the `test` variable exported from [`fixtures/shared-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/tests/e2e/fixtures/shared-fixture.ts).
```
mowi12, line 144: We might have an issue here, since the new path is not available on the develop branch ...

tests/e2e/fixtures/isolated-fixture.ts
```diff
@@ -0,0 +1,67 @@
...
+    user: User | null;
+    apiClient: AuthSnowballRClient;
+}>({
+    user: [alice, { scope: "test", option: true }],
+    mockBackend: [
+        // eslint-disable-next-line no-empty-pattern
+        async ({}, use) => {
+            const mockBackend = await DockerMockBackend.create();
+            await use(mockBackend);
+            mockBackend.dispose();
```
mowi12, line 50: shouldn't we also `await` for the `dispose()` to finish?
Same in the shared fixture of course.

tests/e2e/fixtures/isolated-fixture.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export const test = base.extend<{
+    mockBackend: DockerMockBackend;
+    user: User | null;
+    apiClient: AuthSnowballRClient;
+}>({
+    user: [alice, { scope: "test", option: true }],
+    mockBackend: [
+        // eslint-disable-next-line no-empty-pattern
+        async ({}, use) => {
```
mowi12, line 47: Why don't you use an underscore?
This allows you to remove the eslint-disable comment.

Same in the shared-fixture of course.

```suggestion
        async (_, use) => {
```

mowi12 (changes requested):
Nicely done. Really good code quality. There are only some questions we need to talk about.

tests/e2e/fixtures/isolated-fixture.ts
luca-schlecker: Because playwright won't compile when using `_`. It errored out with a message saying it requires `{ ... }` like syntax. I originally started with `_`.

tests/e2e/fixtures/isolated-fixture.ts
luca-schlecker: `dispose` is not `async`. So no, we shouldn't use `await`.

wiki/Testing.md
luca-schlecker: I know, but it will be... Don't know what I should do about this.

src/lib/grpc-api.ts
luca-schlecker: I've introduced an environment variable which can be used to set this value. I think this is the best approach as (depending on the environment) this value may change.

src/lib/grpc-api.ts
luca-schlecker: SvelteKit provides its own fetch which should be used. This code defaults to `globalThis.fetch` until the fetch is overwritten later on (as it is not instantly available afaik). The indirection is needed because otherwise the modification of `fetch` would not impact the variable passed to the transport.

image

src/routes/+layout.ts
luca-schlecker: I implemented better error handling and an alert dialog.

src/lib/grpc-api.ts
mowi12: Okay, makes sense!

src/routes/+layout.ts
mowi12: I will open a new conversation to reference the new code.

wiki/Testing.md
mowi12: I would suggest to remove the link and insert a comment with TODO: will be done in: and create a new Issue with PR in which this string will be inserted after we merged into dev, what do you think?

tests/e2e/fixtures/isolated-fixture.ts
mowi12: You're right. I should have looked at the method...

tests/e2e/fixtures/isolated-fixture.ts
mowi12: Okay, good to know.

src/lib/grpc-api.ts
```diff
@@ -9,8 +9,28 @@ if (!env.PUBLIC_API_BASE_URL) {
     process.exit(1);
 }
 
+let credentials: RequestCredentials = "same-origin";
+if (env.PUBLIC_CREDENTIAL_MODE) {
+    credentials = (env.PUBLIC_CREDENTIAL_MODE as RequestCredentials) ?? "same-origin";
+}
```
mowi12:
```suggestion
let credentials: RequestCredentials = env.PUBLIC_CREDENTIAL_MODE
    ? (env.PUBLIC_CREDENTIAL_MODE as RequestCredentials)
    : "same-origin";
}
```

src/routes/+layout.svelte
```diff
@@ -17,6 +19,21 @@
...
+                An Error Occurred!
+            </AlertDialog.Title>
+            <AlertDialog.Description class="text-default">
+                The backend could not be reached or a severe and unrecoverable error occured during
+                communication.
+            </AlertDialog.Description>
+        </AlertDialog.Header>
+    </AlertDialog.Content>
+</AlertDialog.Root>
+
```
mowi12, line 36: Generally, this is a very good idea to show this AlertDialog, but for me, this dialog always appeared. Even when the mock backend is running and i go the signup page, this pop up. I don't think this should happen.

Screenshot 2025-05-16 at 17 24 41

src/routes/+layout.ts
```diff
@@ -1,27 +1,63 @@
...
+    if (authStatusCall.status.code !== GrpcStatusCode[GrpcStatusCode.OK]) {
+        return {
+            user: undefinedUser,
+        };
+    }
+
+    const authStatus = authStatusCall.response.authenticationStatus;
+
+    if (authStatus === AuthenticationStatus.ACCESS_TOKEN_EXPIRED) {
+        await backendService.renewSession({});
```
mowi12: Above, you used `Nothing`. Please use `Nothing` here too.

tests/e2e/unreachable.test.ts
```diff
@@ -0,0 +1,15 @@
...
+        await expect(page.getByRole("alertdialog", { name: "An Error Occurred!" })).toBeVisible();
+    });
+
+    test("When the backend is reachable, then no alert dialog is displayed", async ({ page }) => {
+        await page.goto("/");
+        await expect(
+            page.getByRole("alertdialog", { name: "An Error Occurred!" }),
+        ).not.toBeVisible();
+    });
+});
```
mowi12: As already described, locally, this should not work because the AlertDialog is alway popping up. How do these tests even work?

Isn't the backend running in both tests?

mowi12 (changes requested):
You fixed a lot, but now there are other things that need to get fixed.

And one last question. Is it possible to stop/kill the e2e tests if docker is not running?

wiki/Testing.md
luca-schlecker: I don't think the extra issue and pr is worth it, but if you prefer that, we can go that route.

src/lib/grpc-api.ts
luca-schlecker: This doesn't cover `env.PUBLIC_CREDENTIAL_MODE` having an invalid value, but I've got something better.

src/routes/+layout.svelte
luca-schlecker: It was an error on my part, I didn't distinguish between an expected valid route without a user and an unexpected error. Implemented this using `null` (present but null) and `undefined` (not present).

tests/e2e/unreachable.test.ts
luca-schlecker: Quite a sneaky bug there. The backend was running in none of the tests (see the used fixture). The latter one just succeeds because the page is currently loading, and thus the dialog is not yet visible, instantly returning a positive assertion.

luca-schlecker:
Tests now immediately exit if docker isn't up and running.

wiki/Testing.md
mowi12: It seems like, in the last push, there is no error any more. Any idea why?

src/lib/grpc-api.ts
mowi12: Looks way better!

mowi12 (changes requested):
So, almost done!
You only need to fix the e2e tests, as they do not work locally and in the pipeline.

Screenshot 2025-05-16 at 19 33 01

Screenshot 2025-05-16 at 19 33 35

Its allways the same error for all the tests (at least locally)

wiki/Testing.md
mowi12: We skipped it, as discussed : )

mowi12 (dismissed):
We finally got it working, hopefully : D

tests/e2e/unreachable.test.ts
mowi12, line 1: Can you explain why we use the basic test import for the first test and the shared fixture for the second test?

tests/e2e/mock-backend.ts
```diff
@@ -0,0 +1,202 @@
...
+     * Create a new `DockerMockBackend` listening on a random port and a new
+     * playwright driver page targeting the newly created mock backend instance.
+     * This automatically sets up routing of api calls.
+     *
+     * DO NOT FORGET to call `dispose` after you are finished using this instance
+     * to stop the container and free up resources.
+     */
+    static async createWithPage(
+        browser: Browser,
+        baseURL: string = "http://localhost:4173",
```
mowi12, line 193: Why do we use port 4173 here, but in the playwright config we specify port 3001?

mowi12 (changes requested):
Nice! It looks like everything is working now.
Can you explain what you changed? Based on the code I see no clear difference why the pipeline did fail before.

And there are two questions which my need changes, but based on your anwers may not.

tests/e2e/mock-backend.ts
luca-schlecker: This should be the Frontend.

luca-schlecker: > Why do we use port 4173 here, but in the playwright config we specify port 3001?

The playwright config only has to define `PUBLIC_API_BASE_URL` because otherwise all tests will crash (like we observed before). It is not needed for backend communication though, as the backend reroutes the traffic anyway.

tests/e2e/unreachable.test.ts
luca-schlecker: The first one tests if there is an error if no backend is reachable, it thus uses the basic playwright fixture which does not provide a backend.

The second one tests if everything works alright if there is a backend, that's why one of the backend fixtures is used, as these provide and automatically setup the backend connection.

luca-schlecker:
> Nice! It looks like everything is working now.
> Can you explain what you changed? Based on the code I see no clear difference why the pipeline did fail before.
> 
> And there are two questions which my need changes, but based on your anwers may not.

I defined `PUBLIC_API_BASE_URL` even though it is not used in the E2E tests, because otherwise the Frontend malfunction.

And then I increased the timeout for the mock backend fixture only, as it may need extra time pulling the image and starting the container on slow machines.

tests/e2e/unreachable.test.ts
mowi12: Okay, makes sense!

mowi12 (approved):

PR #233, opened 2025-03-03
Added prettier tailwind plugin

### LINKED ISSUE #232: [Feature]: Add prettier tailwind plugin

_No response_

Currently, we don't use the prettier tailwind plugin, although we use prettier and tailwind. The plugin sorts the tailwind classes according to their order.

There's an explanation [here](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted).

Add the `prettier-plugin-tailwindcss` and configure it to work with shadcn as well.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #232 

## What I have made

- see issue description

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no code changes)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -5,7 +5,7 @@
     "singleQuote": false,
     "trailingComma": "all",
     "printWidth": 100,
-    "plugins": ["prettier-plugin-svelte"],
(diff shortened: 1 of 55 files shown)


Merseleo (approved):
Great idea!

PR #250, opened 2025-03-11
Setup E2E tests with example test

### LINKED ISSUE #127: [Feature]: E2E CI using Mock Backend

_No response_

E2E testing is currently not possible.

GitHub CI Pipeline executing E2E tests against the mock backend.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #241: [Feature]: Setup E2E testing

#127

At the moment it is not really possible to add E2E tests to the frontend. As the mock backend now exists and can be loaded with data, it is possible to create E2E tests

Check the current test setup and extend it, so E2E tests can be added and executed and add a demo E2E test suite for the project creation process.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #241

## What I have made

- I extended the playwright configuration, so the E2E tests are not only executed in one browser but with all three commonly used browser (engines) (please comment in the review, whether what your opinion is on adding more browsers, e.g. [mobile browsers](https://playwright.dev/docs/test-projects#configure-projects-for-multiple-browsers)
  - this means that each test is executed three times, one for each browser and with a separate mock backend instance
- I added example test cases for testing the process of creating a new project (as far as it is possible at the moment as there are functions and frontend code missing, especially regarding the invitations of additional project member which will be added with #16)
  - for these tests I experimented with the so called [Page Object Model](https://playwright.dev/docs/pom) and started one model for the homepage (please let me know what you think of it and whether we should continue using this as it prevents duplicate code, but decrease the test readability imo)
  - I fixed some "bugs" during the project creation process or on the homepage of the created project
- I updated the the CI, so the E2E tests are executed in the pipeline (Closes #127)
- I added a script to start a single or multiple mock backends

## Checklist

  - ~~[ ] integration tests~~ (not necessary)
  - [x] end-to-end tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,2 @@
 ; Prefix client code env variables with 'PUBLIC_' (https://svelte.dev/tutorial/kit/env-static-public)
 PUBLIC_API_BASE_URL=http://localhost:8080
-
-PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_CHROMIUM=3002
-PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_FIREFOX=3003
-PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT=3004
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,6 @@
 ; Prefix client code env variables with 'PUBLIC_' (https://svelte.dev/tutorial/kit/env-static-public)
 PUBLIC_API_BASE_URL=http://localhost:8080
+
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_CHROMIUM=3002
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_FIREFOX=3003
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT=3004
--- a/.github/workflows/code_quality_checks.yml
+++ b/.github/workflows/code_quality_checks.yml
@@ -100,13 +100,6 @@ jobs:
               with:
                   node_version: ${{ env.node_version }}
 
-            - name: Login to GitHub Container Registry
-              uses: docker/login-action@v3
-              with:
-                  registry: ghcr.io
-                  username: ${{ github.actor }}
-                  password: ${{ secrets.GITHUB_TOKEN }}
-
             - name: Install Playwright
               run: npx playwright install --with-deps
 
--- a/.github/workflows/code_quality_checks.yml
+++ b/.github/workflows/code_quality_checks.yml
@@ -90,6 +90,9 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v4
               with:
+                  # Force a non-shallow checkout, so that we can reference $GITHUB_BASE_REF.
+                  # See https://github.com/actions/checkout for more details.
+                  fetch-depth: 0
                   submodules: "recursive"
 
             - name: Setup
@@ -97,8 +100,19 @@ jobs:
               with:
                   node_version: ${{ env.node_version }}
 
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v3
+              with:
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
             - name: Install Playwright
               run: npx playwright install --with-deps
+
+            - name: Run only tests for changed code (fail fast)
+              if: github.event_name == 'pull_request'
+              run: npm run test:e2e -- --only-changed=origin/${{ github.base_ref }}
 
             - name: Run Playwright
               run: npm run test:e2e
--- a/README.md
+++ b/README.md
@@ -101,11 +101,11 @@ npm run test:e2e
 
 Additionally, you can configure the E2E tests by setting the following process environment variables:
 
-| Variable                                     | Default (if not variable not) | Description                                                                            |
-| -------------------------------------------- | :---------------------------: | -------------------------------------------------------------------------------------- |
-| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_CHROMIUM` |             3002              | The port that the mock backend will listen on for E2E testing of the chromium browser. |
-| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_FIREFOX`  |             3003              | The port that the mock backend will listen on for E2E testing of the firefox browser.  |
-| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT`   |             3004              | The port that the mock backend will listen on for E2E testing of the webkit browser.   |
+| Variable                                     | Default (if the variable not set) | Description                                                                            |
+| -------------------------------------------- | :-------------------------------: | -------------------------------------------------------------------------------------- |
+| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_CHROMIUM` |               3002                | The port that the mock backend will listen on for E2E testing of the chromium browser. |
+| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_FIREFOX`  |               3003                | The port that the mock backend will listen on for E2E testing of the firefox browser.  |
+| `PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT`   |               3004                | The port that the mock backend will listen on for E2E testing of the webkit browser.   |
 
 ### Lighthouse
 
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ npm run compile:proto
 
 To run the app, you need to create a `.env` file in the root directory of the project. The file should contain the following environment variables (see [.env.example](./.env.example) for an example):
 
-| Variable              |      Required      | Default | Description                                                                            |
-| --------------------- | :----------------: | :-----: | -------------------------------------------------------------------------------------- |
-| `PUBLIC_API_BASE_URL` | :white_check_mark: |    -    | The URL of the backend API.                                                            |
-| `PUBLIC_IS_DEV_MODE`  |        :x:         |  false  | Whether the app is in development mode. This may enable additional development tooling |
+| Variable              |      Required      | Default | Description                                                                             |
+| --------------------- | :----------------: | :-----: | --------------------------------------------------------------------------------------- |
+| `PUBLIC_API_BASE_URL` | :white_check_mark: |    -    | The URL of the backend API.                                                             |
+| `PUBLIC_IS_DEV_MODE`  |        :x:         |  false  | Whether the app is in development mode. This may enable additional development tooling. |
 
 After completing these steps, you can start a development server with:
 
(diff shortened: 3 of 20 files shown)


Merseleo:
In my opinion we should document some of the stuff I developed in this PR, e.g. that a script for launching mock backends exist and how it can be used or how to add new E2E tests or browsers. But I am not sure, whether we should add this only to the wiki or I should update the README, please let me know, what you think.

.github/workflows/code_quality_checks.yml
```diff
@@ -90,16 +90,30 @@ jobs:
...
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
             - name: Install Playwright
               run: npx playwright install --with-deps
 
+            - name: Run only tests for changed code (fail fast)
+              if: github.event_name == 'pull_request'
+              run: npm run test:e2e -- --only-changed=origin/${{ github.base_ref }}
```
Slartibartfass2, line 108: Are we sure to only test the changed code?
What if existing code breaks due to a dependency?

playwright.config.ts
```diff
@@ -12,11 +13,73 @@ export default defineConfig({
     ],
     use: {
         baseURL: "http://localhost:4173",
-        screenshot: "only-on-failure",
+        screenshot: "on",
```
Slartibartfass2, line 16: Is there a reason for creating a screenshot every time?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -193,6 +197,7 @@
                 "h-fit w-full py-3 text-xl [&_svg]:size-5",
             )}
             data-testid="dialog-trigger-button"
+            disabled={loadingPossibleMembers}
```
Slartibartfass2, line 199: Shouldn't we bind the state here?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -37,14 +37,18 @@
...
             const allUsers = await backendService.getAllUsers(Nothing).response;
 
             initialPossibleMembers = allUsers.users.filter((user) => user.id !== currentUser.id);
+            loadingPossibleMembers = false;
         } catch (error) {
             isErrorOnUsersLoading = true;
+            loadingPossibleMembers = false;
             console.error(`Couldn't get users from server (${error})`);
         }
     });
```
Slartibartfass2: Just put it after the try-catch block
```suggestion
        } catch (error) {
            isErrorOnUsersLoading = true;
            console.error(`Couldn't get users from server (${error})`);
        }
        loadingPossibleMembers = false;
    });
```

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -268,6 +273,9 @@
                     projectWasCreated = false;
                     projectId = undefined;
                     membersInput = [];
+
+                    // trigger reload of the homepage, so the created project is shown in the projects list
+                    window.location.reload();
```
Slartibartfass2: Are we sure that this is the correct way?
Shouldn't we just update the list of projects via a callback?
Reloading everything seems a lot to me.

Slartibartfass2 (changes requested):
Great work Leonhard!
Note this isn't a full review, I just skimmed through the changes out of interest and that's what I noticed.

playwright.config.ts
Merseleo: This was mainly because I was sure whether some test cases in the CI (because of the headless browser and strange "error messages") were running correctly and wanted to see the results. But it generally turned out that it is practical to look at the screenshots afterwards. But of course I can also undo it, because I also understand that many screenshots are unnecessary.

.github/workflows/code_quality_checks.yml
Merseleo: I first test the changed code / tests to detect errors as fast as possible and only if the changed code / tests passes, I run the entire E2E test suite. But if we introduced a faulty test or code in general, we get quicker feedback by "prioritizing" these.

Merseleo: Moreover, this behavior is suggested from playwright, it was not my idea ^^.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Good idea, I changed it.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Afaik the `disabled` property is not bindable.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: This is one way to solve the problem, but I agree it isn't the best way. I try to figure out, how we can improve this.

Merseleo: I found a solution with only refetching the page data, but I am not sure, whether the behavior is not as you imaged? Otherwise I can try something else.

Merseleo (commented):
Thank you for the first ideas.

.github/workflows/code_quality_checks.yml
Slartibartfass2: As discussed, I didn't see that we also test everything below

.env.example
```diff
@@ -1,2 +1,6 @@
 ; Prefix client code env variables with 'PUBLIC_' (https://svelte.dev/tutorial/kit/env-static-public)
 PUBLIC_API_BASE_URL=http://localhost:8080
+
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_CHROMIUM=3002
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_FIREFOX=3003
+PUBLIC_MOCK_BACKEND_GRPC_WEB_PORT_WEBKIT=3004
```
Slartibartfass2: As discussed, let's remove these.
```suggestion
```

.github/workflows/code_quality_checks.yml
```diff
@@ -90,16 +90,30 @@ jobs:
...
               with:
                   node_version: ${{ env.node_version }}
 
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v3
+              with:
+                  registry: ghcr.io
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
```
Slartibartfass2: This is not necessary anymore, now that they're public.
```suggestion
```

package.json
```diff
@@ -18,7 +18,7 @@
         "test:unit:ui": "npm run test:unit -- --ui --watch --open",
         "test:integration": "vitest --project integration",
         "test:integration:ui": "npm run test:integration -- --ui --watch --open",
-        "test:e2e": "playwright test",
+        "test:e2e": "bash scripts/start_mock_backend.sh -c 3002 3003 3004 && playwright test",
         "test:no-e2e": "vitest",
```
Slartibartfass2: What about a fast test like in the CI
```suggestion
        "test:e2e": "bash scripts/start_mock_backend.sh -c 3002 3003 3004 && playwright test",
        "test:e2e:fast": "bash scripts/start_mock_backend.sh -c 3002 3003 3004 && playwright test --only-changed=origin/develop",
        "test:no-e2e": "vitest",
```

tests/e2e/project-creation.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        await expect(page.getByRole("link", { name: "Demo project 3" })).not.toBeVisible();
+
+        await expect(page.getByText("stage 0")).toBeVisible();
+        await expect(page.getByText("reviewed 0 / 0")).toBeVisible();
+        await expect(page.getByText("No open reviews")).toBeVisible();
+        await expect(page.getByText("estimated remaining time")).not.toBeVisible();
+    });
+
+    /* TODO: add E2E tests here for checking, that the user creating this project is project admin, the other members
+       were invited and the settings are the same as the default settings for new project from the user */
```
Slartibartfass2, line 97: What is required to test these properties?

tests/integration/statistics/project-information.test.ts
```diff
@@ -36,4 +36,33 @@ describe("ProjectInformation", () => {
             "Couldn't load project information.",
         );
     });
+
+    test("When all the project was just created, then the project information component does not show information about the estimated remaining time.", async () => {
```
Slartibartfass2: typo
```suggestion
    test("When the project was just created, then the project information component does not show information about the estimated remaining time.", async () => {
```

tests/e2e/fixtures/general-fixture.ts
```diff
@@ -0,0 +1,21 @@
+import { test as base } from "@playwright/test";
+
+export type TestOptions = {
+    mockBackendUrl: string;
+};
+
+export const test = base.extend<{ forEachTest: void } & TestOptions>({
```
Slartibartfass2, line 22: I had a hard time understanding what this snippet does. Please add a bit of documentation as it seems to be important for the all e2e tests.

Slartibartfass2 (changes requested):
Great work! Only some minor things left.

.env.example
Merseleo: I deleted it.

.github/workflows/code_quality_checks.yml
Merseleo: You are right, I think need to adapt the pipeline in the mock backend repository as well.

tests/e2e/project-creation.test.ts
Merseleo: Your implementation on the invite user settings page is necessary #16, as I can't navigate to this page and see als invited users at the moment.

tests/integration/statistics/project-information.test.ts
Merseleo: I will fix this.

tests/e2e/fixtures/general-fixture.ts
Merseleo: Yes, I will add an explanation, but probably we need to elaborate on the entire testing process in the wiki / general documentation.

Merseleo (commented):
Thank you for the review, I added your suggestions and improved the documentation. 
I suggest that we discuss the rest of the documentation tomorrow.

README.md
```diff
@@ -99,6 +99,14 @@ End-to-end tests are used to test the functionality of the app as a whole. They
 npm run test:e2e
 ```
 
+Additionally, you can configure the E2E tests by setting the following process environment variables:
+
+| Variable                                     | Default (if not variable not) | Description                                                                            |
```
Slartibartfass2: "(if not variable not)" doesn't make sense. What about just removing the parentheses?

Slartibartfass2 (changes requested):
Nearly done. Can you also add the section to the wiki?

README.md
Merseleo: Yeah, I agree, I dont remember what I wanted to say, when I wrote this comment.

Merseleo:
> Nearly done. Can you also add the section to the wiki?

Yes I can add the section as soon as we agree how we structure it.

README.md
```diff
@@ -99,6 +99,14 @@ End-to-end tests are used to test the functionality of the app as a whole. They
 npm run test:e2e
 ```
 
+Additionally, you can configure the E2E tests by setting the following process environment variables:
+
+| Variable                                     | Default (if the variable not set) | Description                                                                            |
+| -------------------------------------------- | :-------------------------------: | -------------------------------------------------------------------------------------- |
```
Slartibartfass2:
```suggestion
| Variable                                     | Default (if the variable is not set) | Description                                                                            |
```

Slartibartfass2 (changes requested):
Let's make it at least an english sentence.

Slartibartfass2 (dismissed):

Slartibartfass2 (approved):

PR #251, opened 2025-03-11
Avatars of reviewers are shown in paper view

### LINKED ISSUE #242: [Bug]: Avatars of reviewers are empty on paper view

#45

### What happened?

When looking at the review decisions on the paper view, the avatars of the reviewers are empty. No initials are shown.

This has the reason that in #195 this was overseen.

The fix is to change the following lines in `src/routes/project/[projectId]/paper/[paperId]/+page.ts`:
from
```ts
    const reviewers: Promise<User[]> = criteriaWithReviews.then(async (criteria) => {
        const users: User[] = [];
        const reviews = criteria.flatMap((criterion) => criterion.reviews);
        for (const review of reviews) {
            if (!users.some((user) => user.id === review.userId)) {
                await backendService.getUserById({ id: review.userId }).response;
            }
        }
        return users;
    });
```
to
```ts
    const reviewers: Promise<User[]> = criteriaWithReviews.then(async (criteria) => {
        const users: User[] = [];
        const reviews = criteria.flatMap((criterion) => criterion.reviews);
        for (const review of reviews) {
            if (!users.some((user) => user.id === review.userId)) {
                const user = await backendService.getUserById({ id: review.userId }).response;
                users.push(user);
            }
        }
        return users;
    });
```

### Steps to reproduce

1. Go on paper view
2. Navigate to review information
3. the avatars of the reviewers are empty

### What did you expect?

The initials should be shown.

### Anything else?

_No response_

Slartibartfass2:
Closes #242 

## What I have made

- See issue

## Checklist

- ~[ ] I have added tests that prove my fix is effective or that my feature works~ (only e2e tests would've made sense, but they do not exist yet)

### DIFF
--- a/src/routes/project/[projectId]/paper/[paperId]/+page.ts
+++ b/src/routes/project/[projectId]/paper/[paperId]/+page.ts
@@ -34,14 +34,13 @@ export const load: PageLoad = ({ params }) => {
     criteriaWithReviews.catch(() => {});
 
     const reviewers: Promise<User[]> = criteriaWithReviews.then(async (criteria) => {
-        const users: User[] = [];
         const reviews = criteria.flatMap((criterion) => criterion.reviews);
-        for (const review of reviews) {
-            if (!users.some((user) => user.id === review.userId)) {
-                const user = await backendService.getUserById({ id: review.userId }).response;
-                users.push(user);
-            }
-        }
+        const userIds: Set<string> = new Set(reviews.map((review) => review.userId));
+        const users: User[] = await Promise.all(
+            Array.from(userIds)
+                .map((id) => backendService.getUserById({ id }).response.catch(() => undefined))
+                .filter((user) => user !== undefined) as Promise<User>[],
+        );
         return users;
     });
 
--- a/src/routes/project/[projectId]/paper/[paperId]/+page.ts
+++ b/src/routes/project/[projectId]/paper/[paperId]/+page.ts
@@ -38,7 +38,8 @@ export const load: PageLoad = ({ params }) => {
         const reviews = criteria.flatMap((criterion) => criterion.reviews);
         for (const review of reviews) {
             if (!users.some((user) => user.id === review.userId)) {
-                await backendService.getUserById({ id: review.userId }).response;
+                const user = await backendService.getUserById({ id: review.userId }).response;
+                users.push(user);
             }
         }
         return users;


src/routes/project/[projectId]/paper/[paperId]/+page.ts
```diff
@@ -38,7 +38,8 @@ export const load: PageLoad = ({ params }) => {
         const reviews = criteria.flatMap((criterion) => criterion.reviews);
         for (const review of reviews) {
             if (!users.some((user) => user.id === review.userId)) {
-                await backendService.getUserById({ id: review.userId }).response;
+                const user = await backendService.getUserById({ id: review.userId }).response;
+                users.push(user);
```
Comment (Merseleo) [inline src/routes/project/[projectId]/paper/[paperId]/+page.ts]: This change is correct, but I think we should add a try catch block in the case, the user id is not found on the error?
However, it does not work every time, as can be seen here: 

Bildschirmfoto 2025-03-11 um 18 18 49

Merseleo (changes requested):
I think we need to search deeper in the code for solving this problem.

src/routes/project/[projectId]/paper/[paperId]/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/paper/[paperId]/+page.ts]: And I think I found the problem and it is not in the frontend but in the mock backend. As this screenshot shows 
Bildschirmfoto 2025-03-11 um 18 23 43
the mock backend generated two reviews, where criteria where selected that do not exist in this project, because everything is highly random generated.

So imo we have two options:
1) Check in the frontend, whether the reviews make sense and really belong to this project (not my preference)
2) Improve the mock backend and only assign reviews to projects, where at least one selected criterion is in the project, so this bug does not occur.
What do you think?

Comment (Slartibartfass2) [reply src/routes/project/[projectId]/paper/[paperId]/+page.ts]: It's not exactly clear, to me, what you mean. I integrated a fallback and improved the asynchronism.

A thing I just noticed: Why is the ID the email?

Comment (Merseleo) [reply src/routes/project/[projectId]/paper/[paperId]/+page.ts]: I saw your fallback and this should work (nice idea with the set, but the solution with catching errors, set it to "undefined" to prevent the rejection and filter all undefined looks quite inelegant, isn't it possible with allSettled as well?).

Did you understood the problem in the first screenshot, so that the names of the reviewers are not loaded in the banner, because the selected criteria do not exist in this project 0 and so your filtering does not work.
So solve this problem we could either check in the frontend, whether the review from the server id valid (so the selected criteria exist in the project) and check this in the mock backend and improve the generation of random reviews to fix this bug.

I dont know, why Luca wanted to use the email as id, but he said that as it is unique anyway it can also be used as id.

Comment (Slartibartfass2) [reply src/routes/project/[projectId]/paper/[paperId]/+page.ts]: I think in this, we should assume that the backend sends us valid data.

When using `allSettled`, we would need to filter for fulfilled promises too. I don't see the advantages, and I even think this solution is shorter. How would've you solved this?

Comment (Merseleo) [reply src/routes/project/[projectId]/paper/[paperId]/+page.ts]: Ok, then we ignore it and eventually fix this in the mock backend.

I would have solved it as you described, but you solution works as well, so lets finish this.

Merseleo (approved):

Issue #39, opened 2024-10-31
[Use Case]: Sort Papers


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to sort papers of a project.

### Acceptance criteria

- sortable properties are
  - paper title (alphabetically)
  - year (numerical)
  - decision (Yes, Maybe, No or No, Maybe, Yes)
  - paper ID (numerical)
- the papers are sorted per stage
- the project member can switch between ascending and descending for each property

### Anything else?

Related to #30

Extends #37

PR #252, opened 2025-03-11
Stages are sorted according to their index

### LINKED ISSUE #248: [Bug]: Stages are not sorted on Papers page

#31

### What happened?

When opening the "Papers" page in a project, it can happen that the stages are not sorted, as can be seen here

Image

When all of the papers for the project are requested they must not be sorted by their stage, but the method `organizePapersByStage` in the `routes/project/[projectId]/papers/page.ts` file assign the papers to their stages without sorting the stages or the papers by their stage. So it can happen that the first paper is in stage 1 and the second paper is in stage 0 and then stage 1 is added to the list of stages before stage 0.

### Steps to reproduce

1. Go to the "Papers" view for different projects
2. Check the stages list

### What did you expect?

The stages to be sorted from stage _0_ to stage _n_

### Anything else?

_No response_

Slartibartfass2:
Closes #248 

## What I have made

- See issue

## Checklist

- ~[ ] I have added tests that prove my fix is effective or that my feature works~ (only e2e tests would've made sense, but they do not exist yet)

### DIFF
--- a/src/routes/project/[projectId]/papers/+page.ts
+++ b/src/routes/project/[projectId]/papers/+page.ts
@@ -1,6 +1,7 @@
 import { backendService } from "$lib/grpc-api";
 import { Project_Paper } from "$lib/model/api/project";
 import { type Stage } from "$lib/model/general";
+import { comparePaperId } from "$lib/utils/common-helper";
 import type { PageLoad } from "./$types";
 import { asPaper } from "$lib/utils/model-helper";
 
@@ -71,5 +72,6 @@ function organizePapersByStage(papers: Project_Paper[]): Stage[] {
 
         stage.papers.push(paper);
     }
-    return stages.toSorted((a, b) => Number(a.stageIndex - b.stageIndex));
+    stages.forEach((stage) => stage.papers.sort((a, b) => comparePaperId(b.id, a.id)));
+    return stages.sort((a, b) => Number(a.stageIndex - b.stageIndex));
 }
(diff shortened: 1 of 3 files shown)


src/routes/project/[projectId]/papers/+page.ts
```diff
@@ -70,5 +71,6 @@ function organizePapersByStage(papers: Project_Paper[]): Stage[] {
 
         stage.papers.push(paper);
     }
-    return stages;
+    stages.forEach((stage) => stage.papers.sort((a, b) => comparePaperId(b.id, a.id)));
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.ts:75]: I appreciate that you sort the papers in the stage. But as there exist an issue for that (see #39), I would appreciate this line even more, you maybe move the sorting of the papers from the load function to the `StageEntry` component and add a hint to this issue. Would that be ok or do you want to keep the sorting at this point the `StageEntry` component should override the order if necessary?

Merseleo (commented):
Only a little question, but I think the bug has been fixed.

src/routes/project/[projectId]/papers/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.ts:75]: I think this is fine for the initial order, other sorting should be made in `StageEntry`. If that's okay for you too.

Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.ts:75]: It is, especially as the sorting will be integrated soon.

Merseleo (approved):
All questions answered ...

PR #254, opened 2025-03-12
Added tsdoc eslint plugin

### LINKED ISSUE #253: [Feature]: Add TSDoc ESLint plugin

_No response_

I propose using the TSDoc ESLint plugin to assure quality throughout code documentation.

Add the TSDoc ESLint plugin and fix current issues.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #253 

## What I have made

- see issue

## Checklist

- ~[ ] I have added tests that prove my fix is effective or that my feature works~ (only added an ESLint plugin)

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -4,6 +4,7 @@ import svelte from "eslint-plugin-svelte";
 import svelteParser from "svelte-eslint-parser";
 import globals from "globals";
 import ts from "typescript-eslint";
+import tsdoc from "eslint-plugin-tsdoc";
 
 export default ts.config(
     js.configs.recommended,
@@ -27,6 +28,12 @@ export default ts.config(
                 parser: ts.parser,
             },
         },
+        plugins: {
+            tsdoc,
+        },
+        rules: {
+            "tsdoc/syntax": "warn",
+        },
     },
     {
         files: ["**/*.svelte.ts", "*.svelte.ts"],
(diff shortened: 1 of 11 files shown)


eslint.config.js
```diff
@@ -36,6 +43,24 @@ export default ts.config(
...
+        languageOptions: {
+            parser: ts.parser,
+        },
+        plugins: {
+            tsdoc,
+        },
+        rules: {
+            "tsdoc/syntax": "warn",
+        },
     },
```
Merseleo, line 64: Why do you always add the same block? Is it not possible to only extend the list of files (or regexes for file endings)?

src/lib/resource.svelte.ts
```diff
@@ -7,6 +18,11 @@
  *
  * The value of the state can be accessed via the `value` property.
  *
+ * The default value of {@link options.onSuccess} is the identity function
+ * (**Note:** This might cause errors when {@link TPromise} and {@link TValue} are not they same).
```
Merseleo, line 22:
```suggestion
 * (**Note:** This might cause errors when *TPromise* and *TValue* are not they same).
```

TPromise and TValue are not defined anywhere so it is not possible to create a link to description or so.

Merseleo (changes requested):
Your changes show very well that I probably sometimes didn't quite stick to the style, so it was a very good idea to add the linter and after your changes no error is displayed anymore, so it is (almost) already to be merged and used.

src/lib/resource.svelte.ts
Slartibartfass2: They are. In the resource function. Are you sure that this is not WebStorm not working correctly with tsdoc?

eslint.config.js
Slartibartfass2: I just thought it makes sense to keep them separate if we, e.g., only add a plugin to svelte files later.

src/lib/resource.svelte.ts
Merseleo: This is also possible. I tested it again in VS Code and the it jumps to the generics when I click on the "TPromise" link for example. So probably it is WebStorm, although i am still the opinion that it shouldn't be a link as they are generic values, but we can keep it.

eslint.config.js
Merseleo: Ok, I understand the thought process.

Merseleo (approved):
I think we are good to go!

PR #260, opened 2025-03-20
Add settings for start and stop the review mode

### LINKED ISSUE #50: [Use Case]: Start Review Mode

Project Member

### Describe the use case

As project member, I want to start the review mode to review not decided papers.

### Acceptance criteria

- review mode is the paper view but without information about other reviews/decisions

### Anything else?

Related to #30, #34 and #51

### LINKED ISSUE #51: [Use Case]: Stop Review Mode

Project Member

### Describe the use case

As project member, I want to stop the review mode after I've started it.

### Acceptance criteria

- if the review mode is not active, the review mode cannot be stopped
- only when the review mode was previously started, it can be stopped

### Anything else?

Related to #30, #34 and #50

Merseleo:
Closes #50
Closes #51

## What I have made

- I added the `UserSettingsLayout` with the three basic "tabs" and added a fourth tab for user review settings, as imo the review mode is no project settings (like in the mockup), but a user setting (same for keywords settings, which should be moved imo).
- I added a `ReviewModeSettings` component that allows the user to activate / deactivate the review mode via a switch
  - the review mode is stored as global state variable in svelte and I replaced every occurrence of a `reviewMode` or `showReviewStatus` variable by the access to the global state
  - furthermore this state is permanently stored in the local storage of the browser, to the state doesn't change, if the user reloads the page, navigates to another page or close and reopen the browser (as we still use SSR and probably should continue the use, I do faced the problem of always reading the local storage, as it doesn't exist on the server. I solved this problem by reading the local storage in the layout.svelte when the layout is mounted and the local storage exists, but it may not the best solution, so I am open to new ideas).

> **Note**: The large amount of changed files is due to the fact that I had to refactor a lot, because the review mode was always a property everywhere until now and therefore had to be deleted in many places and there were no user settings yet, whose rough layout I added

## Checklist

  - [x] integration tests


Merseleo:
Note for myself: check, whether we introduce a risk of flaky tests with the global state in the tests

src/lib/components/composites/navigation-bar/ProjectNavigationBar.svelte
```diff
@@ -41,6 +41,6 @@
     backRef="/"
     {defaultTabValue}
     loadingTitle={loadingProject.then((project) => project.name)}
-    tabs={tabs as unknown as Tab[]}
+    tabs={tabs as unknown as LinkTab[]}
```
luca-schlecker: Couldn't this cast to `LinkTab[]` be moved to the definition of `tabs`. This should also allow you to drop the cast to `unknown`.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -91,7 +90,7 @@ In the bottom, there are buttons to accept, decline or mark the paper as undecid
 Additionally, there are buttons to navigate to the previous or next paper.
 
 - when `showButtonBar` is false, then no buttons are shown at the bottom of the page
-- when `userConfig.isReviewMode` is false, then no decision buttons are shown
+- when `reviewModeState.inReviewMode` is false, then no decision buttons are shown
```
luca-schlecker:
```suggestion
- when `reviewMode.isActivated` is false, then no decision buttons are shown
```

src/lib/components/composites/settings/user-settings/ReviewModeSettings.svelte
```diff
@@ -0,0 +1,33 @@
...
+@component
+Component for changing the review mode setting.
+
+Usage:
+```svelte
+    <ReviewModeSettings />
+```
+-->
+<SettingsSection sectionTitle="Review mode">
+    <div class="items-top flex flex-row space-x-2">
```
luca-schlecker, line 18: image
image

I tested it with `items-center` here and no gap for the inner div. Which one do you prefer?

luca-schlecker (changes requested):
Looking great! :+1: I skipped over the parts I already reviewed in #171 (mostly the layouts and tabs). Just three minor things and then this should be good to go.

src/lib/components/composites/navigation-bar/ProjectNavigationBar.svelte
Merseleo: Wasn't my code, but you are right, I typed the `tabs` variable and deleted the casts.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
Merseleo: Great catch, thank you.

src/lib/components/composites/settings/user-settings/ReviewModeSettings.svelte
Merseleo: In my opinion the first solution looks better, as the toggle visually belongs to the "Activate ..." text and the explanation text is additional. The second solution mix this a bit and also seems unusual to me as it is not very common to vertically center a switch afaik.

Merseleo (commented):
Thank you for the review. I fixed the two errors.

Regarding the stuff I picked from #171: Which files exactly you didnt review, because I also updated some stuff, which should be reviewed.

src/lib/components/composites/settings/user-settings/ReviewModeSettings.svelte
luca-schlecker: Android does something like I proposed. iOS afaik has the current behavior (though they mostly don't have a description). It's a design choice, I am fine with either way. If you feel like it, you can ask @Slartibartfass2 what he thinks about it and "break the tie", but that's not necessary.

Android Settings Toggle:
image

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
luca-schlecker: It doesn't seem like you changed anything?

src/lib/components/composites/navigation-bar/ProjectNavigationBar.svelte
luca-schlecker: The branch doesn't seem to have the changes you've made.

luca-schlecker (changes requested):

src/lib/components/composites/navigation-bar/ProjectNavigationBar.svelte
Merseleo: I am lost, it didn't push the changes.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
Merseleo: I am lost, it didn't push the changes.

src/lib/components/composites/settings/user-settings/ReviewModeSettings.svelte
Merseleo: But I agree, it doesn't hurt to ask.

Merseleo (commented):
So I hope my changes are pushed now.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -91,7 +90,7 @@ In the bottom, there are buttons to accept, decline or mark the paper as undecid
 Additionally, there are buttons to navigate to the previous or next paper.
 
 - when `showButtonBar` is false, then no buttons are shown at the bottom of the page
-- when `userConfig.isReviewMode` is false, then no decision buttons are shown
+- when `reviewMode.inReviewMode` is false, then no decision buttons are shown
```
luca-schlecker:
```suggestion
- when `reviewMode.isActivated` is false, then no decision buttons are shown
```

luca-schlecker (changes requested):

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
Merseleo: Oh oh oh, i am very unconcentrated

src/lib/components/composites/settings/user-settings/ReviewModeSettings.svelte
Merseleo: I asked Felix and he prefers also the first variant.

luca-schlecker (approved):
Great Work! Everything I had in mind seems to be addressed. LGTM! 🚀

PR #261, opened 2025-03-21
Remove Slide Animation of Dialogs

luca-schlecker:
The diagonal sliding of dialogs looks weird to me. At first, I thought it was a bug, but it is in fact intentional.
I'd like to propose removing the slide animation, which is exactly what this PR is doing.
What are your thoughts on this? Do you prefer the sliding look? Let me know! :smiley:

### DIFF
--- a/src/lib/components/primitives/dialog/dialog-content.svelte
+++ b/src/lib/components/primitives/dialog/dialog-content.svelte
@@ -21,7 +21,7 @@
     <Dialog.Overlay />
     <DialogPrimitive.Content
         class={cn(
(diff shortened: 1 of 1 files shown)


Slartibartfass2:
Not completely sure about this, but I think this happens since we migrated to tailwind 4, which was in retrospect not the best idea because shadcn/svelte doesn't support it yet. But I read something about it in the shadcn/svelte repo.

Merseleo:
> Not completely sure about this, but I think this happens since we migrated to tailwind 4, which was in retrospect not the best idea because shadcn/svelte doesn't support it yet. But I read something about it in the shadcn/svelte repo.

At least on the shadcn website the dialog don't slide in but only appears. I just compared both options and I agree to Luca that it looks weird. So in my opinion, opening the dialog looks better with the fix.

Merseleo (approved):

PR #265, opened 2025-03-23
Open paper details card in project papers view

### LINKED ISSUE #219: [Use Case]: View Paper Details in Project Papers View

Project Member

### Describe the use case

As a project member, I want to view the paper details in a side panel, when I select a paper in the project papers view.

### Acceptance criteria

- when paper is selected, then paper details are shown like in the paper view
- in the side panel, the user can bookmark and navigate to the paper
- if the user clicks anywhere else, the side panel is hidden
- the panel entry and exit should be animated e.g., sliding in and out

### Anything else?

_No response_

Merseleo:
Closes #219

## What I have made

- I completed the card containing the paper details in the project papers view
  - to reuse the `PaperDetailsCard`, i extracted its content to a separate svelte component that can be used for this card, as the entire `PaperDetailsCard` could not be used in the project papers view because of size and styling differences
  - I extended this card by adding two icon buttons for opening the paper or add / remove the paper to / from the reading list
  - I added an animation, so the card for the paper details slides / flies in and out
- I added an action that dispatches an event, when a user clicks outside the referring component or presses "Escape" and integrated this into the project papers view, so the "`PaperDetailsCard`" is closed when the user clicks outside this card or any other `PaperListEntry` component (when the user clicks on another `PaperListEntry` component we get a closing and reopening animation, which somehow looks weird, so I am open for suggestions)

## Checklist

  - ~~[ ] unit tests~~ (not necessary, as action was tested by integration test)
  - [x] integration tests
  - [x] end-to-end tests


src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -143,13 +151,37 @@
...
-        </Card.Root>
+    {#if loadingPaper && selectedPaper}
+        <aside
+            class="h-full w-[65%] min-w-75"
+            data-testid="paper-details-card"
+            onClickedOutsideOrEscape={() => {
+                selectedPaper = undefined;
+            }}
+            use:clickOutsideOrEscape
+            transition:fly={{ duration: 100, x: 100, opacity: 0 }}
```
Comment (mowi12) [inline src/routes/project/[projectId]/papers/+page.svelte]:
```suggestion
            transition:fly={{ duration: 250, x: 150, opacity: 0 }}
```

I tried it with these values. Imo it looked better but felt a bit slow. Maybe something in between?
All the other animation types don't look as good as `fly`.

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,26 +4,32 @@
     import BookmarkMinus from "lucide-svelte/icons/bookmark-minus";
     import Tooltip from "./utils/Tooltip.svelte";
     import { resource } from "$lib/resource.svelte";
+    import { onMount } from "svelte";
 
     interface Props {
         loadingPaperId: Promise<string>;
-        isBookmarkedDefault: boolean;
+        isBookmarkedDefault?: boolean;
```
mowi12, line 11: Why is the modification of `isBookmarkedDefault` to `isBookmarkedDefault?` needed?

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,26 +4,32 @@
...
+    const { loadingPaperId, isBookmarkedDefault = false }: Props = $props();
 
     let isBookmarked = $state(isBookmarkedDefault);
     let isHovered = $state(false);
-    const tooltipText = $derived(isBookmarked ? "Remove from Reading List" : "Add to Reading List");
-    const paperId = resource<string, string | undefined>(loadingPaperId, {
-        initialValue: undefined,
-        resourceName: "paper ID",
-    });
+    const tooltipText = $derived(isBookmarked ? "Remove from reading list" : "Add to reading list");
```
mowi12, line 18: Why did you change the case of `Reading List`. Since it's not even a complete sentence, why not using header case or what it's called? Do we handle this similarly in other cases?

tests/integration/actions/actions.test.ts
```diff
@@ -0,0 +1,37 @@
+import { describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/svelte";
+import userEvent from "@testing-library/user-event";
+import ClickOutsideOrEscapeActionComponent from "./ClickOutsideOrEscapeActionComponent.svelte";
+
+describe("Click outside or press escape action", () => {
+    it("When a component is created with the 'clickOutsideOrEscape' action and the user clicks outside, then the 'ClickedOutsideOrEscape' event is fired.", async () => {
```
mowi12:
```suggestion
    test("When a component is created with the 'clickOutsideOrEscape' action and the user clicks outside, then the 'ClickedOutsideOrEscape' event is fired.", async () => {
```

As I understand it, there are two keywords for tests `it` and `test`. The reasons is to make a sentence with the keyword and the following sentence. So maybe use `test` instead of `it`?

This, of course, applies to all vitest tests.

tests/e2e/fixtures/general-fixture.ts
```diff
@@ -33,7 +33,7 @@ export const defaultTest = base.extend<{ forEachTest: void } & TestOptions>({
 
             await use();
         },
-        { auto: true },
+        { auto: true, box: true },
```
mowi12, line 36: What exactly is `box: true` needed for?

tests/e2e/pom/create-project-dialog-model.ts
```diff
@@ -40,6 +40,9 @@ export class DevCreateProjectDialog {
         await this.projectNameInput.fill(projectName);
         await this.projectMemberInput.fill("john@doe.com");
         await this.projectMemberInput.press("Tab");
+        await this.projectMemberInput.fill("Alice");
+        await this.projectMemberInput.press("ArrowDown");
+        await this.projectMemberInput.press("Enter");
```
mowi12, line 45: This is not important, but is it wanted that the user Alice has the email address of John Doe?

mowi12 (changes requested):
All in all, your code is very clean, and I couldn't find much to criticise.

I am not quite sure, but I found 2 other issue regarding filtering and searching the pages in the different stages. Currently, this functionality does not work, but in my understanding this is part of other issues (as well as adding a new paper). If this is the case, this comment can be ignored.

After fiddling around for a long time, I couldn't find a fix for the next problem. Please see to it, if you can find a way around.
As soon as you click on a paper, and the `PaperDetailsCard` appears, a horizontal scrollbar appears as well, but it's not necessary since all the content is still visible.

---

I think the following problem can be handled in this issue:
Because the value of a `PaperDetail` is displayed in a `textarea`, long titles (and maybe also other fields) can reach a length where a scrollbar appears. This looks very weird, and maybe just a line break should appear. I have no further suggestions since this can also become a separate issue since it is also effecting the paper details view of each separate paper.

Screenshot 2025-04-20 at 08 54 47

tests/e2e/fixtures/general-fixture.ts
Merseleo: This is not needed, but kind of useful in my opinion. Using `box: true` causes this fixture step to disappear from the UI / log (see [here](https://playwright.dev/docs/test-fixtures#box-fixtures) for more information) and as this step is always applied I think we could remove it from the log.

Merseleo: But if you think we should keep this in the log for completeness I will remove it.

tests/e2e/pom/create-project-dialog-model.ts
Merseleo: It shouldn't have this email address. Actually we add two members, `john@doe.com` by email address and `Alice Smith` by name? Or did you find a but with this test?

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte]: I changed it so a duration of 200ms, what's your opinion on this change?

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: It is not needed, but imo it removes unnecessary code. I prefer to use optional parameters when you often set a prop to the same value (as here) and only want to change it occasionally. Using optional parameters then makes the code more readable and less cluttered as you only set the values you explicitly want to set.

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: I think we agreed to generally not use this "header style" but only capitalize the first letter.

tests/integration/actions/actions.test.ts
Merseleo: I adapted it.

Merseleo (commented):
> All in all, your code is very clean, and I couldn't find much to criticise.
> 
> I am not quite sure, but I found 2 other issue regarding filtering and searching the pages in the different stages. Currently, this functionality does not work, but in my understanding this is part of other issues (as well as adding a new paper). If this is the case, this comment can be ignored.

You are right, these are part of multiple other issues (e.g. #38) but should also be done soon in this sprint.

> I think the following problem can be handled in this issue: Because the value of a `PaperDetail` is displayed in a `textarea`, long titles (and maybe also other fields) can reach a length where a scrollbar appears. This looks very weird, and maybe just a line break should appear. I have no further suggestions since this can also become a separate issue since it is also effecting the paper details view of each separate paper.

Regarding this problem. I did not implement the `PaperDetail` but @Slartibartfass2. I wrote an issue for this bug and do not want to mix this bug fix with this new feature (see #339)

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: You're right.

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: Okay.

tests/e2e/fixtures/general-fixture.ts
mowi12: No, we can definitely keep it. Makes sense now!

tests/e2e/pom/create-project-dialog-model.ts
mowi12: I had the wrong component in mind. You're right!

mowi12 (changes requested):
I'm happy with the changes you made!
The only things you missed are the vertical scroll bar that appears, as soon as you open the paper details card.

And on my end, the integration tests do not pass. You changed the case of `Reading List` and probably forgot to change it in all tests.
For me, it was the `reading-list-entry.test.ts` file.

Lastly, it seems you have a bug.
If you click on a paper in the papers view, wait a second and click again, this click is counted as a double click instead of simply clicking again on the paper.

src/routes/project/[projectId]/papers/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/papers/+page.svelte]: Yes, looks good to me.

Merseleo:
> I'm happy with the changes you made! The only things you missed are the vertical scroll bar that appears, as soon as you open the paper details card.
> 
> And on my end, the integration tests do not pass. You changed the case of `Reading List` and probably forgot to change it in all tests. For me, it was the `reading-list-entry.test.ts` file.
> 
> Lastly, it seems you have a bug. If you click on a paper in the papers view, wait a second and click again, this click is counted as a double click instead of simply clicking again on the paper.

I hopefully fixed all problems. If you do not find anything and the CI runs without any fails, you can merge it on your own.

mowi12 (approved):
I found no problem with the current changes, nicely done!

PR #268, opened 2025-03-23
Added initial docker setup

### LINKED ISSUE #267: [Feature]: Add docker image

_No response_

Let's add a docker image for this project.

A Dockerfile and a compose.yaml

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #267

## What I have made

- we now use the node adapter to build a node server that serves our frontend
- I added a docker image and several compose profiles to run the frontend (+mock backend) with different configurations
- `PUBLIC_API_BASE_URL` is now a dynamic variable because it needs to be loaded after the server was built, i.e., it is passed at run time. I added a check that immediately shuts down the server if the variable is not provided.
- I added a deployment script that is very similar to the one used in the mock backend.

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+# Include any files or directories that you don't want to be copied to your
+# container here (e.g., local build artifacts, temporary files, etc.).
+#
+# For more help, visit the .dockerignore file reference guide at
+# https://docs.docker.com/go/build-context-dockerignore/
+
+**/.classpath
+**/.dockerignore
+**/.env
+**/.git
+**/.gitignore
+**/.project
+**/.settings
+**/.toolstarget
+**/.vs
+**/.vscode
+**/.next
+**/.cache
+**/*.*proj.user
+**/*.dbmdl
+**/*.jfm
+**/charts
+**/docker-compose*
+**/compose.y*ml
+**/Dockerfile*
+**/node_modules
+**/npm-debug.log
+**/obj
+**/secrets.dev.yaml
+**/values.dev.yaml
+**/build
+**/dist
+LICENSE
+README.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -32,3 +32,4 @@
 **/dist
 LICENSE
 README.md
+src/lib/model/api/**
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,7 +57,6 @@ jobs:
                       type=semver,pattern={{major}},enable=${{ github.ref_type == 'tag' }}
                       type=raw,value=latest-dev,enable=${{ github.ref_name == 'develop' }}
                       type=raw,value=latest,enable=${{ github.ref_name == 'main' }}
-                      type=ref,event=branch,enable=${{ github.ref_type == 'branch' }}
                   flavor: |
                       latest=false
 
@@ -79,6 +78,7 @@ jobs:
                   push-to-registry: true
 
     deploy-frontend:
+        name: Deploy Frontend
         if: github.ref_name == 'develop' && github.ref_type == 'branch'
         runs-on: ubuntu-latest
         needs: build-and-publish-docker-image
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,105 @@
+name: Deploy
+
+on:
+    push:
+        branches: ["main", "develop"]
+        tags: ["v*"]
+
+env:
+    REGISTRY: ghcr.io
+    IMAGE_NAME: ${{github.repository}}
+    CONTAINER_NAME: ${{ github.event.repository.name }} # only repository name
+
+jobs:
+    build-and-publish-docker-image:
+        name: Build and Publish Docker Image
+        runs-on: ubuntu-latest
+        permissions:
+            contents: read
+            packages: write
+            attestations: write
+            id-token: write
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+              with:
+                  submodules: "recursive"
+
+            - name: Login to GitHub Container Registry
+              uses: docker/login-action@v3
+              with:
+                  registry: ${{ env.REGISTRY }}
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Set up QEMU
+              uses: docker/setup-qemu-action@v3
+
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v3
+
+            # This step uses metadata-action to extract tags and labels that will be
+            # applied to the specified image. The `id` "meta" allows the output of
+            # this step to be referenced in a subsequent step. The `images` value
+            # provides the base name for the tags and labels.
+            # We add the following tags
+            # - version tags: for v1.2.3 -> v1, v1.2, v1.2.3
+            # - latest tags: for develop -> latest-dev; for main -> latest
+            # - branch tags: just the branch name
+            - name: Extract metadata (tags, labels) for Docker
+              id: meta
+              uses: docker/metadata-action@v5
+              with:
+                  images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+                  tags: |
+                      type=semver,pattern={{version}},enable=${{ github.ref_type == 'tag' }}
+                      type=semver,pattern={{major}}.{{minor}},enable=${{ github.ref_type == 'tag' }}
+                      type=semver,pattern={{major}},enable=${{ github.ref_type == 'tag' }}
+                      type=raw,value=latest-dev,enable=${{ github.ref_name == 'develop' }}
+                      type=raw,value=latest,enable=${{ github.ref_name == 'main' }}
+                      type=ref,event=branch,enable=${{ github.ref_type == 'branch' }}
+                  flavor: |
+                      latest=false
+
+            - name: Build and push Docker image
+              id: push
+              uses: docker/build-push-action@v6
+              with:
+                  context: .
+                  push: true
+                  platforms: linux/amd64,linux/arm64
+                  tags: ${{ steps.meta.outputs.tags }}
+                  labels: ${{ steps.meta.outputs.labels }}
+
+            - name: Generate artifact attestation
+              uses: actions/attest-build-provenance@v1
+              with:
+                  subject-name: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME}}
+                  subject-digest: ${{ steps.push.outputs.digest }}
+                  push-to-registry: true
+
+    deploy-frontend:
+        if: github.ref_name == 'develop' && github.ref_type == 'branch'
+        runs-on: ubuntu-latest
+        needs: build-and-publish-docker-image
+        steps:
+            - name: Set up SSH
+              run: |
+                  mkdir -p ~/.ssh
+                  echo "${{ secrets.SSH_PRIVATE_KEY }}" > ~/.ssh/id_rsa
+                  chmod 600 ~/.ssh/id_rsa
+            - name: Convert image name to lowercase
+              run: |
+                  echo "IMAGE_NAME_LOWER=${IMAGE_NAME,,}" >> $GITHUB_ENV
+            - name: Pull and (re)start docker image
+              run: |
+                  ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ${{ secrets.REMOTE_USER }}@${{ secrets.REMOTE_HOST }} << EOF
+                    echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+                    docker pull ${{ env.REGISTRY }}/${{ env.IMAGE_NAME_LOWER }}:latest-dev
+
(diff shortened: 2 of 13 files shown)


.dockerignore
```diff
@@ -0,0 +1,35 @@
...
+**/node_modules
+**/npm-debug.log
+**/obj
+**/secrets.dev.yaml
+**/values.dev.yaml
+**/build
+**/dist
+LICENSE
+README.md
+src/lib/model/api/**
```
Merseleo, line 35: Why do you exclude these files? Are the irrelevant for gRPC runtime?

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+################################################################################
+# Create a new stage to run the application with minimal runtime dependencies
+# where the necessary files are copied from the build stage.
+FROM base AS final
+
+# Use production node environment by default.
+ENV NODE_ENV=production
+
+# Copy env variables to file so that it can be read by svelte
+RUN printenv | sed -n '/^PUBLIC_API_BASE_URL=/p;' > .env
```
Merseleo, line 44: Do you think it is clever to only copy the PUBLIC_API_BASE_URL? This does not seem very easy to maintain to me, if we expect other environment variables in the .env file later on, we will have to add them manually. Or is there a good reason why you want to restrict it like this?

Merseleo (commented):
Regarding the "problem" with the .env file: Did you read this section [here](https://svelte.dev/docs/kit/adapter-node#Environment-variables)? I think this could also be solution to your problem (if I understood it correctly) and seems a bit more clean to me.

README.md
```diff
@@ -12,56 +12,41 @@ You can find all use cases
...
 ```bash
 git clone git@github.com:SE-UUlm/snowballr-frontend.git
 cd snowballr-frontend
 git submodule update --init --recursive
-npm install
+docker-compose up
 ```
 
-Next, generate the API code[^tag-checkout]:
+Be sure to have the environment variables set or create a `.env` file in the root directory of the project (see below).
```
Merseleo: Maybe link to the section below?

README.md
```diff
@@ -12,56 +12,41 @@ You can find all use cases
...
 
-You can find the generated API code in the `src/lib/model/api` directory.
+- `mock-local`: Starts the frontend and the mock backend in a local setup.
+- `mock-public`: Starts the frontend and the deployed mock backend.
 
-To run the app, you need to create a `.env` file in the root directory of the project.
-The file should contain the following environment variables
-(see [.env.example](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/.env.example)
-for an example or copy it using `cp .env.example .env`):
+Use `docker-compose --profile <profile> up` to start the desired profile.
```
Merseleo:
```suggestion
Use `docker-compose --profile <profile> up` to start the frontend with the desired profile.
```

README.md
```diff
@@ -12,56 +12,41 @@ You can find all use cases
...
 
-```bash
-npm run build
-```
-
-You can preview the production build with `npm run preview`.
-
-If you don't have a backend server running, you can use the mock backend.
-Head over to the [snowballr-mock-backend](https://github.com/SE-UUlm/snowballr-mock-backend) repository and follow the
-instructions there.
```
Merseleo, line 60: You deleted this section, but in the "Getting Started" section you speak of the mock backend. I understand that this paragraph does not fit here, but maybe you shouldn't delete the link to the mock backend and move it to the docker compose profile paragraph?

README.md
```diff
@@ -12,56 +12,41 @@ You can find all use cases
...
-
-To build the project from source, run the following commands:
+The fastest way to get started is to use the provided Docker setup. To do so, run the following commands:
 
 ```bash
 git clone git@github.com:SE-UUlm/snowballr-frontend.git
 cd snowballr-frontend
 git submodule update --init --recursive
-npm install
+docker-compose up
```
Merseleo: Are you sure that it is `docker-compose`? Because I use `docker compose` (same applies for the other commands below)

compose.yaml
```diff
@@ -0,0 +1,54 @@
+services:
+    # standalone frontend
+    frontend:
+        build:
+            context: .
+        environment:
+            NODE_ENV: production
+            PORT: ${PORT:-4173}
```
Merseleo, line 8: If I am not lost, you use the PORT environment variable or 4173 as default port, correctly?
If so, please add the PORT environment to the environment variable table in the REAMDE.md and add a sentence after the QuickStart describing on port (+ url) the frontend can be opened.

Dockerfile
Slartibartfass2: I thought it made sense to only copy the values which are required by the app. This eliminates the risk that some other variable might interfere with the program.

.dockerignore
Slartibartfass2: The generated API code shouldn't be copied to the docker image, as it is generated in there separately. This way, it is not required to generate the API code before building the image.
TL;DR; it should be possible to create the image without having to execute commands.

Slartibartfass2:
> Regarding the "problem" with the .env file: Did you read this section [here](https://svelte.dev/docs/kit/adapter-node#Environment-variables)? I think this could also be solution to your problem (if I understood it correctly) and seems a bit more clean to me.

I'm not sure what you mean by that. We already use the `--env-file` flag in the Dockerfile.

Slartibartfass2 (commented):
Thanks for the review. I'm very excited to merge this.

compose.yaml
```diff
@@ -0,0 +1,54 @@
...
+            PUBLIC_API_BASE_URL: http://snowballr.informatik.uni-ulm.de:3001
+        ports:
+            - ${PORT:-4173}:${PORT:-4173}
+        profiles:
+            - mock-public
```
Merseleo, line 54: Did you check this profile? Because when I try it, it doesn't work and the frontend displays errors indicating that the mock backend is not found.

compose.yaml
Merseleo: And why do you need the PORT environment variable at all?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,105 @@
...
+            # - branch tags: just the branch name
+            - name: Extract metadata (tags, labels) for Docker
+              id: meta
+              uses: docker/metadata-action@v5
+              with:
+                  images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+                  tags: |
+                      type=semver,pattern={{version}},enable=${{ github.ref_type == 'tag' }}
+                      type=semver,pattern={{major}}.{{minor}},enable=${{ github.ref_type == 'tag' }}
+                      type=semver,pattern={{major}},enable=${{ github.ref_type == 'tag' }}
```
Merseleo, line 57: Why do you want to add these tags as well? Isn't the v1.2.3 enough?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,105 @@
...
+            - name: Generate artifact attestation
+              uses: actions/attest-build-provenance@v1
+              with:
+                  subject-name: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME}}
+                  subject-digest: ${{ steps.push.outputs.digest }}
+                  push-to-registry: true
+
+    deploy-frontend:
+        name: Deploy Frontend
+        if: github.ref_name == 'develop' && github.ref_type == 'branch'
```
Merseleo, line 82: Why do we only deploy the "develop" version? But if we keep it that way, maybe we should add a note in the README.md that we have a running deployed frontend, but that it is only a dev version and frequently changes?!

wiki/Getting-Started.md
Merseleo, line 1: If you apply the changes in the README.md dont forgot to change it here as well.

wiki/Getting-Started.md
```diff
@@ -1,32 +1,59 @@
...
+To build the project from source, run the following commands:
+
+```bash
+git clone git@github.com:SE-UUlm/snowballr-frontend.git
+cd snowballr-frontend
+git submodule update --init --recursive
+npm install
+```
+
+Next, generate the API code[1](#footnote-1):
```
Merseleo:
```suggestion
Next, generate the API code [[1](#footnote-1)]:
```

Merseleo (changes requested):
So now the real review. Very good work and I'm glad that we will soon have a deployed frontend with which we can test snowballr like a real user, thanks for doing this so quickly.

.github/workflows/deploy.yml
Slartibartfass2: As [other docker images](https://hub.docker.com/_/alpine) use this too, I thought it would be quite nice.
For instance, one could pull the version `1.2` and will always receive the latest patches in that minor version.

Slartibartfass2: I see that there's a very low chance that this will be the case, but it doesn't harm either.

.github/workflows/deploy.yml
Slartibartfass2: Currently, we do not have releases, later I would rather only deploy on `main`. As we are in an early stage, I think it's fine to assume that there are frequent changes.

README.md
Slartibartfass2: That's a typo. Good catch!

README.md
Slartibartfass2: I added it.

README.md
Slartibartfass2: Added it.

README.md
Slartibartfass2: I added it to the section above. How do you like it now?

wiki/Getting-Started.md
Slartibartfass2: They should be in sync now. Maybe we should only name the existing docker image in the readme and refer to the wiki instead.

wiki/Getting-Started.md
Slartibartfass2: Changed it.

compose.yaml
Slartibartfass2: Yes that's true. The variable is required for better configuration possibilities.

Slartibartfass2: I added these variables to the documentation

compose.yaml
Slartibartfass2: I did. I only receive CORS errors, which we need to handle in the mock backend.

Slartibartfass2 (commented):
Thank you very much for the review.

.dockerignore
Merseleo: Yes, I saw later in the Dockerfile that this is generated, my comment was stupid, sorry.

Dockerfile
Merseleo: That makes sense.

README.md
Merseleo: Great 👍

README.md
Merseleo: I would like it more, if you would write "Mock Backend" lowercase as we do it everywhere.

Merseleo: But apart from that: yes, it should fit.

README.md
```diff
@@ -12,56 +12,48 @@ You can find all use cases
...
-```bash
-npm run build
-```
-
-You can preview the production build with `npm run preview`.
+| Variable              |      Required      | Default | Description                                                                             |
+| --------------------- | :----------------: | :-----: | --------------------------------------------------------------------------------------- |
+| `PUBLIC_API_BASE_URL` | :white_check_mark: |    -    | The URL of the backend API.                                                             |
+| `PUBLIC_IS_DEV_MODE`  |        :x:         |  false  | Whether the app is in development mode. This may enable additional development tooling. |
+| `PORT`                |       :x:\*        |  4173   | The port where the front end is served                                                  |
```
Merseleo:
```suggestion
| `PORT`                |       :x:\*        |  4173   | The port where the frontend is served                                                  |
```

compose.yaml
Merseleo: Thank you.

compose.yaml
Merseleo: Ok now it works on my machine as well, strange. Probably I had Internet connection problems.

wiki/Getting-Started.md
Merseleo: Lets discuss this stuff later and perhaps even in a larger group in general.

.github/workflows/deploy.yml
Merseleo: Ok, i am also fine with that.

.github/workflows/deploy.yml
Merseleo: Ok, didn't know or ever recognized that this is a thing, but your explanation makes sense and you are right it doesn't harm, so good idea 👍, it would have forgotten such an "addon".

Merseleo (changes requested):
> > Regarding the "problem" with the .env file: Did you read this section [here](https://svelte.dev/docs/kit/adapter-node#Environment-variables)? I think this could also be solution to your problem (if I understood it correctly) and seems a bit more clean to me.
> 
> I'm not sure what you mean by that. We already use the `--env-file` flag in the Dockerfile.

Nevermind - I didn't read correctly and thought that `dotenv` was needed in addition and not as an earlier alternative to `--env-file`. So forget the comment.

README.md
Slartibartfass2: I fixed it.

Merseleo (approved):

PR #269, opened 2025-03-23
Move documentation to wiki

### LINKED ISSUE #266: [Feature]: Move documentation to wiki

_No response_

We should have a wiki, move most of the infomation to it

Information in the wiki.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #266

## What I have made

In the long run, it probably is better to keep the wiki in this repo and copy it over in the pipeline, so we can do reviews on it.

## Checklist

- [ ] I have updated the documentation accordingly and commented my code
  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests


Merseleo:
@Slartibartfass2 What do you think of moving the documentation, so the current Wiki as well as documentation from the README.md into a new Wiki in the *snowballr-documentation* repository?
Because in the long run we should create a wiki like the team before us, so a static website which can be hosted, e.g. on the snowballr server, what do you think? Or is this a point, we should discuss with the other team members and / or Alex?

PR #287, opened 2025-03-31
Add option to hover over user avatar to show full name

### LINKED ISSUE #123: [Feature]: Make user avatar hoverable

_No response_

Currently, the user avatar only shows the initials and for the navigation bar that's fine.
However, e.g. in the paper list entry, when review decisions are shown with their respective user avatar, the user might wonder, to whom the initials belong to. If there are two project members with the same initials, they should be distinguishable.

I propose showing the user's full name in a tooltip while hovering.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #123 

## What I have made

- I extended the user avatar, so it can be (optionally) hovered to show the full name of the user

## Checklist

  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (not necessary)

### DIFF
--- a/src/lib/components/composites/user-avatar/UserAvatar.svelte
+++ b/src/lib/components/composites/user-avatar/UserAvatar.svelte
@@ -80,7 +80,7 @@ Usage:
                 </Avatar.Root>
             {/snippet}
             {#snippet content()}
-                {user?.firstName ?? ""} {user?.lastName ?? ""}
+                {user === undefined ? "Unknown" : `${user.firstName} ${user.lastName}`}
             {/snippet}
         </Tooltip>
     {:else}
--- a/src/lib/components/composites/user-avatar/UserAvatar.svelte
+++ b/src/lib/components/composites/user-avatar/UserAvatar.svelte
@@ -62,6 +62,8 @@ Usage:
         email: "john@doe.com",
         firstName: "John",
         lastName: "Doe",
+        role: UserRole.DEFAULT,
+        status: UserStatus.ACTIVE,
     }} reviewDecision={ReviewDecision.ACCEPTED} showUsernameOnHover={false} />
 ```
 -->
--- a/src/lib/components/composites/user-avatar/UserAvatar.svelte
+++ b/src/lib/components/composites/user-avatar/UserAvatar.svelte
@@ -68,22 +68,7 @@ Usage:
 ```
 -->
 <div class="relative" data-testid="user-avatar">
-    {#if showUsernameOnHover}
-        <Tooltip>
-            {#snippet trigger()}
-                <Avatar.Root class={style.avatarSize}>
-                    <Avatar.Fallback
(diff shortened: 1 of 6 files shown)


src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -57,15 +62,36 @@ Usage:
...
+        </Tooltip>
+    {:else}
+        <Avatar.Root class={style.avatarSize}>
+            <Avatar.Fallback
+                class={cn("group-hover/paper-list-entry:bg-slate-200", style.textSize)}
+            >
+                {userInitials}
+            </Avatar.Fallback>
+        </Avatar.Root>
+    {/if}
```
luca-schlecker: I suggest using [snippets](https://svelte.dev/docs/svelte/snippet) to reduce code duplication:
```
<Avatar.Root class={style.avatarSize}>
  <Avatar.Fallback
    class={cn("group-hover/paper-list-entry:bg-slate-200", style.textSize)}
  >
    {userInitials}
  </Avatar.Fallback>
</Avatar.Root>
```
This code could be a snippet shown using `{@render ...}`

src/lib/components/composites/user-avatar/UserAvatar.svelte
```diff
@@ -57,15 +62,36 @@ Usage:
...
+                <Avatar.Root class={style.avatarSize}>
+                    <Avatar.Fallback
+                        class={cn("group-hover/paper-list-entry:bg-slate-200", style.textSize)}
+                    >
+                        {userInitials}
+                    </Avatar.Fallback>
+                </Avatar.Root>
+            {/snippet}
+            {#snippet content()}
+                {user?.firstName ?? ""} {user?.lastName ?? ""}
```
luca-schlecker: This shows nothing (`""`) if the user is `undefined`, probably should show something like `Unknown`. Maybe put it inside `{#if} ... {/if}`?

luca-schlecker (changes requested):

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: Very good idea, I had no idea how to handle this case, but I like your suggestion and will implement it.

src/lib/components/composites/user-avatar/UserAvatar.svelte
Merseleo: You are right, we should avoid duplicate code.

Merseleo (commented):
So I hope I didn't forget anything in this PR ;)

luca-schlecker (dismissed):
LGTM 🚀

Merseleo:
I excuse the non-excusable: I didn't rebase, but merged develop as i wanted to resolve the conflict in the GitHub editor and tested this. My advice: dont do that!

luca-schlecker (approved):

PR #289, opened 2025-03-31
Fix "moving" navigation bar

### LINKED ISSUE #118: "Moving" navigation bar
# Problem

Currently, the height of the navigation bar is depending on the contents height. This results in "moving" components
in the navigation bar and a "moving" of the navigation bar itself, if the user navigates through the pages (see demonstration.mov:)

https://github.com/user-attachments/assets/8f8d88f2-48cd-404d-9c86-c1cb57e375ea

# Possible Solution

Set a fixed height of the navigation bar, so it only extends to the right, if more content is provided.

Merseleo:
Closes #118 

## What I have made

- I removed the padding of the error indicator, as the padding often was too much
- I set the height of the navigation bar(s) to a fixed height of 72px, which was the maximum (in the `ProjectNavigationBar`) value

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (not necessary)
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (not necessary)
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - ~~[ ] end-to-end tests~~

### DIFF
--- a/src/lib/components/composites/navigation-bar/NavigationBar.svelte
+++ b/src/lib/components/composites/navigation-bar/NavigationBar.svelte
@@ -26,7 +26,7 @@
 
 <header>
     <Card.Root class="w-fit shadow-lg">
-        <nav class="grid grid-flow-col items-center gap-3 px-4 py-3">
+        <nav class="grid h-18 grid-flow-col items-center gap-3 px-4 py-2">
             <UserMenu {user} />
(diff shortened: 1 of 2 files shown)


luca-schlecker (approved):
Can confirm, fixes the moving navbar. Way better now, great work! :rocket:

Issue #290, opened 2025-04-01
[Use Case]: Download paper


Merseleo:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to download / export a paper, I am interested in.

### Acceptance criteria

- when the download button (see `DownloadButton` component) is clicked, a file choose dialog should be opened where I can select at which location the paper should be saved and using which format
- when the format and location is selected, the paper object is formatted corresponding to the format and saved in the filesystem.

Slartibartfass2:
@Merseleo let's make this a use case

Merseleo:
Is it possible to convert this issue? Or do I have to manually apply the use case template?

Slartibartfass2:
You would have to copy the markdown but imo you just apply the label and the title prefix

Merseleo:
Ok

Merseleo:
For selecting an export file, the svelte shadcn button has a file chooser type: https://next.shadcn-svelte.com/docs/components/input#file

Slartibartfass2:
See https://github.com/SE-UUlm/snowballr-frontend/issues/98

Issue #100, opened 2024-11-21
[Use Case]: Remove Paper From Reading List


Slartibartfass2:
### Who is the primary actor in this use case?

User

### Describe the use case

As user, I want to remove a paper from my reading list.

### Acceptance criteria

- when removed, the paper won't be shown in the private reading list anymore

### Anything else?

Related to #7 and #99

Issue #99, opened 2024-11-21
[Use Case]: Add Paper To Reading List


Slartibartfass2:
### Who is the primary actor in this use case?

User

### Describe the use case

As user, I want to add a paper to my reading list.

### Acceptance criteria

- when added, the paper will be shown in the private reading list

### Anything else?

Related to #7 and #100

mowi12:
The corresponding PR can be found here: #351.

PR #291, opened 2025-04-01
Add view for reading list

### LINKED ISSUE #7: [Use Case]: View Reading List

User

### Describe the use case

As user, I want to see a list of all the papers that I added to my reading list.

### Acceptance criteria

- show reading list
- if a paper was added to the reading list, it should be shown
- if a paper wasn't added to the reading list, it shouldn't be shown

### Anything else?

Related to #99 and #100

Merseleo:
Closes #7

## What I have made

- I added a download button that allows the user to download / export a paper (download functionality is not implemented and will be done in #290!)
- I added a `ReadingListEntry` component along with a skeleton component
  - I refactored some skeleton components and created a `PaperInfoSkeleton` that is reused
- I added the view for the reading list and made the reading list searchable

## Checklist

  - ~~[] unit tests~~ (not necessary)
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/DownloadButton.svelte
+++ b/src/lib/components/composites/DownloadButton.svelte
@@ -1,0 +1,45 @@
+<script lang="ts">
+    import Tooltip from "./utils/Tooltip.svelte";
+    import { resource } from "$lib/resource.svelte";
+    import { Download } from "lucide-svelte";
+
+    interface Props {
+        loadingPaperId: Promise<string>;
+    }
+
+    const { loadingPaperId }: Props = $props();
+
+    const tooltipText = "Download this paper";
+    const paperId = resource<string, string | undefined>(loadingPaperId, {
+        initialValue: undefined,
+        resourceName: "paper ID",
+    });
+
+    function downloadPaper() {
+        // TODO: Will be implemented in #290
+        console.log(`Download paper with id ${paperId.value}.`);
+        dispatchEvent(new CustomEvent("downloadPaper"));
+    }
+</script>
+
+<!--
+@component
+Button download a paper.
+
+Usage:
+```svelte
+<DownloadButton loadingPaperId={42} />
+```
+-->
+<Tooltip
+    class="border-container-border-grey text-primary h-fit w-fit border bg-transparent p-1.5 hover:bg-transparent"
+    aria-label={tooltipText}
+    onclick={() => downloadPaper()}
+>
+    {#snippet trigger()}
+        <Download />
+    {/snippet}
+    {#snippet content()}
+        <p>{tooltipText}</p>
+    {/snippet}
+</Tooltip>
--- a/src/lib/components/composites/DownloadButton.svelte
+++ b/src/lib/components/composites/DownloadButton.svelte
@@ -24,7 +24,7 @@
 
 <!--
 @component
-Button download a paper.
+Button to download a paper.
 
 Usage:
 ```svelte
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -1,0 +1,53 @@
+<script lang="ts">
+    import { goto } from "$app/navigation";
+    import { handleSingleOrDoubleClick } from "$lib/utils/common-helper";
+    import type { ReadingListEntryInterface } from "$lib/model/component-interfaces";
+    import PaperInfo from "$lib/components/composites/paper-components/PaperInfo.svelte";
+    import PaperBookmarkButton from "$lib/components/composites/PaperBookmarkButton.svelte";
+    import DownloadButton from "$lib/components/composites/DownloadButton.svelte";
+
+    type ReadingListEntryProps = ReadingListEntryInterface & {
+        onClick?: () => void;
+    };
+
+    const navigateToPaperView = () => {
+        goto(`/paper/${id}`);
+    };
+
+    const { paper, onClick = navigateToPaperView }: ReadingListEntryProps = $props();
+    const { id, ...paperWithoutId } = paper;
+
+    let timeoutId: ReturnType<typeof setTimeout> | null = $state(null);
+    const handleClick = () => {
+        timeoutId = handleSingleOrDoubleClick(timeoutId, onClick, navigateToPaperView);
+    };
+</script>
+
+<!--
+@component
+Container displaying the information of a paper on the reading list.
+
+This component shows the
+  - title of the paper
+  - authors (or 'unknown authors' if none are specified)
+
+Furthermore this component is clickable and navigates to the corresponding paper view,
+if the onClick() event handler is not overridden. Otherwise it executes the custom event handler
+on a single click. A double click always causes the navigation to the paper view.
+
+Usage:
+```svelte
+    <ReadingListEntry paper={paper} />
+```
+-->
+<div
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-5 py-1.5"
+>
+    <button class="flex flex-auto" onclick={handleClick} type="button">
+        <PaperInfo class="gap-1" loadingPaper={Promise.resolve(paperWithoutId)} />
+    </button>
+    <div class="flex flex-row items-center gap-4">
+        <PaperBookmarkButton isBookmarkedDefault={true} loadingPaperId={Promise.resolve(id)} />
+        <DownloadButton loadingPaperId={Promise.resolve(id)} />
+    </div>
+</div>
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -42,7 +42,7 @@ Usage:
 ```
 -->
 <div
-    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-2.5 py-1.5"
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-2.5 py-2"
 >
     <button
         class="flex flex-auto"
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -1,14 +1,15 @@
 <script lang="ts">
     import { goto } from "$app/navigation";
     import { handleSingleOrDoubleClick } from "$lib/utils/common-helper";
-    import type { ReadingListEntryInterface } from "$lib/model/component-interfaces";
     import PaperInfo from "$lib/components/composites/paper-components/PaperInfo.svelte";
     import PaperBookmarkButton from "$lib/components/composites/PaperBookmarkButton.svelte";
     import DownloadButton from "$lib/components/composites/DownloadButton.svelte";
+    import type { Paper } from "$lib/model/api/paper";
 
-    type ReadingListEntryProps = ReadingListEntryInterface & {
+    interface ReadingListEntryProps {
+        paper: Paper;
         onClick?: () => void;
-    };
+    }
 
     const navigateToPaperView = () => {
         goto(`/paper/${id}`);
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -42,7 +42,7 @@ Usage:
 ```
 -->
 <div
-    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-2.5 py-2"
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border px-3 py-2"
 >
     <button
         class="flex flex-auto"
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -43,7 +43,12 @@ Usage:
 <div
     class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-5 py-1.5"
 >
-    <button class="flex flex-auto" onclick={handleClick} type="button">
+    <button
+        class="flex flex-auto"
+        aria-label="Paper info for reading list entry"
+        onclick={handleClick}
+        type="button"
+    >
         <PaperInfo class="gap-1" loadingPaper={Promise.resolve(paperWithoutId)} />
     </button>
     <div class="flex flex-row items-center gap-4">
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -41,7 +41,7 @@ Usage:
 ```
 -->
 <div
-    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-5 py-1.5"
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-2.5 py-1.5"
 >
     <button
         class="flex flex-auto"
--- a/src/lib/components/composites/paper-components/ReadingListEntrySkeleton.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntrySkeleton.svelte
@@ -1,0 +1,9 @@
+<script lang="ts">
+    import PaperInfoSkeleton from "$lib/components/composites/paper-components/PaperInfoSkeleton.svelte";
+</script>
+
+<div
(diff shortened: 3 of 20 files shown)


src/lib/components/composites/DownloadButton.svelte
```diff
@@ -0,0 +1,45 @@
...
+    const tooltipText = "Download this paper";
+    const paperId = resource<string, string | undefined>(loadingPaperId, {
+        initialValue: undefined,
+        resourceName: "paper ID",
+    });
+
+    function downloadPaper() {
+        // TODO: Will be implemented in #290
+        console.log(`Download paper with id ${paperId.value}.`);
+        dispatchEvent(new CustomEvent("downloadPaper"));
```
Slartibartfass2, line 21: Is this required or just for testing?

src/lib/components/composites/DownloadButton.svelte
```diff
@@ -0,0 +1,45 @@
...
+    }
+</script>
+
+<!--
+@component
+Button download a paper.
```
Slartibartfass2:
```suggestion
Button to download a paper.
```

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -0,0 +1,58 @@
...
+if the onClick() event handler is not overridden. Otherwise it executes the custom event handler
+on a single click. A double click always causes the navigation to the paper view.
+
+Usage:
+```svelte
+    <ReadingListEntry paper={paper} />
+```
+-->
+<div
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-2.5 py-1.5"
```
Slartibartfass2: What about using a full border? I know we did this in the design for whatever reason, but I think it looks weird. Especially when the paper list entry looks entirely different.
```suggestion
    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border px-2.5 py-1.5"
```

src/lib/components/composites/paper-components/ReadingListEntrySkeleton.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+    import PaperInfoSkeleton from "$lib/components/composites/paper-components/PaperInfoSkeleton.svelte";
+</script>
+
+<div
+    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border-b px-5 py-1.5"
```
Slartibartfass2: Same here with the same border and increase the padding to match the actual list entry.
```suggestion
    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border px-5 py-2"
```

Slartibartfass2: Now, only the paper list entry + skeleton has a different height. Maybe we should normalize this.
The project list entry has a height of 66px and the reading list entry too with this fix.

I propose the following changes:
`PaperListEntrySkeleton`:
```svelte
<div
    class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center justify-end gap-3 rounded-md border pe-3"
>
    <div class="flex flex-auto rounded-md border-l-4 px-3 py-2">
        <PaperInfoSkeleton />
    </div>
    {#if isSkeletonWithReview}
        <UserAvatarSkeleton />
    {/if}
</div>
```

and `PaperListEntry`:
```svelte
<!-- ... -->
    <div
        class={cn(
            "flex flex-auto rounded-md px-3 py-2",
            !reviewMode.isActivated && isProjectPaper(paper)
                ? `border-l-4 ${getStatusColor(getStatusText(paper), "border")}`
                : "",
        )}
    >
<!-- ... -->
```

src/lib/model/component-interfaces.ts
```diff
@@ -49,9 +49,14 @@ type PaperListEntryInterface =
           paper: Project_Paper;
       };
 
+interface ReadingListEntryInterface {
+    paper: Paper;
+}
```
Slartibartfass2: Why do we extract this to here? It's just a paper and I don't see what else is coming.

src/routes/readinglist/+page.svelte
```diff
@@ -1,11 +1,44 @@
...
-    const { user } = data;
+    const { user, loadingReadingList } = data;
+
+    let filteredReadingList = $state(loadingReadingList);
+
+    function filterReadingList(searchText: string) {
+        filteredReadingList = loadingReadingList.then((allEntries) =>
+            filterReadingListEntries(allEntries, searchText),
+        );
+    }
```
Slartibartfass2: I think you can use the resource helper here.

tests/e2e/project-creation.test.ts
```diff
@@ -11,6 +11,7 @@ test.describe("Creating a new project", () => {
     test.afterAll(async ({ mockBackendService }) => {
         mockBackendService.softDeleteProject({ id: "0" });
         mockBackendService.softDeleteProject({ id: "1" });
+        mockBackendService.softDeleteProject({ id: "2" });
     });
```
Slartibartfass2, line 15: What about this?
```suggestion
    test.afterEach(async ({ mockBackendService }) => {
        mockBackendService.softDeleteProject({ id: "0" });
    });
```

src/lib/utils/common-helper.ts
```diff
@@ -169,6 +169,32 @@ function comparePaperId(a: string, b: string): number {
...
+ * whether it was a single click (so no further click after 350ms) or a double click
+ * and call the corresponding functions:
+ *  - single click =\> onSingleClick()
+ *  - double click =\> onDoubleClick()
+ */
+function handleSingleOrDoubleClick(
+    timeoutId: ReturnType<typeof setTimeout> | null,
+    onSingleClick: () => void,
+    onDoubleClick: () => void,
+): ReturnType<typeof setTimeout> | null {
```
Slartibartfass2, line 183: This doesn't really belong here, but is related to it: Let's make an element without special double click functionality directly an a tag. This would work great with Svelte's preloading functionality.

tests/e2e/pom/reading-list-page-model.ts
```diff
@@ -0,0 +1,25 @@
+import { type Locator, type Page } from "@playwright/test";
+
+export class DevReadingListPage {
+    readonly page: Page;
+    readonly searchBar: Locator; // only locator to search bar input and not entire component
```
Slartibartfass2: What about this?
```suggestion
    readonly searchBarInput: Locator;
```

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+import { test } from "./fixtures/reading-list-page";
+import type { Paper } from "$lib/model/api/paper";
+import { createPaper } from "$tests/model-builder";
+import type { Project } from "$lib/model/api/project";
+
+test.describe("Add, view or remove papers to / from the reading list", () => {
+    let projectId: string = "";
+    const paperIds: string[] = [];
+
+    const NUMBER_OF_PAPER_ON_READING_LIST = 11;
```
Slartibartfass2:
```suggestion
    const NUMBER_OF_PAPERS_ON_READING_LIST = 11;
```
or
```suggestion
    const NUMBER_OF_PAPERS = 11;
```

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        const createdPapers: Paper[] = await Promise.all(
+            Array.from(
+                { length: NUMBER_OF_PAPER_ON_READING_LIST + 1 },
+                (_, i) =>
+                    mockBackendService.createPaper(
+                        createPaper({ title: `Paper ${i} on reading list` }),
+                    ).response,
+            ),
+        );
+        for (const paper of createdPapers.slice(0, NUMBER_OF_PAPER_ON_READING_LIST)) {
```
Slartibartfass2: Why are we slicing the whole list?

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+            paperIds.push(paper.id);
+        }
+    });
+
+    test.afterAll(async ({ mockBackendService }) => {
+        for (const paperId of paperIds) {
+            await mockBackendService.removePaperFromReadingList({ id: paperId });
+        }
+        mockBackendService.softDeleteProject({ id: projectId });
+    });
```
Slartibartfass2, line 42: Let's also delete the papers to be sure that nothing interferes.

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    }) => {
+        await expect(page.getByRole("listitem")).toHaveCount(NUMBER_OF_PAPER_ON_READING_LIST);
+
+        await expect(page.getByText(`Paper 0 on reading list`)).toBeVisible();
+        await expect(
+            page.getByText(`Paper ${NUMBER_OF_PAPER_ON_READING_LIST} on reading list`),
+        ).toBeHidden();
+    });
+
+    test("When the user search in the reading list, then only papers matching the search input are shown.", async ({
```
Slartibartfass2:
```suggestion
    test("When the user searches in the reading list, then only papers matching the search input are shown.", async ({
```

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        page,
+        readingListPage,
+    }) => {
+        await expect(page.getByRole("listitem")).toHaveCount(NUMBER_OF_PAPER_ON_READING_LIST);
```
Slartibartfass2: We use this several times in this file, let's put it into the reading list POM.

tests/e2e/reading-list.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        await expect(
+            page.getByRole("navigation").getByText("Paper 0 on reading list", { exact: false }),
+        ).toBeVisible();
+    });
+
+    test.fixme(
+        "When the user adds a paper to reading list in the paper view, " +
+            "then it will be displayed on the reading list page.",
+        async () => {},
+    );
```
Slartibartfass2: When will these be implemented?

Slartibartfass2 (changes requested):
Great work! Only some minor changes.

src/lib/components/composites/DownloadButton.svelte
Merseleo: No it is not required and only used for testing, especially since this method is not implemented and has no side effects that can be tested.

tests/e2e/project-creation.test.ts
Merseleo: This wouldn't work, as the ids increases as we only soft delete the projects.

tests/e2e/reading-list.test.ts
Merseleo: I create multiple paper but want to add some (in this case n-1) paper to the reading list to check, whether any created paper is displayed on the reading list or only the ones, that where added. And the slice operator is exclusive, the last element of the paper list should not be on the reading list.

tests/e2e/reading-list.test.ts
Merseleo: As far as I now we cannot delete papers?

tests/e2e/reading-list.test.ts
Merseleo: As soon as the issues #99 and #100 will be implemented.

tests/e2e/reading-list.test.ts
Merseleo: I updated the readingList locator and used this instead, is this solution what you imagined?

src/lib/model/component-interfaces.ts
Merseleo: I can try to only use the paper.

src/lib/utils/common-helper.ts
Merseleo: I agree, but do you have something I mind where we should use this and currently dont use it?

src/routes/readinglist/+page.svelte
Merseleo: But why should I, so what advantage does this have? It would make the code longer and more complicated, because I don't want to wait for the values anywhere, do I?

src/lib/components/composites/paper-components/ReadingListEntry.svelte
Merseleo: For me it's the other way round: in my opinion it looks weird with a full border and I've tried it myself and I don't think it looks nice. It also gives you a visual distinction between `PaperListEntry` and `ReadingListEntry`.
I would suggest asking someone else, what do you think?

src/lib/components/composites/paper-components/ReadingListEntrySkeleton.svelte
Merseleo: But I agree, it would look better if all list entry components have a padding of `py-2`.

Merseleo (commented):
Thank you for the review, I fixed all problems I understood. But some things were not entirely clear and some things we should discuss again

src/lib/utils/common-helper.ts
Slartibartfass2: We only use the double click feature on the project papers view. On the home page (open reviews), project dashboard (open reviews), paper view (references and citations), and now the reading list, we could wrap the list entries in `a` tags.

tests/e2e/project-creation.test.ts
Slartibartfass2: Ah yes, you're right. I forgot about that.

tests/e2e/reading-list.test.ts
Slartibartfass2: True, I thought we had a call for this.

tests/e2e/reading-list.test.ts
Slartibartfass2: Can you please reference these issues in a comment?

tests/e2e/reading-list.test.ts
Slartibartfass2: Looks better now. In general, I would extract the whole statement to the POM, e.g., `expectNumberOfEntries(4)`, but I keep that up to you.

tests/e2e/reading-list.test.ts
Slartibartfass2: I see. What are thoughts on adding a comment, explaining this?

src/routes/readinglist/+page.svelte
Slartibartfass2: I just saw that this doesn't make sense in this case.

src/lib/model/component-interfaces.ts
Slartibartfass2: Looks better now, imo, no unnecessary abstraction.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
Slartibartfass2: haha
In my opinion, it's weird that there's a distinction, as they represent the same thing.
I started a poll in our team channel.

Slartibartfass2 (changes requested):
Nearly done!

src/lib/components/composites/paper-components/ReadingListEntry.svelte
Merseleo: Ok, I think the opinion of our team is clear: uniformity.

tests/e2e/reading-list.test.ts
Merseleo: Ah ok, now I understand what you did mean. Yes, good idea, I will change it.

tests/e2e/reading-list.test.ts
Merseleo: Yes, I add it.

tests/e2e/reading-list.test.ts
Merseleo: I can.

src/lib/utils/common-helper.ts
Merseleo: But if you wrap these elements in `a` tags you have a `div` inside a `button` inside an `a` tag? Is that want to you want? And I think this shouldn't be added with this PR, but we can create an issue where we solve this problem. What is your opinion / suggestion?

Merseleo (commented):
Thank you for your thoughtful comments, I hope I have taken everything into account.

src/lib/utils/common-helper.ts
Slartibartfass2: No, what I mean is instead of a button, we use an `a` tag. But I agree, this shouldn't be part of this feature.

Slartibartfass2: No, what I mean is instead of a button, we use an `a` tag. But I agree, this shouldn't be part of this feature.

Slartibartfass2 (commented):
Let's discuss this last point and then merge it.

Edit: Also don't forget to rebase.

src/lib/utils/common-helper.ts
Merseleo: Ok, i will add this issue and I think I already found a good solution to allow a change of the top level html tag, so you easily decide whether to use a `button` or `a` tag via props (using `svelte:element`).

Slartibartfass2 (approved):

PR #306, opened 2025-04-10
Fix inconsistent paddings

### LINKED ISSUE #293: Bug: Inconsistent Paddings Between Main Page and Project Page
There seem to be inconsistent paddings, which lead to the `Open Reviews` header text jumping when directly switching between `/` and `/project/0` pages.

https://github.com/user-attachments/assets/a6751af9-8dc8-4449-9d3f-d7064633b5f9

mowi12:
Closes #293 

## What I have made

Removed vertical and horizontal from different components in order to make them consistent.
Add 'global' horizontal padding to keep all components spacing the same.

## Checklist

  - No source code edited

### DIFF
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -23,7 +23,7 @@
     }
 
     const {
-        listName,
+        listName = "",
         items,
         listItemComponent,
         listItemSkeleton,
@@ -71,7 +71,9 @@ This can be e.g. a search bar.
 -->
 <div class="flex h-full w-full flex-col gap-y-5 overflow-hidden">
     {#await items}
-        <h2>{listName}</h2>
+        {#if listName}
+            <h2>{listName}</h2>
+        {/if}
         {@render preListContent?.()}
         <ul class="scroll-box space-y-4 pb-1">
             {#each { length: numberOfSkeletons }}
@@ -81,10 +83,14 @@ This can be e.g. a search bar.
             {/each}
         </ul>
     {:then loadedItems}
-        {#if showNumberOfListItems}
-            <h2>{listName} ({numberOfItems === undefined ? loadedItems.length : numberOfItems})</h2>
-        {:else}
-            <h2>{listName}</h2>
+        {#if listName}
+            {#if showNumberOfListItems}
+                <h2>
+                    {listName} ({numberOfItems === undefined ? loadedItems.length : numberOfItems})
+                </h2>
+            {:else}
+                <h2>{listName}</h2>
+            {/if}
         {/if}
         {@render preListContent?.()}
         {#if loadedItems.length === 0}
@@ -100,7 +106,9 @@ This can be e.g. a search bar.
         {/if}
     {:catch error}
         {console.error(`Couldn't load items: ${error}`)}
-        <h2>{listName}</h2>
+        {#if listName}
+            <h2>{listName}</h2>
+        {/if}
         <ErrorIndicator errorMessage={errorHint ? errorHint : error} />
     {/await}
 </div>
--- a/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
@@ -131,7 +131,7 @@ Usage:
     <!-- TODO: Set `isBookmarkedDefault` as soon as endpoint is available -->
     <PaperBookmarkButton isBookmarkedDefault={false} {loadingPaperId} />
 </div>
-<main class="flex h-full w-full flex-col gap-5 px-2 py-4">
+<main class="flex h-full w-full flex-col gap-5 px-5">
     <div class="flex h-full w-full flex-row gap-5">
         <PaperDetailsCard {allowEditModeToggle} {loadingPaper} {startInEditMode} />
         <PaperResearchContextCard
--- a/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
@@ -131,7 +131,7 @@ Usage:
     <!-- TODO: Set `isBookmarkedDefault` as soon as endpoint is available -->
     <PaperBookmarkButton isBookmarkedDefault={false} {loadingPaperId} />
 </div>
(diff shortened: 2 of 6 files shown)


src/routes/+page.svelte
```diff
@@ -16,7 +16,7 @@
     <title>SnowballR</title>
 </svelte:head>
 <SimpleNavigationBar loadingTitle={Promise.resolve("SnowballR")} {user} />
-<main class="mb-10 flex h-full w-full flex-row gap-x-5 overflow-hidden">
+<main class="mb-10 flex h-full w-full flex-row gap-x-5 overflow-hidden px-5">
```
Merseleo:
```suggestion
<main class="mb-10 flex h-full w-full flex-row gap-x-15 overflow-hidden px-5">
```
I agree that it is smarter to handle the padding on the homepage and not within the `NamedList` component, however the gap between the two list on the homepage very small

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -120,7 +120,7 @@ Usage:
     <!-- TODO: Set `isBookmarkedDefault` as soon as endpoint is available -->
     <PaperBookmarkButton isBookmarkedDefault={false} {loadingPaperId} />
 </div>
-<main class="flex h-full w-full flex-col gap-5 px-2 py-4">
+<main class="flex h-full w-full flex-col gap-5 px-5">
```
Merseleo:
```suggestion
<main class="flex h-full w-full flex-col gap-5 px-5 pb-2">
```

I agree that the horizontal padding should be same and that the gap to the navigation bar should be consistent, but removing the vertical padding completely position the decision buttons and navigation buttons very close to the bottom margin. What do you think of keeping the bottom padding?

src/routes/project/[projectId]/dashboard/+page.svelte
```diff
@@ -22,8 +22,8 @@
     {/await}
 </svelte:head>
 <ProjectNavigationBar defaultTabValue="dashboard" {loadingProject} {projectId} {user} />
-<main class="mb-10 flex h-full w-full flex-row gap-x-5 overflow-hidden">
-    <section class="h-full w-full py-5">
+<main class="mb-10 flex h-full w-full flex-row gap-x-5 overflow-hidden px-5">
```
Comment (Merseleo) [inline src/routes/project/[projectId]/dashboard/+page.svelte]:
```suggestion
<main class="mb-10 flex h-full w-full flex-row gap-x-15 overflow-hidden px-5">
```

Merseleo (changes requested):
Not only fixed the one ‘problem’, but also corrected the "jumping" of headers in several pages - very attentive, good. 
But there's one thing you could perhaps do at the same time: In the settings, the menu on the left is also very indented, what do you think? Or should we leave it as it is for now or discuss it later?

PS: dont forget to rebase after the changes and re-review please if you do not apply the suggestions.

Slartibartfass2:
@mowi12 we keep this open until #310 is closed so that we see the effect.

src/routes/+page.svelte
mowi12: This is a good idea, but in my opinion 15 is too much. I think 10 would be a better fit.

```suggestion
<main class="mb-10 flex h-full w-full flex-row gap-x-10 overflow-hidden px-5">
```

mowi12: Yes, this is a good idea! But I think 15 is a bit much. Please take a look at the new changes, there i used 10 instead. I also added another gap in another file which we missed.

Merseleo: I agree, 10 is better.

Merseleo (dismissed):

Slartibartfass2:
Let's not wait for the visual regression tests, I expect that they will still take some time.

Merseleo:
@mowi12 do you want to rebase before I review it again? Otherwise I will review it tomorrow.

src/lib/components/composites/list/NamedList.svelte
```diff
@@ -71,7 +71,9 @@ This can be e.g. a search bar.
 -->
 <div class="flex h-full w-full flex-col gap-y-5 overflow-hidden">
     {#await items}
-        <h2>{listName}</h2>
+        {#if listName}
```
Merseleo, line 74: Why do you check for the `listname`? This should be always true, as you set a default for `listname`?

src/routes/project/[projectId]/dashboard/+page.svelte
```diff
@@ -22,8 +22,8 @@
     {/await}
 </svelte:head>
 <ProjectNavigationBar defaultTabValue="dashboard" {loadingProject} {projectId} {user} />
-<main class="mb-10 flex h-full w-full flex-row gap-x-5 overflow-hidden">
-    <section class="h-full w-full py-5">
+<main class="mb-10 flex h-full w-full flex-row gap-x-10 overflow-hidden px-5">
```
Comment (Merseleo) [inline src/routes/project/[projectId]/dashboard/+page.svelte:25]: Why only `gap-x-10` and not `gap-10`?

Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.svelte:25]: Same for reading list ...

Merseleo (changes requested):
I only have some questions, but maybe we already discussed this, this PR is already open for too long ^^.

src/lib/components/composites/list/NamedList.svelte
mowi12: This is true, but this also checks if the `listName` is not empty. If it is (e.g. on the reading list page), there is an extra gap between the not existing title and the search bar.
In my opinion, the gap should only exists between component that are visible.

src/routes/project/[projectId]/dashboard/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/dashboard/+page.svelte:25]: I actually did only updated the values from `5` to `10`. The `x` was already there. I just tested it, and noticed no difference, which makes sense.
If you want, we can remove the `x` part.

Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.svelte:25]: No, if it was `gap-x-5` before, I think we've given it some thought.

src/lib/components/composites/list/NamedList.svelte
Merseleo: Ah ok, I did not consider this extra gap, this makes sense.

Merseleo (approved):

Issue #10, opened 2024-10-28
[Use Case]: Toggle Shortcuts visibility


Slartibartfass2:
### Who is the primary actor in this use case?

User

### Describe the use case

As user, I want to toggle the visibility of shortcut hints in the application

### Acceptance criteria

- when the visibility is toggled on, then shortcut hints are shown in the application and vice versa

### Anything else?

Part of #8

PR #309, opened 2025-04-11
[Use Case]: Change project name

### LINKED ISSUE #18: [Use Case]: Change Project Name

Project Admin

### Describe the use case

As project admin, I want to change the name of the project.

### Acceptance criteria

- the new project name must be non-blank
- the maximum length of the project name is limited

### Anything else?

Part of #15

domienderle:
Closes #18 
## What I have made

Created class for project name settings. Added the labels, input field and button to rename the project name. Load the current project name on mount and save the project name if button is clicked. 

## Checklist

  - ~~[ ] unit tests~~ (no unit tests needed for the change of project name)
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
@@ -1,0 +1,138 @@
+<script lang="ts">
+    import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
+    import Input from "$lib/components/composites/input/Input.svelte";
+    import { Button } from "$lib/components/primitives/button";
+    import { backendService } from "$lib/grpc-api";
+    import { onMount } from "svelte";
+    import { Project } from "$lib/model/api/project";
+    import { Schema } from "$lib/schemas";
+    import { generateFieldMask } from "protobuf-fieldmask";
+    import type { ApiError } from "$lib/model/general";
+    import ErrorAlert from "$lib/components/composites/utils/ErrorAlert.svelte";
+    import { invalidate } from "$app/navigation";
+    import { LoaderCircle } from "lucide-svelte";
+    import { toast } from "svelte-sonner";
+
+    interface Props {
+        projectId: string;
+        loadingProject: Promise<Project>;
+    }
+
+    const { projectId, loadingProject }: Props = $props();
+
+    let projectName: string = $state("");
+
+    let projectNameInput: Input;
+
+    let updateProjectError: ApiError | undefined = $state(undefined);
+
+    let loading = $state(true);
+    let showLoadingSpinner = $state(false);
+
+    /**
+     * Loading the project name on mount.
+     */
+    onMount(async () => {
+        loadingProject
+            .then((project) => {
+                projectName = project.name;
+                loading = false;
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while loading the project name.",
+                };
+                console.error(`Couldn't load the project name: ${error}`);
+            });
+    });
+
+    /**
+     * Changes the name of the project, if a non-blank project name is provided.
+     *
+     * Therefore, send a request to the backend to update the project with the
+     * provided name.
+     */
+    async function handleSubmit(event: Event) {
+        updateProjectError = undefined;
+        loading = showLoadingSpinner = true;
+        event.preventDefault();
+        const newProjectName: string = projectNameInput.getValue().trim();
+        const projectNameValid = projectNameInput.validate();
+        if (projectNameValid) {
+            const projectData: Partial<Project> = {
+                id: projectId,
+                name: newProjectName,
+            };
+            const fieldMask = generateFieldMask(projectData);
+            if (projectName === newProjectName) {
+                toast("Please enter a new project name.");
+            } else {
+                try {
+                    await backendService.updateProject({
+                        project: Project.create(projectData),
+                        mask: {
+                            paths: fieldMask,
+                        },
+                    }).response;
+                    projectName = newProjectName;
+                    toast("Successfully updated project name.");
+                    await invalidate("data:getProjectById");
+                } catch (error) {
+                    updateProjectError = {
+                        errorTitle: "Something went wrong while updating the project name.",
+                    };
+                    console.error(`Couldn't update project: ${error}`);
+                }
+            }
+        }
+        loading = showLoadingSpinner = false;
+    }
+</script>
+
+<!--
+@component
+Component for changing the project name settings.
+
+Usage:
+```svelte
+    <ProjectNameSettings {loadingProject} {projectId}/>
+```
+-->
+<SettingsSection sectionTitle="General">
+    <form
+        id="project-update"
+        class="flex w-full max-w-xl flex-col items-center gap-2.5 md:h-fit md:max-w-xl md:flex-row md:items-start"
+        onsubmit={handleSubmit}
+    >
+        <Input
+            bind:this={projectNameInput}
+            class="h-full w-full"
+            disabled={loading}
+            inputId="project-name-input"
+            label="Project Name"
+            placeholder={loading ? "Loading" : "New Project Name"}
+            required
+            schema={Schema.projectName}
+            type="text"
+            value={projectName}
+        />
+        <Button
+            class="text-md w-full md:mt-5.5 md:w-42"
+            disabled={loading}
+            form="project-update"
+            type="submit"
+        >
+            {#if showLoadingSpinner}
+                <LoaderCircle class="animate-spin" />
+                Renaming
+            {:else}
+                Rename
+            {/if}
+        </Button>
+    </form>
+    {#if updateProjectError}
+        <div class="max-w-xl">
+            <ErrorAlert errorTitle={updateProjectError.errorTitle} />
+        </div>
+    {/if}
+</SettingsSection>
--- a/tests/e2e/fixtures/project-settings-fixture.ts
+++ b/tests/e2e/fixtures/project-settings-fixture.ts
@@ -1,5 +1,7 @@
 import { DevProjectSettingsPage } from "$tests/e2e/pom/project-settings-model";
 import { test as base } from "./general-fixture";
+import { expect } from "@playwright/test";
+import { projectId } from "../project-name-settings.test";
 
 type ProjectSettingsPage = {
     projectSettingsPage: DevProjectSettingsPage;
@@ -11,6 +13,8 @@ type ProjectSettingsPage = {
  */
 export const test = base.extend<ProjectSettingsPage>({
     projectSettingsPage: async ({ page }, use) => {
+        await page.goto(`/project/${projectId}/settings/general`);
+        await expect(page.getByRole("heading", { name: "Project 1" })).toBeVisible();
         await use(new DevProjectSettingsPage(page));
     },
 });
--- a/tests/e2e/fixtures/project-settings-fixture.ts
+++ b/tests/e2e/fixtures/project-settings-fixture.ts
@@ -1,0 +1,16 @@
+import { DevProjectSettingsPage } from "$tests/e2e/pom/project-settings-model";
+import { test as base } from "./general-fixture";
+
+type ProjectSettingsPage = {
+    projectSettingsPage: DevProjectSettingsPage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - project settings page
+ */
+export const test = base.extend<ProjectSettingsPage>({
+    projectSettingsPage: async ({ page }, use) => {
+        await use(new DevProjectSettingsPage(page));
+    },
+});
--- a/tests/e2e/pom/project-settings-model.ts
+++ b/tests/e2e/pom/project-settings-model.ts
@@ -1,0 +1,27 @@
+import { expect, type Locator, type Page } from "@playwright/test";
+
+export class DevProjectSettingsPage {
+    readonly page: Page;
+    readonly projectNameInput: Locator;
+    readonly renameButton: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.projectNameInput = page.getByLabel("Project Name");
+        this.renameButton = page.getByRole("button", { name: "Rename" });
+    }
+
+    /**
+     * Changes the name of the project.
+     *
+     * @param projectName - the new project name
+     */
+    async changeProjectName(projectName: string) {
+        await this.projectNameInput.fill(projectName);
+        await this.renameButton.click();
+    }
+
+    async checkForErrors() {
+        await expect(this.page.getByRole("alert")).not.toBeVisible();
+    }
+}
--- a/tests/e2e/project-name-settings.test.ts
+++ b/tests/e2e/project-name-settings.test.ts
@@ -1,0 +1,57 @@
+import { test } from "./fixtures/project-settings-fixture";
+import { expect } from "@playwright/test";
+
+export let projectId: string = "";
+export const projectReviewSettingsProjectName = "Project 1";
+
+test.describe("Renaming a project", () => {
+    /**
+     * Creates a new project before all tests in this test suite.
+     */
+    test.beforeAll(async ({ mockBackendService }) => {
+        await mockBackendService
+            .createProject({ name: projectReviewSettingsProjectName })
+            .then((project) => (projectId = project.response.id));
+    });
+
+    test.afterAll(async ({ mockBackendService }) => {
+        mockBackendService.softDeleteProject({ id: projectId });
+    });
+
+    test("When the user enters an invalid project name, then the name of the project remains unchanged.", async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("New Project");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByRole("heading", { name: "New Project" });
+        await expect(projectNameHeader).toBeVisible();
+        const toast = page.getByText("Please enter a new project name.");
+        await expect(toast).toBeVisible();
+    });
+
+    test("When the user enters an empty project name, then the name of the project remains unchanged.", async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName(" ");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByRole("heading", { name: "Project 1" });
+        await expect(projectNameHeader).toBeVisible();
+        // Here is no toast necessary, because it is handled by the input field.
+    });
+
+    test("When the user enters a valid project name, then the name of the project should be updated to this name.", async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("New Project");
+        await projectSettingsPage.checkForErrors();
+        const oldHeader = page.getByRole("heading", { name: "Project 1" });
+        await expect(oldHeader).not.toBeVisible();
+        const projectNameHeader = page.getByRole("heading", { name: "New Project" });
+        await expect(projectNameHeader).toBeVisible();
+        const toast = page.getByText("Successfully updated project name.");
+        await expect(toast).toBeVisible();
+    });
+});
--- a/tests/e2e/project-name-settings.test.ts
+++ b/tests/e2e/project-name-settings.test.ts
@@ -2,7 +2,7 @@ import { test } from "./fixtures/project-settings-fixture";
 import { expect } from "@playwright/test";
 
 export let projectId: string = "";
-export const projectReviewSettingsProjectName = "Project 1";
+const projectName = "Project 1";
 
 test.describe("Renaming a project", () => {
     /**
@@ -10,7 +10,7 @@ test.describe("Renaming a project", () => {
      */
     test.beforeAll(async ({ mockBackendService }) => {
         await mockBackendService
-            .createProject({ name: projectReviewSettingsProjectName })
+            .createProject({ name: projectName })
             .then((project) => (projectId = project.response.id));
     });
 
@@ -22,9 +22,9 @@ test.describe("Renaming a project", () => {
         page,
         projectSettingsPage,
     }) => {
-        await projectSettingsPage.changeProjectName("New Project");
+        await projectSettingsPage.changeProjectName("Project 1");
         await projectSettingsPage.checkForErrors();
-        const projectNameHeader = page.getByRole("heading", { name: "New Project" });
+        const projectNameHeader = page.getByRole("heading", { name: "Project 1" });
         await expect(projectNameHeader).toBeVisible();
         const toast = page.getByText("Please enter a new project name.");
         await expect(toast).toBeVisible();
@@ -38,7 +38,8 @@ test.describe("Renaming a project", () => {
         await projectSettingsPage.checkForErrors();
         const projectNameHeader = page.getByRole("heading", { name: "Project 1" });
         await expect(projectNameHeader).toBeVisible();
-        // Here is no toast necessary, because it is handled by the input field.
+        const toast = page.getByText("Please provide a non-empty project name");
+        await expect(toast).toBeVisible();
     });
 
     test("When the user enters a valid project name, then the name of the project should be updated to this name.", async ({
--- a/tests/integration/settings/project-settings/general/project-name-settings.test.ts
+++ b/tests/integration/settings/project-settings/general/project-name-settings.test.ts
@@ -1,0 +1,129 @@
+import { describe, expect, test } from "vitest";
+import { render, screen, waitFor } from "@testing-library/svelte";
+import ProjectNameSettings from "$lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte";
+import { Projects } from "$tests/example-data";
+import userEvent from "@testing-library/user-event";
+import { ProjectStatus } from "$lib/model/api/project";
+import { mockApiCall, mockFailedApiCall } from "$tests/setupTest";
+
+describe("ProjectNameSettings", () => {
+    test("When all props are provided, then it renders correctly, with two labels, one input field and a button", async () => {
+        render(ProjectNameSettings, {
+            target: document.body,
+            props: {
+                projectId: Projects.demoProject.id,
+                loadingProject: Promise.resolve(Projects.demoProject),
+            },
+        });
+
+        expect(screen.queryByText("General")).toBeInTheDocument();
+        expect(screen.queryByText("Project Name")).toBeInTheDocument();
+        const projectRenameInput = screen.getByLabelText("Project Name");
+        expect(projectRenameInput).toBeInTheDocument();
+        await waitFor(() => {
+            expect(projectRenameInput).toHaveValue(Projects.demoProject.name);
+        });
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+        expect(renameButton).toBeInTheDocument();
+    });
+
+    test("When the input field is empty or the project name is the same, then the button should not change the name", async () => {
+        const mockCall = mockApiCall("getProjectById", {
+            id: "",
+            name: "",
+            status: ProjectStatus.UNSPECIFIED,
+            currentStage: 0n,
+            maxStage: 0n,
+        });
+        const mockUpdateProject = mockApiCall("updateProject", {
+            id: "0",
+            name: "New Project Name",
+            status: ProjectStatus.ACTIVE,
+            currentStage: 0n,
+            maxStage: 1n,
+            settings: undefined,
+        });
+
+        render(ProjectNameSettings, {
+            target: document.body,
+            props: {
+                projectId: Projects.demoProject.id,
+                loadingProject: Promise.resolve(Projects.demoProject),
+            },
+        });
+
+        const projectRenameInput = screen.getByLabelText("Project Name");
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, " ");
+        await userEvent.click(renameButton);
+        expect(mockUpdateProject).toHaveBeenCalled();
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, "Demo Project");
+        await userEvent.click(renameButton);
+        expect(mockUpdateProject).toHaveBeenCalled();
+        expect(mockCall).not.toHaveBeenCalled();
+    });
+
+    test("When the input field is filled with a differing name, then the button should change the name", async () => {
+        const mockCall = mockApiCall("updateProject", {
+            id: "0",
+            name: "New Project Name",
+            status: ProjectStatus.ACTIVE,
+            currentStage: 0n,
+            maxStage: 1n,
+            settings: undefined,
+        });
+
+        render(ProjectNameSettings, {
+            props: {
+                projectId: Projects.demoProject.id,
+                loadingProject: Promise.resolve(Projects.demoProject),
+            },
+        });
+
+        const projectRenameInput = screen.getByLabelText("Project Name");
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, "New Project Name");
+        await userEvent.click(renameButton);
+        expect(mockCall).toHaveBeenCalled();
+    });
+
+    test("When a failed API call to get the project is made, then an error message should be shown", async () => {
+        const failedPromise = Promise.reject(new Error("Error while loading project"));
+        render(ProjectNameSettings, {
+            props: {
+                projectId: Projects.demoProject.id,
+                loadingProject: failedPromise,
+            },
+        });
+        const errorHeading = await screen.findByRole("heading", {
+            name: "Something went wrong while loading the project name.",
+        });
+        expect(errorHeading).toBeInTheDocument();
+    });
+
+    test("When a failed API call to update the project is made, then an error message should be shown", async () => {
+        mockFailedApiCall("updateProject");
+        render(ProjectNameSettings, {
+            props: {
+                projectId: Projects.demoProject.id,
+                loadingProject: Promise.resolve(Projects.demoProject),
+            },
+        });
+        const projectRenameInput = screen.getByLabelText("Project Name");
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, "New Project Name");
+        await userEvent.click(renameButton);
+        expect(
+            screen.getByRole("heading", {
+                name: "Something went wrong while updating the project name.",
+            }),
+        ).toBeInTheDocument();
+    });
+});
--- a/tests/integration/settings/project-settings/general/project-name-settings.test.ts
+++ b/tests/integration/settings/project-settings/general/project-name-settings.test.ts
@@ -3,8 +3,8 @@ import { render, screen, waitFor } from "@testing-library/svelte";
 import ProjectNameSettings from "$lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte";
 import { Projects } from "$tests/example-data";
 import userEvent from "@testing-library/user-event";
-import { ProjectStatus } from "$lib/model/api/project";
 import { mockApiCall, mockFailedApiCall } from "$tests/setupTest";
+import { createProject } from "$tests/model-builder";
 
 describe("ProjectNameSettings", () => {
     test("When all props are provided, then it renders correctly, with two labels, one input field and a button", async () => {
@@ -28,21 +28,10 @@ describe("ProjectNameSettings", () => {
     });
 
     test("When the input field is empty or the project name is the same, then the button should not change the name", async () => {
-        const mockCall = mockApiCall("getProjectById", {
-            id: "",
-            name: "",
-            status: ProjectStatus.UNSPECIFIED,
-            currentStage: 0n,
-            maxStage: 0n,
-        });
-        const mockUpdateProject = mockApiCall("updateProject", {
-            id: "0",
-            name: "New Project Name",
-            status: ProjectStatus.ACTIVE,
-            currentStage: 0n,
-            maxStage: 1n,
-            settings: undefined,
-        });
+        const mockUpdateProject = mockApiCall(
+            "updateProject",
+            createProject({ name: "New Project Name" }),
+        );
 
         render(ProjectNameSettings, {
             target: document.body,
@@ -62,19 +51,11 @@ describe("ProjectNameSettings", () => {
         await userEvent.clear(projectRenameInput);
         await userEvent.type(projectRenameInput, "Demo Project");
         await userEvent.click(renameButton);
-        expect(mockUpdateProject).toHaveBeenCalled();
-        expect(mockCall).not.toHaveBeenCalled();
+        expect(mockUpdateProject).not.toHaveBeenCalled();
     });
 
     test("When the input field is filled with a differing name, then the button should change the name", async () => {
-        const mockCall = mockApiCall("updateProject", {
-            id: "0",
-            name: "New Project Name",
-            status: ProjectStatus.ACTIVE,
-            currentStage: 0n,
-            maxStage: 1n,
-            settings: undefined,
-        });
+        const mockCall = mockApiCall("updateProject", createProject({ name: "New Project Name" }));
 
         render(ProjectNameSettings, {
             props: {
--- a/tests/integration/settings/project-settings/general/project-name-settings.test.ts
+++ b/tests/integration/settings/project-settings/general/project-name-settings.test.ts
@@ -43,11 +43,12 @@ describe("ProjectNameSettings", () => {
 
         const projectRenameInput = screen.getByLabelText("Project Name");
         const renameButton = screen.getByRole("button", { name: "Rename" });
-
+        await waitFor(() => expect(renameButton).toBeEnabled());
         await userEvent.clear(projectRenameInput);
         await userEvent.type(projectRenameInput, " ");
         await userEvent.click(renameButton);
-        expect(mockUpdateProject).toHaveBeenCalled();
+        expect(mockUpdateProject).not.toHaveBeenCalled();
+        await waitFor(() => expect(renameButton).toBeEnabled());
         await userEvent.clear(projectRenameInput);
         await userEvent.type(projectRenameInput, "Demo Project");
         await userEvent.click(renameButton);
@@ -66,7 +67,7 @@ describe("ProjectNameSettings", () => {
 
         const projectRenameInput = screen.getByLabelText("Project Name");
         const renameButton = screen.getByRole("button", { name: "Rename" });
-
+        await waitFor(() => expect(renameButton).toBeEnabled());
         await userEvent.clear(projectRenameInput);
         await userEvent.type(projectRenameInput, "New Project Name");
         await userEvent.click(renameButton);
@@ -97,7 +98,7 @@ describe("ProjectNameSettings", () => {
         });
         const projectRenameInput = screen.getByLabelText("Project Name");
         const renameButton = screen.getByRole("button", { name: "Rename" });
-
+        await waitFor(() => expect(renameButton).toBeEnabled());
         await userEvent.clear(projectRenameInput);
         await userEvent.type(projectRenameInput, "New Project Name");
         await userEvent.click(renameButton);
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
--- a/src/lib/schemas.ts
+++ b/src/lib/schemas.ts
@@ -120,7 +120,7 @@ const passwordSchema = z
 const projectNameSchema = z
     .string()
     .trim()
-    .min(1, { message: "Please provide a project name" })
+    .min(1, { message: "Please provide a non-empty project name" })
     .max(100, { message: "The project name is limited to 100 characters" });
 
 export const Schema = {
--- a/src/routes/project/[projectId]/+layout.ts
+++ b/src/routes/project/[projectId]/+layout.ts
@@ -1,10 +1,10 @@
 import { backendService } from "$lib/grpc-api";
 import type { LayoutLoad } from "./$types";
 
-export const load: LayoutLoad = async ({ params }) => {
+export const load: LayoutLoad = async ({ params, depends }) => {
+    depends("data:getProjectById");
     const projectId = params.projectId;
     const loadingProject = backendService.getProjectById({ id: projectId }).response;
-
     // attach noop-catch to handle promise rejection correctly (see https://svelte.dev/docs/kit/load#Streaming-with-promises)
     loadingProject.catch(() => {});
 
--- a/src/routes/project/[projectId]/settings/+layout.svelte
+++ b/src/routes/project/[projectId]/settings/+layout.svelte
@@ -2,7 +2,7 @@
     import ProjectNavigationBar from "$lib/components/composites/navigation-bar/ProjectNavigationBar.svelte";
 
     const { children, data } = $props();
-    const { user, projectId, loadingProject } = data;
+    const { user, projectId, loadingProject } = $derived(data);
 </script>
 
 <ProjectNavigationBar defaultTabValue="settings" {loadingProject} {projectId} {user} />
--- a/src/routes/project/[projectId]/settings/general/+page.svelte
+++ b/src/routes/project/[projectId]/settings/general/+page.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
     import ProjectSettingsLayout from "$lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte";
+    import ProjectNameSettings from "$lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte";
 
-    let { data } = $props();
-    const { projectId, loadingProject } = data;
+    const { data } = $props();
+    const { projectId, loadingProject } = $derived(data);
 </script>
 
 <svelte:head>
@@ -14,5 +15,8 @@
         <title>General | Settings</title>
     {/await}
 </svelte:head>
-
-<ProjectSettingsLayout {projectId} selectedTab="general">General content</ProjectSettingsLayout>
+<ProjectSettingsLayout {projectId} selectedTab="general">
+    <div class="flex flex-col gap-9 overflow-auto p-2.5">
+        <ProjectNameSettings {loadingProject} {projectId} />
+    </div>
+</ProjectSettingsLayout>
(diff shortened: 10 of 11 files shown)


src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+        const projectNameValid = projectNameInput.validate();
+        if (!(projectNameValid)) return;
+
+        const projectData: Partial<Project> = {
+            id: projectId,
+            name: projectNameInput.getValue(),
+        };
+
+        const fieldMask = generateFieldMask(projectData);
+        console.log("fieldMask", fieldMask);
```
Slartibartfass2: Do we need this logging statement here?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+            .then(() => {return})
+            .catch((error) => {
+                console.log(error.code)
+                if (error.code === StatusCodes.ALREADY_EXISTS) {
+                    updateProjectError = {
+                        errorTitle: "TODO",
+                        errorDetails: "TODO",
+                    }
+                } else {
+                    updateProjectError = { errorTitle: "Somethin went wrong while updating project name."};
```
Slartibartfass2: typo
```suggestion
                    updateProjectError = { errorTitle: "Something went wrong while updating project name."};
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+
+</script>
+
+<!--
+@component
+Component for changing the project name settings.
+
+Usage:
+```svelte
+    <SettingsSection />
```
Slartibartfass2: Provide a correct example how this component can be used.
```suggestion
    <ProjectNameSettings {loadingProject} {projectId}/>
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+            })
+            .then(() => {return})
+            .catch((error) => {
+                console.log(error.code)
+                if (error.code === StatusCodes.ALREADY_EXISTS) {
+                    updateProjectError = {
+                        errorTitle: "TODO",
+                        errorDetails: "TODO",
+                    }
+                } else {
```
Slartibartfass2: Why is the error code logged, and in which case would we receive the `ALREADY_EXISTS` error?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+```svelte
+    <SettingsSection />
+```
+-->
+<SettingsSection sectionTitle="General">
+    <div class="items-top flex flex-row space-x-2">
+        <div class="grid gap-1.5 pt-1 leading-none">
+            <form
+                id="project-update"
+                class="flex h-fit w-full max-w-175 flex-col items-center gap-2.5 md:flex-row md:items-end"
```
Slartibartfass2: I think this could be a bit wider. There's enough space, and project names are probably quite long.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+        console.log("fieldMask", fieldMask);
+
+        backendService
+            .updateProject({
+                project: Project.create(projectData),
+                mask: {
+                    paths: fieldMask,
+                },
+            })
+            .then(() => {return})
```
Slartibartfass2: After successfully updating the project name, we should update the data so that the changes are rendered.
Use `invalidate` and `depends` to invalidate the `getProjectById` request (see https://svelte.dev/tutorial/kit/invalidation).
@Merseleo already used this on the home page.

Slartibartfass2 (approved):

Slartibartfass2 (changes requested):
Misclicked...

There are still some things to be done and fix the linting issues.

Slartibartfass2 (changes requested):
Please first answer the conversations we already had.

Note: currently the CI fails, check the logs to see which parts raise errors.

.github/workflows/lighthouse.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 env:
     node_version: ${{ vars.NODE_VERSION }}
+    PUBLIC_API_BASE_URL: ${{ vars.PUBLIC_API_BASE_URL }}
```
Slartibartfass2: Why is this change required? Did this job fail without the env variable?

package.json
```diff
@@ -81,5 +82,9 @@
             "name": "node",
             "version": ">=22.0.0 <=23.11.0"
         }
+    },
+    "dependencies": {
+        "@rollup/rollup-linux-x64-gnu": "^4.40.0",
+        "rollup": "^4.40.0"
     }
```
Slartibartfass2: Why do we require this dependency? Where is it used?

package.json
```diff
@@ -59,9 +59,10 @@
         "markdownlint-cli": "^0.44.0",
         "mode-watcher": "^0.5.0",
         "postcss": "^8.5.3",
-        "prettier": "^3.5.2",
+        "prettier": "^3.5.3",
```
Slartibartfass2: Was this change required in any way?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,109 @@
...
+        loadingProject: Promise<Project>;
+    }
+
+    const { projectId, loadingProject }: Props = $props();
+
+    let projectName: string = $state("");
+
+    let projectNameInput: Input;
+
+    const updateProjectError: ApiError | undefined = $state(undefined);
```
Slartibartfass2: This error variable is never set. How about using it in the catch clause of the `updateProject` call?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,109 @@
...
+```svelte
+    <ProjectNameSettings {loadingProject} {projectId}/>
+```
+-->
+<SettingsSection sectionTitle="General">
+    <div class="items-top flex flex-row space-x-2">
+        <div class="grid gap-1.5 pt-1 leading-none">
+            <form
+                id="project-update"
+                class="flex h-fit w-xl max-w-175 flex-col items-center gap-2.5 md:flex-row md:items-end"
```
Slartibartfass2: At ~825px the button is cut off. Let's change this so that the overflow works correctly, without introducing horizontal scroll bars.
image

tests/e2e/pom/project-settings-model.ts
```diff
@@ -0,0 +1,23 @@
+import type { Page } from "@playwright/test";
+
+export class DevProjectSettingsPage {
+    readonly page: Page;
+    readonly projectNameInput;
+    readonly renameButton;
```
Slartibartfass2: Annotate the type so that it is clear what the types are.
```suggestion
    readonly projectNameInput: Locator;
    readonly renameButton: Locator;
```

tests/e2e/pom/project-settings-model.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    constructor(page: Page) {
+        this.page = page;
+        this.projectNameInput = page.getByTestId("project-rename-input");
```
Slartibartfass2: In general, try to use locators that are available to the user, such as text, role, or label.
In this case, you can use `getByLabel("Project Name")`.

tests/e2e/pom/home-page-model.ts
```diff
@@ -18,4 +19,31 @@ export class DevHomePage {
...
+        const firstProjectLink = this.page
+            .locator('a[href^="/project/"][href$="/dashboard"]')
+            .first();
+        await expect(firstProjectLink).toBeVisible();
+        await firstProjectLink.click();
+
+        // Navigate to the general settings page of the newly created project
+        const settingsLink = this.page
+            .locator('a[href^="/project/"][href$="/settings/general"]')
+            .first();
```
Slartibartfass2: Use locators that are accessible to the user. The project has a name that can be found on the page and the settings tab has also a constant name.

tests/e2e/pom/home-page-model.ts
```diff
@@ -18,4 +19,31 @@ export class DevHomePage {
         await expect(this.createProjectDialog).not.toBeVisible();
         await this.openCreateProjectDialogButton.click();
     }
+
+    /**
+     * Creates a new Project. After that navigates to the projects general settings page.
+     */
+    async openProjectSettings() {
```
Slartibartfass2: It feels a bit wrong to me that this is in the POM of the home page. Do you think there could be a better place where we could put this?

src/routes/+page.svelte
```diff
@@ -34,7 +34,7 @@
             {/snippet}
         </NamedList>
     </section>
-    <section class="flex h-full w-full min-w-0 flex-col gap-y-5">
+    <section class="flex h-full w-full min-w-0 flex-col gap-y-5" data-testid="named-list-projects">
```
Slartibartfass2: As far as I can tell, this is never used again. Why do we need it?

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    test.beforeEach(async ({ page }) => {
+        await page.goto("/");
+    });
+
+    test('When the user enters an valid project name, the new project name should be updated to "New Project".', async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("New Project");
+        const projectNameHeader = page.getByTestId("project-name-header");
```
Slartibartfass2: As mentioned in `project-settings-model.ts`, don't use the test ID locator if not necessary.
Here could just test if there's a heading if the text "New Project". Same below.

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,32 @@
...
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("Test Project");
+        const projectNameHeader = page.getByTestId("project-name-header");
+        await expect(projectNameHeader).toBeVisible();
+        await expect(projectNameHeader).toHaveText("Test Project");
+
+        await projectSettingsPage.changeProjectName("");
+        await expect(projectNameHeader).toBeVisible();
+        await expect(projectNameHeader).toHaveText("Test Project");
```
Slartibartfass2: Also check if an error message is displayed. See project creation for reference.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,109 @@
...
+                    bind:this={projectNameInput}
+                    class="h-fit w-full"
+                    data-testid="project-rename-input"
+                    inputId="project-name-input"
+                    label="Project Name"
+                    required
+                    schema={Schema.projectName}
+                    type="text"
+                    value={projectName}
+                />
```
Slartibartfass2: Let's also provide a placeholder, such as "Project Name"

tests/integration/settings/project-settings/general/project-name-settings.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                loadingProject: Promise.resolve(Projects.demoProject),
+            },
+        });
+
+        const projectRenameInput = screen.getByTestId("project-rename-input");
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, " ");
+        await userEvent.clear(projectRenameInput);
```
Slartibartfass2: Here we clear, fill, and clear the input again. Are these statements necessary if we don't check what's happening and clear the input either way?

tests/integration/settings/project-settings/general/project-name-settings.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                response: Promise.resolve({
+                    id: "0",
+                    name: "New Project Name",
+                    status: ProjectStatus.ACTIVE,
+                    currentStage: 0n,
+                    maxStage: 1n,
+                    settings: undefined,
+                } as Project),
+            } as ReturnType<(typeof backendService)["updateProject"]>;
+        });
```
Slartibartfass2: Is there a reason why we don't use the `mockApiCall` helper here? And is it even necessary to do so?

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -28,6 +28,7 @@ Usage:
     class="border-container-border-grey highlight-on-hover flex h-fit w-full flex-col justify-between gap-2 rounded-md border px-5
     py-2 lg:flex-row lg:items-center lg:gap-10"
     class:opacity-25={project.status === ProjectStatus.ARCHIVED}
+    data-testid={`project-entry-${project.id}`}
```
Slartibartfass2: This isn't used everywhere? Do we still need it?

Slartibartfass2 (changes requested):
Great work! The feature works and only some minor things need to be further discussed. As already mentioned, earlier, please answer all my comments, don't just implement what I'm suggesting. In general, we want to have a discussion about what's best for the application. As we didn't explain this earlier, I have now added more information about the reviewing process in the wiki. You can find the draft [here](https://github.com/SE-UUlm/snowballr-frontend/blob/docs/321-feature-add-more-contribution-guidelines/wiki/Contributing.md).

Lastly, there appears to be some issues with the line endings in the last two commits. I recommend resetting them with `--mixed` and ensure that git automatically changes the line endings. As an alternative, you can add a file `.gitattributes` and add the following content:
```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto
```

tests/e2e/pom/home-page-model.ts
Slartibartfass2: On a second thought, this seems okay to me.

.github/workflows/lighthouse.yml
domienderle: No, I just tried something and forgot to remove this line.

package.json
domienderle: This line was automatically added during the WSL setup process, but yo u are right, it is not necessary.

package.json
domienderle: No it is not. Prettier works without any further problems in the old version as well. It was also a product of my desperate try to setup the project in WSL.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: That is a valid point. I forgot to include the error in the catch statement. I will include it.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Didn't recognize that. I will change it.

tests/e2e/project-name-settings.test.ts
domienderle: Good point. Will be implemented.

tests/e2e/pom/project-settings-model.ts
domienderle: I didn't know that it is good practice to do it like that. In case I will adapt this in the future.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
domienderle: Yes it is necessary. First we have to clear the input field, which is filled with "Demo Project". Then we type in " " and click the rename button. After this with have to clear the field again to type in "Demi Project" and cklick the rename button. In both cases no mock-call should be performed, which is why it is not checked in between.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
domienderle: No. I just didn't recognize this method existed.

src/lib/components/composites/project-components/ProjectListEntry.svelte
domienderle: No. I forgot to delete it.

src/routes/+page.svelte
domienderle: You are right. It is not needed anymore.

tests/e2e/pom/home-page-model.ts
domienderle: Ok, good to know :)

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Good idea.

.gitattribute
Slartibartfass2, line 1: This file is unused. (and if you want to use it for git it should be `.gitattributes`, with an `s`)

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+        loadingProject
+            .then((project) => {
+                projectName = project.name;
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while loading the project name on mount.",
+                    errorDetails: error,
+                };
+                console.error(updateProjectError);
```
Slartibartfass2: Let's change this to something more descriptive.
```suggestion
                console.error(`Couldn't load the project name: ${error}`);
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+            })
+            ?.response.then(() => {
+                invalidate("data:getProjectById");
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while updating the project name.",
+                    errorDetails: error,
+                };
+                console.error(updateProjectError);
```
Slartibartfass2: Let's change this to something more descriptive.
```suggestion
                console.error(`Couldn't update project: ${error}`);
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+        const fieldMask = generateFieldMask(projectData);
+
+        backendService
+            .updateProject({
+                project: Project.create(projectData),
+                mask: {
+                    paths: fieldMask,
+                },
+            })
+            ?.response.then(() => {
```
Slartibartfass2: Is the `?` required? Did you experience errors without it?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+     * Loading the project name on mount.
+     */
+    onMount(async () => {
+        loadingProject
+            .then((project) => {
+                projectName = project.name;
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while loading the project name on mount.",
```
Slartibartfass2: The user doesn't know what on mount is and they don't need to.
```suggestion
                    errorTitle: "Something went wrong while loading the project name.",
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+                        type="text"
+                        value={projectName}
+                    />
+                </div>
+                <Button class="h-12.5 w-25 flex-shrink-0" form="project-update" type="submit">
+                    Rename
+                </Button>
+            </form>
+            {#if updateProjectError}
+                <ErrorAlert errorTitle="Something went wrong while updating project name." />
```
Slartibartfass2:
```suggestion
                <ErrorAlert errorTitle="Something went wrong while updating the project name." />
```

Slartibartfass2: Also, let's cap the width with the form width. Otherwise, the alert is wider than the form in some cases:
image

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,121 @@
...
+```svelte
+    <ProjectNameSettings {loadingProject} {projectId}/>
+```
+-->
+<SettingsSection sectionTitle="General">
+    <div class="items-top space-x-2">
+        <div class="grid gap-1.5 pt-1 leading-none">
+            <form
+                id="project-update"
+                class="flex max-w-xl flex-nowrap items-end gap-2.5"
```
Slartibartfass2: @mowi12 had the same problem with `items-end` that this is fine until the input displays an error. We fixed this by using a top margin for the button instead.
image

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,33 @@
+import { test } from "./fixtures/project-settings-fixture";
+import { expect } from "@playwright/test";
+
+test.describe("Renaming a project", () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto("/");
```
Slartibartfass2: Let's move this to the project settings page fixture.

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    test('When the user enters an valid project name, the new project name should be updated to "New Project".', async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("New Project");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByText("New Project");
```
Slartibartfass2:
```suggestion
        const projectNameHeader = page.getByRole("heading", { name: "New Project" });
```

tests/integration/settings/project-settings/general/project-name-settings.test.ts
Slartibartfass2: But you don't click the button after typing " ".

tests/integration/settings/project-settings/general/project-name-settings.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        const projectRenameInput = screen.getByLabelText("Project Name");
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.clear(projectRenameInput);
+        await userEvent.type(projectRenameInput, "New Project Name");
+        await userEvent.click(renameButton);
+        expect(mockCall).toHaveBeenCalled();
+    });
+});
```
Slartibartfass2, line 111: Please add two other test cases:
- one with `loadingProject` is promise that rejects to test whether the error alert is shown
- one with `updateProject` failing to test whether the error alert is shown (use `mockFailedApiCall` for that)

Slartibartfass2 (changes requested):
We are nearly done here.
There's still the problem with the wrong line endings in many files. Let's fix this too.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Yes, sometimes I get an error if I run the integration tests. But no test fail, its just like an annotation, but the project-name-settings.test.ts is then marked as failed. But only sometimes. Mostly it runs without any problems.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: All right. I removed unnecessary divs and adjusted it similar to @mowi12. Should be fixed now

tests/e2e/project-name-settings.test.ts
domienderle: True. In the beginning I did not really know what the fixture was about. But makes sense

tests/integration/settings/project-settings/general/project-name-settings.test.ts
domienderle: Good point. Tests are added

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: This seems to origin from the test "When the input field is empty or the project name is the same, then the button should not change the name". You can fix this, when also mocking the "updateProject" method:
```ts
mockApiCall("updateProject", Project.create());
```

Slartibartfass2: In general, you should mock all methods that are called during the test

tests/integration/settings/project-settings/general/project-name-settings.test.ts
```diff
@@ -0,0 +1,118 @@
...
+        await userEvent.type(projectRenameInput, "New Project Name");
+        await userEvent.click(renameButton);
+        expect(mockCall).toHaveBeenCalled();
+    });
+
+    test("When a failed API call to get the project is made, then an error message should be shown", async () => {
+        const failedPromise = new Promise<Project>((_, reject) => {
+            setTimeout(() => reject("Error while loading project"), 0);
+        });
+        vi.spyOn(console, "error").mockImplementation(() => {});
```
Slartibartfass2: Is there a reason why we use a timeout with 0 ms? These lines can be simplified. Also, let's not mock the console logging.
```suggestion
        const failedPromise = Promise.reject(new Error("Error while loading project"));
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,120 @@
...
+     */
+    onMount(async () => {
+        loadingProject
+            .then((project) => {
+                projectName = project.name;
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while loading the project name.",
+                    errorDetails: error,
```
Slartibartfass2: Let's not display the error object to the user. People without knowledge of computer science will probably get more confused by this.
We do the same thing for the update project call.

Furthermore, `<ErrorAlert errorTitle={updateProjectError.errorTitle} />` is fine as alert.

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    let projectId: string = "";
+    test.beforeAll(async ({ mockBackendService }) => {
+        mockBackendService
+            .createProject({ name: "Project 1" })
+            .then((project) => (projectId = project.response.id));
+    });
+
+    test.beforeEach(async ({ page }) => {
+        await page.goto(`/project/${projectId}/settings/general`);
+    });
```
Slartibartfass2: Currently, it seems to me that the input is filled before the file has finished loading. We can fix this by disabling the input and the button until the project is fully loaded.

Slartibartfass2 (changes requested):
We're nearly done here. Great work!
Let's fix these last few points, and please rebase so that we can see whether it works with the new API version.
After that, I will have a last look and I assume that we can merge it after that.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
domienderle: Thank you for the hint. Now it makes sense where the error regarding to this ``` ?.response.then(() => {``` comes from. Totally makes sense to mock all calls that are called. Will do that in the future.

domienderle: Thank you for the hint. Now it makes sense where the error regarding to this  ?.response.then(() => { comes from. Totally makes sense to mock all calls that are called. Will do that in the future.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Probably a good idea, yes.

tests/e2e/project-name-settings.test.ts
domienderle: Yes you are right. I think the way a implemented it now is the prettier way, by just waiting on the project and creating and deleting the project before and after every test so the tests don't interfere. But if you think different, you can surely give your opinion on it.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: It's still there for `loadingProject`

tests/e2e/project-name-settings.test.ts
Slartibartfass2: While this works for the E2E tests, it still doesn't solve the problem. You can learn more about this [here](https://playwright.dev/docs/navigations#hydration).

Slartibartfass2 (changes requested):
We're almost done with this. Let's fix the last few things. There are some open conversations left.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Good call. This was not intended. Must have happened through one of my rebases, that I included it unintentionally.

domienderle (commented):
Now the tests and the user should be able to rename the project after the whole page is loaded.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
Merseleo: As Felix already mentioned: please add a comment or thumbs up or so, so I know you worked on this comment ^^.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Merseleo: ... and you use and you expect something different than the default behavior. It seems like you fixed it, if you only want to set a specific name, you can simply this code as follows: 

```typescript
const mockUpdateProject = mockApiCall("updateProject", createProject({name: "New Project Name"}));
```
This creates a project object as you wanted with some default values that are not interesting and overwrites the project name. Was that your goal?

tests/e2e/project-name-settings.test.ts
Merseleo: I agree with Felix. Actually I would suggest, that you create a new project once before this test, delete this one project afterwards and do not create a new project for each test. In general I would swap the tests, to simulate that a user first tries to change the name to a wrong project name and learns from this mistake and afterwards enters a correct project name und sees the update.

tests/e2e/fixtures/project-settings-fixture.ts
```diff
@@ -0,0 +1,16 @@
...
+    projectSettingsPage: DevProjectSettingsPage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - project settings page
+ */
+export const test = base.extend<ProjectSettingsPage>({
+    projectSettingsPage: async ({ page }, use) => {
+        await use(new DevProjectSettingsPage(page));
```
Merseleo, line 18: As I recently learned you can use to navigate to the settings page every time (see #10 from Moritz) if you would change the setup from `beforeEach` to `beforeAll`.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Merseleo: Same can be used for the next test.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+            })
+            .response.then(() => {
+                invalidate("data:getProjectById");
+            })
+            .catch((error) => {
+                updateProjectError = {
+                    errorTitle: "Something went wrong while updating the project name.",
+                };
+                console.error(`Couldn't update project: ${error}`);
+            });
```
Merseleo: It would be nice, if the user can see, that the project is renamed and whether it was successful (more optional). So please add the feature, that the "Rename" button shows a loading spinner and "Renaming" while the call is processed by the backend (I implemented something similar in the #52 while submitting a review). And if the update was successful, maybe show a toast, but this is optional.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Merseleo, line 1: One thing I want to say: The process has little feedback. So when I load the page and simply click "Rename" nothing happens, which is ok. However, I think it would be nice to get a hint that nothing is happening because the name is the same, e.g. via a toast or so.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Merseleo: Additionally it should not be possible to rename it again, while the project name is changed imo. Your implementation does work and somehow kind of buffer the requests, but please test your feature again with a higher response delay of the mock backend to simulate a slow connection.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+            inputId="project-name-input"
+            label="Project Name"
+            placeholder="Project Name"
+            required
+            schema={Schema.projectName}
+            type="text"
+            value={projectName}
+        />
+        <Button
+            class="text-md h-12.5 w-full md:mt-5.5 md:w-25"
```
Merseleo:
```suggestion
            class="text-md w-full md:mt-5.5 md:w-25"
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,124 @@
...
+        class="flex w-full max-w-xl flex-col items-center gap-2.5 md:h-fit md:max-w-xl md:flex-row md:items-start"
+        onsubmit={handleSubmit}
+    >
+        <Input
+            bind:this={projectNameInput}
+            class="h-full w-full"
+            disabled={loading}
+            inputId="project-name-input"
+            label="Project Name"
+            placeholder="Project Name"
```
Merseleo:
```suggestion
            placeholder={loading ? "Loading" : "New Project Name ..."}
```
What do you think of this change?

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            .createProject({ name: "Project 1" })
+            .then((project) => (projectId = project.response.id));
+        await page.goto(`/project/${projectId}/settings/general`);
+        await expect(page.getByRole("heading", { name: "Project 1" })).toBeVisible();
+    });
+
+    test.afterEach(async ({ mockBackendService }) => {
+        mockBackendService.softDeleteProject({ id: projectId });
+    });
+    test('When the user enters an valid project name, the new project name should be updated to "New Project".', async ({
```
Merseleo:
```suggestion
    test("When the user enters a valid project name, then the name of the project should be updated to this name.", async ({
```

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,41 @@
...
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName("New Project");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByRole("heading", { name: "New Project" });
+        await expect(projectNameHeader).toBeVisible();
```
Merseleo: Maybe also add a check, that the old header is not visible anymore.

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    test("When the user enters an invalid project name, the new project name should not be updated.", async ({
```
Merseleo:
```suggestion
    test("When the user enters an invalid project name, then the name of the project remains unchanged.", async ({
```

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,41 @@
...
+    }) => {
+        await projectSettingsPage.changeProjectName("Test Project");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByRole("heading", { name: "Test Project" });
+        await expect(projectNameHeader).toBeVisible();
+
+        await projectSettingsPage.changeProjectName("");
+        await projectSettingsPage.checkForErrors();
+        await expect(projectNameHeader).toBeVisible();
+        await expect(projectNameHeader).toHaveText("Test Project");
```
Merseleo: An invalid project name is also the same name, right? So either add this check here in this test or even better create a new test case where you check, that "changing" the project name to the same name will obviously have no effect, but a hint is provided that it is the same name. I know actually you should know this because the title is in the upper left corner, but maybe it is too long and trimmed or and you changed it multiple times or so. But I am open for discussion regarding this point (relates to some other of my comments).

Merseleo (dismissed):
So I would continue Felix review as he is on vacation. There are some comments kind of duplicates so do not be afraid of the number of comments. But imo there are some things open regarding edge cases like slow server responses or something similar. I hope we can discuss some of my suggestions.

tests/e2e/fixtures/project-settings-fixture.ts
domienderle: How exactly do you mean? Because, in my case I need the ```projectId```, so I would have to pass the ```projectId``` to the fixture in some way. Additionally, as I understood I have to navigate to the page before each single test and not only before all tests, right? So maybe I just don't understand what you mean correctly, so lighten me up.

Slartibartfass2:
@Merseleo thanks for replacing me, while I was gone. I will continue from here.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: This is a good suggestion. Is implemented

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: Yes you are completely right. I didn't focus on the delay enough and thus there is no real delay handling right now. Will be implemented.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: That is true. In terms of good usability this is necessary.

Slartibartfass2: Looks better now

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: There's still this overflow
You can simply fix this by setting the max width
```svelte
        <div class="max-w-xl">
            <ErrorAlert errorTitle={updateProjectError.errorTitle} />
        </div>
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    /**
+     * Changes the name of the project, if a non-blank project name is provided.
+     *
+     * Therefore, send a request to the backend to update the project with the
+     * provided name.
+     */
+    async function handleSubmit(event: Event) {
+        loading = true;
```
Slartibartfass2: Let's set `updateProjectError` to undefined at the beginning because otherwise it will always show an error even if subsequent calls were successful.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: As discussed, we should define a unified way to show the loading state of the button. Until this is decided, let's at least add `Renaming` while the button is in the loading state.
`md:w-42` should be fine to fit the loading state without resizing the button.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: IMO the dots at the end are not necessary. What are your thoughts?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,149 @@
...
+
+            try {
+                await backendService.updateProject({
+                    project: Project.create(projectData),
+                    mask: {
+                        paths: fieldMask,
+                    },
+                }).response;
+                console.log(projectName);
+                console.log(newProjectName);
```
Slartibartfass2: These seem like debug logs, do we need them in production?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,149 @@
...
+                switch (newProjectName) {
+                    case projectName:
+                        toast("Please enter a new project name.");
+                        break;
```
Slartibartfass2: Can we check this before sending the request? In this case, we want to omit sending a request because it is not necessary to change the name.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,149 @@
...
+                    case "":
+                        toast("Please enter a project name.");
+                        break;
```
Slartibartfass2: This is already checked by the validation and we therefore don't need to check for it.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,149 @@
...
+    >
+        <Input
+            bind:this={projectNameInput}
+            class="h-full w-full"
+            disabled={loading}
+            inputId="project-name-input"
+            label="Project Name"
+            placeholder={loading ? "Loading" : "New Project Name ..."}
+            required
+            schema={Schema.projectName}
```
Slartibartfass2, line 115: Currently, we show the message "Please provide a project name", when the input is blank. To me, this seems not enough because technically a project name was provided. My suggestion would be to change it to "Please provide a non-blank project name", which provides a bit more information.
What are your thoughts on this?

tests/e2e/fixtures/project-settings-fixture.ts
Slartibartfass2: I don't understand what he means either. So @Merseleo please explain.
What I see is that we currently create a project before each test, but only delete the last one in the `afterAll` call. This means that there are some leftover non-deleted projects, which may cause conflicts with other tests.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        expect(projectRenameInput).toBeInTheDocument();
+        await waitFor(() => {
+            expect(projectRenameInput).toHaveValue(Projects.demoProject.name);
+        });
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+        expect(renameButton).toBeInTheDocument();
+    });
+
+    test("When the input field is empty or the project name is the same, then the button should not change the name", async () => {
+        const mockCall = mockApiCall("getProjectById", createProject());
```
Slartibartfass2: Do we require mocking this? `getProjectById` is never called in the component. We pass the `loadingProject` promise directly to the component.

Slartibartfass2 (changes requested):
This is what I found so far. There are some open conversations left from the review of Leonhard. I didn't attend to them, as I wasn't sure whether you worked on them yet.

tests/e2e/project-name-settings.test.ts
domienderle: I added this extra cases.

tests/e2e/project-name-settings.test.ts
domienderle: Good point. Is added.

tests/e2e/fixtures/project-settings-fixture.ts
domienderle: I refactored this point, to create a project in ```beforeAll``` and navigate to the page in the ```fixture```.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: You are right. @mowi12 approved it.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: True. I totally missed that point. Is implemented.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: You got me there. Just forgot to remove them.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: That is right. The backend call should not be called and the integration test that checks this was wrong, but if we remove the unnecessary toast for an empty project name it can easily be refactored to check the project name before the call. Is implemented.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: I was not sure about how strict the validation handles the empty string, but you are right. With the validation this check is useless.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: I mean basically the user will see that the project name is empty and should know that he has to type something else than blanks in. But if we want to be really, really strict you are right, then it is better to mention the blanks. Is done.

tests/integration/settings/project-settings/general/project-name-settings.test.ts
domienderle: No, you are right. Since I updated the mockcall as @Merseleo proposed, this mockcall is useless.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: This should've had replaced the already existing surrounding div. Let's fix this:
```svelte
    {#if updateProjectError}
        <div class="max-w-xl">
            <ErrorAlert errorTitle={updateProjectError.errorTitle} />
        </div>
    {/if}
```

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: While this basically works now, there's still a small issue. When I change the project name successfully, and then click "Rename" again (with the same project name) a call is still sent because the variable `projectName` is not updated when the call was successful.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: It's just for better user feedback

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Changes the name of the project, if a non-blank project name is provided.
+     *
+     * Therefore, send a request to the backend to update the project with the
+     * provided name.
+     */
+    async function handleSubmit(event: Event) {
+        updateProjectError = undefined;
+        loading = showLoadingSpinner = true;
```
Slartibartfass2, line 57: Now `showLoadingSpinner` is duplicated or is there a reason why we should keep this flag?

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,57 @@
+import { test } from "./fixtures/project-settings-fixture";
+import { expect } from "@playwright/test";
+
+export let projectId: string = "";
+export const projectReviewSettingsProjectName = "Project 1";
```
Slartibartfass2: This variable is exported but never used outside this file. The variable is only used in the `beforeAll` hook. Should it be used somewhere else?

tests/e2e/project-name-settings.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    test("When the user enters an empty project name, then the name of the project remains unchanged.", async ({
+        page,
+        projectSettingsPage,
+    }) => {
+        await projectSettingsPage.changeProjectName(" ");
+        await projectSettingsPage.checkForErrors();
+        const projectNameHeader = page.getByRole("heading", { name: "Project 1" });
+        await expect(projectNameHeader).toBeVisible();
+        // Here is no toast necessary, because it is handled by the input field.
```
Slartibartfass2: Let's check for the error message.

Slartibartfass2 (changes requested):
Great work! We're nearly done here.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
domienderle: I know what you mean. The reason why I left the flag there is the initial load. If I would remove the ```showLoadingSpinner``` I would have to replace the flag for the loading Button with ```loading```. But this causes, that the rename button shows ```Renaming``` and the loading circle not only on the click of the button, but also on every reload of the page. Imo this is not the expected behaviour. The ```loading``` flag on the other hand is necessary, to disable the button and the input field, the renaming is in process.

tests/e2e/project-name-settings.test.ts
domienderle: You are right. This is a stupid copy paste mistake, I forgot to adjust. Sorry

tests/e2e/project-name-settings.test.ts
domienderle: You are right. Don't know why I didn't test this. Basicallly there is an error-message we can test on.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: I see, I didn't notice this, but it makes sense as we have two different loading states. In this case we need two different loading state flags.

Slartibartfass2 (commented):
Amazing, we're done here!
Now, just rebase and you will get my approval afterward.

Slartibartfass2 (approved):

PR #310, opened 2025-04-11
Added visual regression testing

### LINKED ISSUE #282: [Feature]: Add visual regression testing

_No response_

At the moment we wouldn't recognize, if the appearance of the frontend changes.

To detect unwanted visual changes, we should introduce visual regression testing for the integration and E2E tests.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #282 

## What I have made

- the `loading` and `errorLoading` helpers are used throughout all tests
- I added visual regression assertions including their snapshots, which are stored by the git LFS
- the snapshots are created by the custom `expectMatchingScreenshot` function, which makes setting the path more easily
- To make the screenshots reproducible, I changed the setup for the E2E tests so that the frontend also runs in a docker container.
  - for that I created a separate compose file, which then can be easily extended
  - as the different frontends already contain the correct mock backend URL, we don't require the interceptor fixture anymore
  - I rewrote the start containers script to work with the new docker setup (mostly AI work)

## Checklist

  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - [x] end-to-end tests


Merseleo:
Can this be closed? If we still want to use it, we can simply reopen the pull request, right?

PR #312, opened 2025-04-13
Fix type stripping error for gRPC client in E2E test

### LINKED ISSUE #308: [Bug]: Cannot create gRPC client in a e2e test

_No response_

### What happened?

When I write an end to end test, it is sometimes necessary to work with data, so either add data like paper or projects before a test is executed or delete anything after the test finished. To add / remove data it would be very cumbersome to use the pages and not the API calls directly. However, to use the API directly, we need to create another gRPC client in the setup of the end to end tests and provide this client as fixture.

The problem is, that the generated gRPC Client contains something called `parameter property` and as playwright uses the newest node version and so type stripping `parameter property` is not allowed anymore.
So when I add a gRPC client as fixture I got a syntax error and the test execution crashes

### Steps to reproduce

1. Add a gRPC client in a setup fixture: 
```typescript
export const setup = test.extend<{ mockBackendService: SnowballRClient }>({
    mockBackendService: async ({ mockBackendUrl }, use) => {
        const transport = new GrpcWebFetchTransport({
            baseUrl: mockBackendUrl,
        });
        const mockBackendService = new SnowballRClient(transport);
        await use(mockBackendService);
    },
});
``` 
2. Execute the end to end tests using `npm run test:e2e`

### What did you expect?

I would expect no error and the client is created and can be used

### Anything else?

The error: Image

I found a [possible solution that disables the new experimental type stripping feature](https://github.com/kiesraad/abacus/issues/902). I think this is the only solution as we cannot influence how the gRPC client is generated and should not overwrite the generated code. However the disabling option is not very beautiful and we should investigate how to find a better solution to add this flag.

Merseleo:
Closes #308

## What I have made

- I added a second test fixture that provides a gRPC client (with a correct connection depending on the browser the test runs in)
- I added a script to run playwright or node without type stripping, which is enabled per fault since Node 23 and the gendered gRPC code violates the new guidelines for Node 23
- I demonstrated the usage of the gRPC client in a cleanup function in a E2E test and adapted the package.json commands for running E2E tests, so it also works with Node 23

## Checklist

  ~~- [ ] unit tests~~
  ~~- [ ] integration tests~~
  - [x] end-to-end tests


scripts/run-node.js
```diff
@@ -0,0 +1,7 @@
+// Since with Node 23 type stripping no longer has to be explicitly activated,
+// but is standard and must be explicitly deactivated,
+// this script checks whether node is running in version 23 and if so,
+// type stripping is deactivated, since the generated GRPC client does not work with type stripping.
+
+export const isNode23 = process.version.startsWith("v23");
```
Slartibartfass2: What happens in later versions?

scripts/run-node.js
Slartibartfass2, line 1: Is it necessary to put this into a separate file?

scripts/run-playwright.js
```diff
@@ -0,0 +1,16 @@
...
+
+function runPlaywright(args = []) {
+    const nodeArgs = isNode23 ? [NO_TYPE_STRIPPING_FLAG] : [];
+    spawnSync("node", [...nodeArgs, BASE_COMMAND, ...args], {
+        shell: true,
+        stdio: "inherit",
+    });
+}
+
+runPlaywright(process.argv.slice(2));
```
Slartibartfass2, line 35: Why do we skip the first two arguments?

tests/e2e/fixtures/general-fixture.ts
```diff
@@ -34,3 +36,25 @@ export const test = base.extend<{ forEachTest: void } & TestOptions>({
...
+>({
+    mockBackendUrl: ["http://localhost:3001", { option: true }],
+    mockBackendService: async ({ mockBackendUrl }, use) => {
+        const transport = new GrpcWebFetchTransport({
+            baseUrl: mockBackendUrl,
+        });
+        const mockBackendService = new SnowballRClient(transport);
+        await use(mockBackendService);
+    },
+});
```
Slartibartfass2, line 58: Is there a reason why this is separate from the default test fixture?

tests/e2e/project-creation.test.ts
```diff
@@ -8,6 +8,12 @@ test.describe("Creating a new project", () => {
         await page.goto("/");
     });
 
+    test.afterAll(async ({ mockBackendService }) => {
+        mockBackendService.softDeleteProject({ id: "0" });
+        mockBackendService.softDeleteProject({ id: "1" });
+        mockBackendService.softDeleteProject({ id: "2" });
+    });
```
Slartibartfass2: Which projects are deleted here?

Slartibartfass2 (changes requested):
Great work! I'm excited to use it for my tests.
I tested it with node 22 and 23.

Could you please add more information to the wiki that explains how to use these fixtures.

scripts/run-node.js
Merseleo: No, that's not necessary, I just wanted to keep it apart, because theoretically this problem could also occur with other tools.

scripts/run-node.js
Merseleo: I don't know. At the moment this feature is experimental so can still change. Maybe they delete it in Node 24, maybe not, but as far as we dont know what happens in the future I would only distinguish between Node 23 and Node 22 and earlier versions.

Merseleo: But I they keep it **and** protobuf-ts doesn't get updated to conform to the new type stripping **and** somebody updates to Node 24, we probably need to update the script to parse the version number and check whether it is >= 23 or <23.

scripts/run-playwright.js
Merseleo: Short answer: because the person I copied the script from did this. Long answer: You want to skip the first two arguments as they are not relevant and should not be use as _...args_ in the spawn command, because these two arguments are the path of the node executable and of this script (see [here](https://nodejs.org/api/process.html#processargv) for details).

tests/e2e/project-creation.test.ts
Merseleo: The projects that are created in this test group. But you are right, only two projects are created, that should be deleted afterwards.

I do this 1) for demonstration how to use the mockBackendService and 2) to "rollback" the changes I made.

tests/e2e/fixtures/general-fixture.ts
Merseleo: Yes there are two reasons:
1) if you dont want to use the `mockBackendService` in your tests, you could also import only the `defaultTest` fixture (or vice versa) and
2) in my opinion that are two different use cases and should not be mixed in one fixture. Otherwise I would add two functionalities to one mixture and would need to explain both in one comment.

Merseleo (commented):
I hope I could answer all your questions.

scripts/run-playwright.js
Slartibartfass2: Please add this explanation as a comment.

tests/e2e/fixtures/general-fixture.ts
Slartibartfass2: Alright, that's fine for me.

tests/e2e/project-creation.test.ts
Slartibartfass2: I wonder if we should do this in each test instead of the `afterAll` hook. Not sure what's the best way to clean this up.

scripts/run-node.js
Slartibartfass2: I think that's fine for now. Maybe we should declare the used node version for this project.

Slartibartfass2 (changes requested):
Some minor things left.

Slartibartfass2:
Don't forget to rebase when you're finished.

scripts/run-playwright.js
Merseleo: I added it, is it clearer now?

tests/e2e/project-creation.test.ts
Merseleo: In my opinion we should not do this after each test, as this whole test group tests an entire workflow I want to explicitly test, that if I already created a project, the second project is added below the first project and both exist. So if I would cleanup after each test in this case, it would be something different. In general, there are of course some scenarios where we should cleanup after each test. 

Another solution would be add further calls to the API that allow a cleanup, but this shouldn't be content of this PR.

scripts/run-node.js
Merseleo: But the point is that we don't just use one node version, but several. At most, we could list the node version up to which it works in any case (and perhaps the oldest tested version and in between it should also work). I would about this to the README.md or do you have a better solution?

scripts/run-playwright.js
Slartibartfass2: Now it's clear.

scripts/run-node.js
Slartibartfass2: You can use the [engines field](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines) in `package.json` to define the supported range of versions.

tests/e2e/project-creation.test.ts
Slartibartfass2: That's fine for me.

But so far, we do not test if another project was created before, or am I missing something?

Slartibartfass2 (changes requested):
Nearly done!

scripts/run-node.js
Merseleo: Ah thank you for the hint.

tests/e2e/project-creation.test.ts
Merseleo: No you're not missing anything, you are right. We dont test it and it wouldn't make a difference if I would not cleanup. But then we come back to 1): I wanted to demonstrate how to use the `mockBackendService` and as a nice add-on we get the cleanup if we will test whether another project was created.

Merseleo (commented):
Thank you for the hints.

Slartibartfass2 (dismissed):

wiki/Contributing.md
```diff
@@ -103,21 +104,65 @@ instead of classes or text content.
...
+As running the entire end-to-end test suite can take some time, this can be useful for testing new
+end-to-end tests and getting feedback on them more quickly.
+
+#### Requirements
+
+The E2E tests require a mock-backend to be running for each supported browser (Chromium, Firefox, and WebKit).
+These mock-backends simulate the backend responses.
+
+To manually start all necessary mock-backends, you can use the provided script, which will start one instance per
+browser if you provide the three ports:
```
Slartibartfass2: What was that again? "Keep it consistent"? :D
```suggestion
The E2E tests require a mock backend to be running for each supported browser (Chromium, Firefox, and WebKit).
These mock backends simulate the backend responses.

To manually start all necessary mock backends, you can use the provided script, which will start one instance per
browser if you provide the three ports:
```

wiki/Contributing.md
```diff
@@ -103,21 +104,65 @@ instead of classes or text content.
...
+
+   In general, it makes sense to create an own fixture, if you want to use an object in multiple tests or
+   run a certain function before / after each test.
+
+2. **Use Page Object Models (POMs)**:
+   To make tests more maintainable and readable, use page object models (pom) to encapsulate UI logic.
+   For example, the `pom/create-project-dialog-model.ts` file wraps a page with custom selectors and
+   helper functions for the `CreateProjectDialog` component,
+   improving clarity and reusability across tests. In particular, these _pom_s can be wrapped in a fixture
+   so that they can be accessed directly in each test without having to create a separate \_pom_ in each test.
```
Slartibartfass2: Let's link to the files so that they can be viewed in the browser.
```suggestion
1. **Use the test fixture**:
   Instead of importing the `test` fixture directly from `@playwright/test` directly,
   use the `test` variable exported from
   [`fixtures/general-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/6ac47c3f0330d72d63fe977e476e51ba29937c7b/tests/e2e/fixtures/general-fixture.ts).
   This ensures that the API calls to the mock backend are automatically redirected to the correct mock backend
   based on the browser the test is running in.

   In general, it makes sense to create an own fixture, if you want to use an object in multiple tests or
   run a certain function before / after each test.

2. **Use Page Object Models (POMs)**:
   To make tests more maintainable and readable, use page object models (POMs) to encapsulate UI logic.
   For example, the [`pom/create-project-dialog-model.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/6ac47c3f0330d72d63fe977e476e51ba29937c7b/tests/e2e/pom/create-project-dialog-model.ts)
   file wraps a page with custom selectors and helper functions for the `CreateProjectDialog` component,
   improving clarity and reusability across tests. In particular, these POMs can be wrapped in a fixture
   so that they can be accessed directly in each test without having to create a separate POM in each test.
```

Slartibartfass2 (dismissed):
Only minor things

wiki/Contributing.md
Merseleo: You are completely right, I fixed it.

wiki/Contributing.md
Merseleo: I agree but I wouldn't add this link with the sha, but the branch name develop as these links would work always after the merge, are you ok with that?

Slartibartfass2 (approved):

PR #319, opened 2025-04-16
 Fix NO_FCP error in "Lighthouse Auditing" CI job

### LINKED ISSUE #263: [Bug]: Lighthouse fails with error: NO_FCP

_No response_

### What happened?

When a review is requested, the lighthouse job is executed in the CI and always fails with the error NO_FCP.
Regarding to the error description, we face a timeout, so after one second no first contentful paint is shown, even after building the application to production. 

### Steps to reproduce

1. Request review in any pull request
2. Wait until "Lighthouse Auditing" job is run.

### What did you expect?

That the job doesn't fail, e.g. by a higher timeout for FCP or a higher performance of the web application.

### Anything else?

Image

Slartibartfass2:
Closes #263

## What I have made

- We now use the docker image together with the mock backend
- the timeout probably originates in the network timeout

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no source code added)


Merseleo (approved):
LGTM, the job now runs without timeout and I didnt find any errors. Maybe we have to find a better way than checking project 1 and paper 1, but if I remember correctly we always create at least 1 project and 1 paper so it should always work.

PR #320, opened 2025-04-16
[Use Case]: Change name

### LINKED ISSUE #58: [Use Case]: Change Name

User

### Describe the use case

As user, I want to change my name, which I set when I signed up.

### Acceptance criteria

- the user can change their first and last name
- the first and last name have to be non-blank and have a maximum length

### Anything else?

Related to #2

Part of #8

mowi12:
Closes #58 

## What I have made

- Added new account settings section ("Change Name"), allowing users to change their name.

## Checklist

  - ~~[ ] unit tests~~ not needed
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
+++ b/src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
@@ -1,0 +1,113 @@
+<script lang="ts">
+    import { invalidate } from "$app/navigation";
+    import Button from "$lib/components/primitives/button/button.svelte";
+    import { backendService } from "$lib/grpc-api";
+    import { User } from "$lib/model/api/user";
+    import type { ApiError } from "$lib/model/general";
+    import { Schema } from "$lib/schemas";
+    import { toast } from "svelte-sonner";
+    import Input from "../../input/Input.svelte";
+    import ErrorAlert from "../../utils/ErrorAlert.svelte";
+    import SettingsSection from "../SettingsSection.svelte";
+    import { generateFieldMask } from "protobuf-fieldmask";
+
+    interface Props {
+        user: User;
+    }
+
+    const { user }: Props = $props();
+
+    let firstNameInput: Input;
+    let lastNameInput: Input;
+
+    let updateUserError: ApiError | undefined = $state(undefined);
+
+    async function handleSubmit(event: Event) {
+        event.preventDefault();
+
+        const isFirstNameValid = firstNameInput.validate();
+        const isLastNameValid = lastNameInput.validate();
+        if (!(isFirstNameValid && isLastNameValid)) return;
+
+        const userData: Partial<User> = {
+            id: user.id,
+            firstName: firstNameInput.getValue(),
+            lastName: lastNameInput.getValue(),
+        };
+
+        const updatedUser = User.create(userData);
+        const maskPaths = generateFieldMask(userData).filter((path) => path !== "id");
+
+        backendService
+            .updateUser({
+                user: updatedUser,
+                mask: {
+                    paths: maskPaths,
+                },
+            })
+            .response.then(() => {
+                invalidate("data:getCurrentUser");
+                toast("Successfully updated your name.");
+
+                // Make sure the input fields are not focused after submitting
+                (document.activeElement as HTMLElement)?.blur();
+            })
+            .catch((error) => {
+                updateUserError = { errorTitle: "Something went wrong while updating user." };
+                console.error(error);
+            });
+    }
+</script>
+
+<!--
+@component
+Component for changing the user's name.
+
+Usage:
+```svelte
+    <ChangeNameSettings {user} />
+```
+-->
+<SettingsSection sectionTitle="Change Name">
+    <p class="text-base">
+        Your name is displayed in the member list of projects and when hovering over your character.
+    </p>
+    <form
+        class="flex w-full max-w-100 flex-col items-center gap-2.5 md:h-fit md:max-w-200 md:flex-row md:items-start"
+        onsubmit={handleSubmit}
+    >
+        <Input
+            bind:this={firstNameInput}
+            class="h-full w-full"
+            errorMessagePrefix="First Name must contain"
+            inputId="first-name-input"
+            label="First Name"
+            placeholder="First Name"
+            required
+            schema={Schema.firstName}
+            type="text"
+            value={user.firstName}
+        />
+        <Input
+            bind:this={lastNameInput}
+            class="h-full w-full"
+            errorMessagePrefix="Last Name must contain"
+            inputId="last-name-input"
+            label="Last Name"
+            placeholder="Last Name"
+            required
+            schema={Schema.lastName}
+            type="text"
+            value={user.lastName}
+        />
+        <Button class="text-md h-12.5 w-full md:mt-5.5 md:w-25" type="submit">Rename</Button>
+    </form>
+    {#if updateUserError}
+        <div class="max-w-100 md:max-w-200">
+            <ErrorAlert
+                errorDetails={updateUserError.errorDetails}
+                errorTitle={updateUserError.errorTitle}
+            />
+        </div>
+    {/if}
+</SettingsSection>
--- a/tests/e2e/change-account-settings.test.ts
+++ b/tests/e2e/change-account-settings.test.ts
@@ -1,0 +1,34 @@
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/account-settings-fixture";
+
+test.describe("Changing username", () => {
+    test("When the user enters a valid first and last name, the new name should be updated.", async ({
+        page,
+        accountSettingsPage,
+    }) => {
+        accountSettingsPage.changeUsername("Zeta", "Zeta");
+
+        await expect(page.getByRole("textbox", { name: "First Name" })).toHaveValue("Zeta");
+        await expect(page.getByRole("textbox", { name: "Last Name" })).toHaveValue("Zeta");
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+    });
+
+    test("When the user enters an invalid first or last name, the new name should not be updated.", async ({
+        page,
+        accountSettingsPage,
+    }) => {
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+
+        accountSettingsPage.changeUsername(" ", "Beta");
+        await expect(
+            page.getByText("First Name must contain at least 1 non-whitespace character"),
+        ).toBeVisible();
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+
+        accountSettingsPage.changeUsername("Alpha", " ");
+        await expect(
+            page.getByText("Last Name must contain at least 1 non-whitespace character"),
+        ).toBeVisible();
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+    });
+});
--- a/tests/e2e/fixtures/account-settings-fixture.ts
+++ b/tests/e2e/fixtures/account-settings-fixture.ts
@@ -1,0 +1,20 @@
+import { DevAccountSettingsPage } from "../pom/account-settings-model";
+import { DevHomePage } from "../pom/home-page-model";
+import { test as base } from "./general-fixture";
+
+type AccountSettingsPage = {
+    accountSettingsPage: DevAccountSettingsPage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - account settings page
+ */
+export const test = base.extend<AccountSettingsPage>({
+    accountSettingsPage: async ({ page }, use) => {
+        await page.goto("/");
+        const homepage = new DevHomePage(page);
+        await homepage.openAccountSettings();
+        await use(new DevAccountSettingsPage(page));
+    },
+});
--- a/tests/e2e/fixtures/account-settings-fixture.ts
+++ b/tests/e2e/fixtures/account-settings-fixture.ts
@@ -1,3 +1,4 @@
+import { expect } from "@playwright/test";
 import { DevAccountSettingsPage } from "../pom/account-settings-model";
 import { DevHomePage } from "../pom/home-page-model";
 import { test as base } from "./general-fixture";
@@ -13,6 +14,7 @@ type AccountSettingsPage = {
 export const test = base.extend<AccountSettingsPage>({
     accountSettingsPage: async ({ page }, use) => {
         await page.goto("/");
+        await expect(page.getByRole("heading", { name: "SnowballR" })).toBeVisible(); // TODO: Remove after user context is added
         const homepage = new DevHomePage(page);
         await homepage.openAccountSettings();
         await use(new DevAccountSettingsPage(page));
--- a/tests/e2e/pom/home-page-model.ts
+++ b/tests/e2e/pom/home-page-model.ts
@@ -18,4 +18,13 @@ export class DevHomePage {
         await expect(this.createProjectDialog).not.toBeVisible();
         await this.openCreateProjectDialogButton.click();
     }
+
+    /**
+     * Opens the account settings page.
+     */
+    async openAccountSettings() {
+        await expect(this.page.getByRole("heading", { name: "Settings" })).not.toBeVisible();
+        await this.page.getByRole("button", { name: /[A-Z]{2}/ }).click();
+        await this.page.getByRole("link", { name: "Settings" }).click();
+    }
 }
--- a/tests/integration/settings/user-settings/account-settings.test.ts
+++ b/tests/integration/settings/user-settings/account-settings.test.ts
@@ -1,0 +1,103 @@
+import { render, screen } from "@testing-library/svelte";
+import { describe, expect, test, vi } from "vitest";
+import ChangeNameSettings from "$lib/components/composites/settings/user-settings/ChangeNameSettings.svelte";
+import { Users } from "$tests/example-data";
+import { backendService } from "$lib/grpc-api";
+import userEvent from "@testing-library/user-event";
+import type { User } from "$lib/model/api/user";
+import { mockApiCall, mockFailedApiCall } from "$tests/setupTest";
+
+describe("ChangeNameSettings", () => {
+    test("When the component is rendered, some text, two input fields and a button are shown in order to change the name", () => {
+        render(ChangeNameSettings, {
+            props: {
+                user: Users.johnDoe,
+            },
+        });
+
+        const changeNameSection = screen.getByTestId("settings-section-change-name");
+        expect(changeNameSection).toBeInTheDocument();
+
+        expect(screen.queryByText("Change Name")).toBeInTheDocument();
+
+        expect(screen.getByLabelText("First Name")).toBeInTheDocument();
+        expect(screen.getByLabelText("First Name")).toHaveValue("John");
+        expect(screen.getByLabelText("Last Name")).toBeInTheDocument();
+        expect(screen.getByLabelText("Last Name")).toHaveValue("Doe");
+        expect(screen.getByRole("button", { name: "Rename" })).toBeInTheDocument();
+    });
+
+    test("When one of the input fields is empty, then the button should not change the name", async () => {
+        const mockCall = vi.spyOn(backendService, "updateUser");
+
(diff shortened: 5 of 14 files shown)


src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+    }
+</script>
+
+<!--
+@component
+Component for changing the user name setting.
+
+Usage:
+```svelte
+    <SettingsSection {user} />
```
Slartibartfass2:
```suggestion
    <ChangeNameSettings {user} />
```

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+            .catch((error) => {
+                updateUserError = { errorTitle: "Something went wrong while updating user." };
+                console.error(error);
+            });
+    }
+</script>
+
+<!--
+@component
+Component for changing the user name setting.
```
Slartibartfass2: What are your thoughts on this?
```suggestion
Component for changing the user's name.
```

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+Usage:
+```svelte
+    <SettingsSection {user} />
+```
+-->
+<SettingsSection sectionTitle="Change Name">
+    <p class="text-base">
+        Your name is displayed in the member list of projects and when hovering over your character.
+        <span class="font-bold">Note</span>: This doesn’t change your registered name in the city
+        where you live in.
```
Slartibartfass2: 😄 This was just for fun in the design, I don't think we should add this here haha
```suggestion
```

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+            label="Last Name"
+            placeholder="Last Name"
+            required
+            schema={Schema.lastName}
+            type="text"
+            value={user.lastName}
+        />
+        <div>
+            <!-- Spacer -->
+            <div class="md:h-5.5"></div>
```
Slartibartfass2: Didn't you use `item-end` before? What was the reason for the change?

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+            <Button class="text-md h-12.5 w-25" type="submit">Rename</Button>
```
Slartibartfass2: Minor: Should we let the button fill the width when the inputs are ordered vertically?

tests/e2e/pom/home-page-model.ts
```diff
@@ -18,4 +18,15 @@ export class DevHomePage {
...
+
+    /**
+     * Opens the account settings page.
+     */
+    async openAccountSettings() {
+        await expect(this.page.getByRole("heading", { name: "Settings" })).not.toBeVisible();
+        await this.page
+            .getByRole("button", { name: /[A-Z]{2}/i, exact: true, disabled: false })
+            .click();
+        await this.page.getByRole("link", { name: "Settings ⌘⇧S" }).click();
```
Slartibartfass2: As the shortcut changes in the future and may be hidden, it's best to not match it.

tests/e2e/fixtures/account-settings-fixture.ts
```diff
@@ -0,0 +1,19 @@
...
+    accountSettingsPage: DevAccountSettingsPage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - account settings page
+ */
+export const test = base.extend<AccountSettingsPage>({
+    accountSettingsPage: async ({ page }, use) => {
+        const homepage = new DevHomePage(page);
```
Slartibartfass2, line 18: Is there a way to ensure that we are on the home page?

tests/integration/settings/user-settings/account-settings.test.ts
```diff
@@ -0,0 +1,81 @@
+import { render, screen } from "@testing-library/svelte";
+import { describe, expect, test, vi } from "vitest";
+import ChangeNameSettings from "$lib/components/composites/settings/user-settings/ChangeNameSettings.svelte";
+import { Users } from "$tests/example-data";
+import { backendService } from "$lib/grpc-api";
+import userEvent from "@testing-library/user-event";
+import type { User } from "$lib/model/api/user";
+
+describe("ChangeNameSettings", () => {
+    test("When component is rendered, some text, two input fields and a button are shown in order to change the name", () => {
```
Slartibartfass2:
```suggestion
    test("When the component is rendered, some text, two input fields and a button are shown in order to change the name", () => {
```

tests/integration/settings/user-settings/account-settings.test.ts
```diff
@@ -0,0 +1,81 @@
...
+            },
+        });
+
+        const changeNameSection = screen.getByTestId("settings-section-change-name");
+        expect(changeNameSection).toBeInTheDocument();
+
+        expect(screen.queryByText("Change Name")).toBeInTheDocument();
+
+        expect(screen.getByLabelText("First Name")).toBeInTheDocument();
+        expect(screen.queryByLabelText("First Name")).toHaveValue("John");
```
Slartibartfass2: Is there a reason why you chose different ways to query the input?

tests/integration/settings/user-settings/account-settings.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        await userEvent.click(renameButton);
+        expect(mockCall).not.toHaveBeenCalled();
+    });
+
+    test("When both input fields are filled, then the button should change the name", async () => {
+        const mockCall = vi.spyOn(backendService, "updateUser").mockImplementation(() => {
+            return {
+                response: Promise.resolve({ firstName: "Peter", lastName: "Fox" } as User),
+            } as ReturnType<(typeof backendService)["updateUser"]>;
+        });
```
Slartibartfass2: Is there a reason why we don't use `mockApiCall` here?

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+                user: updatedUser,
+                mask: {
+                    paths: maskPaths,
+                },
+            })
+            .response.then(() => {
+                invalidate("data:getCurrentUser");
+                toast("Successfully updated your name.");
+
+                (document.activeElement as HTMLElement)?.blur();
```
Slartibartfass2, line 53: Maybe add a comment that explains what this does.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,114 @@
...
+            <!-- Spacer -->
+            <div class="md:h-5.5"></div>
+            <Button class="text-md h-12.5 w-25" type="submit">Rename</Button>
+        </div>
+    </form>
+    {#if updateUserError}
+        <ErrorAlert
+            errorDetails={updateUserError.errorDetails}
+            errorTitle={updateUserError.errorTitle}
+        />
```
Slartibartfass2: When an error appears it should align with the form imo, so I suggest using the same max width. What are your thoughts on this?

tests/e2e/change-account-settings.ts
```diff
@@ -0,0 +1,40 @@
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/account-settings-fixture";
+
+test.describe("Changing username", () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto("/");
+    });
```
Slartibartfass2: It looks a bit misleading when going to the homepage here, but by using the fixture we go to the account settings page. Do you think there's a better way to do this?

tests/e2e/change-account-settings.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    test("When the user enters a valid first and last name, the new name should be updated.", async ({
+        page,
+        accountSettingsPage,
+    }) => {
+        accountSettingsPage.changeUsername("Zeta", "Zeta");
+
+        await expect(
+            page.getByRole("button", { name: "ZZ", exact: true, disabled: false }),
+        ).toBeVisible();
```
Slartibartfass2: Is it necessary to define `exact` and `disabled`?

Furthermore, shouldn't we also check whether the input fields are updated or rather stay the same?

tests/e2e/change-account-settings.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        accountSettingsPage.changeUsername("", "Beta");
+
+        await expect(
+            page.getByRole("button", { name: "ZZ", exact: true, disabled: false }),
+        ).toBeVisible();
```
Slartibartfass2: Let's also check for errors (`alert` role).

Slartibartfass2 (changes requested):
Great work! It appears that you have accustomed yourself to our project.

There are only some minor things left. The functionality works as expected.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Yes, this is fine. It's shorter and more descriptive.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Okay, then let's remove it!

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: The reason for the change are the error notifications, when the user enters an invalid name. They shift the layout of the button on appearance. I did not find any layout configuration for the button, so it stays at the correct position with and without the error notifications.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: You're right. There already is a max width restriction, but it is too large for the vertical layout. I add the same restriction for the vertical layout.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Good idea!

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Yes, good point!

tests/integration/settings/user-settings/account-settings.test.ts
mowi12: No, this probably was code pilot after trying different things. I must have not seen that there were two different queries used. I will change it to the same one.

tests/integration/settings/user-settings/account-settings.test.ts
mowi12: In order to test, if a mocked API call has been called, we need the `MockInstance` which is not returned by the `mockApiCall()` function. But you're right. I shouldn't have simply used the `spyOn` function directly, but rather added the wanted functionality to either a new method or extend the `mockApiCall()` method.
I will update the code in this regard!

mowi12 (commented):
This review was very helpful! I agree with most/all of you points. Everything concerning the e2e tests will be done as soon as the API works again.

tests/e2e/fixtures/account-settings-fixture.ts
mowi12: By using the for each (goto /) in the test we ensure this.

tests/e2e/change-account-settings.ts
mowi12: I just looked at the other test case and thought this was the way to go. 
I mean, this ensures, that we always start at the homepage, which makes it easier to define the fixtures. So they can assume they are on the homepage, as you questioned in another review point.

tests/e2e/change-account-settings.ts
mowi12: I'm not sure. First it seemed that I needed them, but now I don't need them any more.

tests/e2e/change-account-settings.ts
mowi12: Good idea!

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
Slartibartfass2: What about adding this as margin to the button, i.e., `md:mt-5.5`?

tests/e2e/pom/home-page-model.ts
Slartibartfass2: As discussed, please answer on this and let's rather not use a regex but instead just set `exact` to false.

tests/integration/settings/user-settings/account-settings.test.ts
Slartibartfass2: Alright great.

tests/e2e/change-account-settings.ts
Slartibartfass2: Let's move this statement to the fixture, so the whole navigation is in one place, and it is not misleading.

tests/e2e/change-account-settings.ts
Slartibartfass2: Shouldn't we also check whether the input fields are updated or rather stay the same?

tests/e2e/change-account-settings.ts
Slartibartfass2: You forgot the parentheses after `toBeVisible`

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
Slartibartfass2: I don't see these changes, did they go missing?

Slartibartfass2 (changes requested):
Nearly done! Let's finish this.

Furthermore, please do not resolve the conversations yourself. That's my job 😄 

Edit: Don't forget to rebase.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Yes, this is a very good idea!

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: It seems as I did not fully understand you the last time. Now the changes should be there.

tests/e2e/change-account-settings.ts
mowi12: Done!

tests/e2e/change-account-settings.ts
mowi12: You're right. In the case, the user updated the username, the values of the input fields should be the name. I added these lines.

Imo the `exact` keyword makes sense because it also matches case, and we know how the value should look like.

tests/e2e/change-account-settings.ts
mowi12: Done!

mowi12 (commented):
I tried to please you and make all the changes. I also added a new integration test in order to test the error alert on a failed api call.

tests/e2e/change-account-settings.ts
Slartibartfass2: Are sure that you committed these changes?

tests/e2e/pom/home-page-model.ts
```diff
@@ -18,4 +18,13 @@ export class DevHomePage {
...
         await this.openCreateProjectDialogButton.click();
     }
+
+    /**
+     * Opens the account settings page.
+     */
+    async openAccountSettings() {
+        await expect(this.page.getByRole("heading", { name: "Settings" })).not.toBeVisible();
+        await this.page.getByRole("button", { name: /[A-Z]{2}/ }).click();
+        await this.page.getByRole("link", { name: "Setting" }).click();
```
Slartibartfass2:
```suggestion
        await this.page.getByRole("link", { name: "Settings" }).click();
```

tests/e2e/change-account-settings.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        accountSettingsPage,
+    }) => {
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+
+        accountSettingsPage.changeUsername("Alpha", " ");
+        await expect(page.getByTestId("error-message")).toBeVisible();
+        await expect(page.getByRole("button", { name: "ZZ", exact: true })).toBeVisible();
+
+        accountSettingsPage.changeUsername(" ", "Beta");
+        await expect(page.getByTestId("error-message")).toBeVisible();
```
Slartibartfass2: Instead of using the test IDs let's use the error messages. We want to test for things that the user can see.

Slartibartfass2 (changes requested):
Only minor things.

Also, the e2e tests still fail. Let's investigate this.

tests/e2e/pom/home-page-model.ts
mowi12: My bad, fixed it.

tests/e2e/change-account-settings.test.ts
mowi12: Generally I would approve, but in this case, the error message depends on the validation criteria of the input field and I don't think it would make sense to test for this message here.
Firstly, because the criteria could change, or the message itself, and imo this shouldn't impact this test.
What do you think?

mowi12:
Locally, all my tests succeed. I think it's the order of the tests which can cause them to fail ...

tests/e2e/change-account-settings.test.ts
Slartibartfass2: I see your point, but still disagree. With this, the error message could be empty as well, so the user wouldn't see any error message. This wouldn't be the case when testing for an actual message.
We don't consider the criteria or message to change in the near future, or at least not frequently.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -0,0 +1,111 @@
...
+            bind:this={lastNameInput}
+            class="h-full w-full"
+            inputId="last-name-input"
+            label="Last Name"
+            placeholder="Last Name"
+            required
+            schema={Schema.lastName}
+            type="text"
+            value={user.lastName}
+        />
```
Slartibartfass2: I nearly have overseen this. To make this component more reusable, we don't use an error message prefix by default. Please add this to get a sensible error message.
```suggestion
        <Input
            bind:this={firstNameInput}
            class="h-full w-full"
            inputId="first-name-input"
            label="First Name"
            placeholder="First Name"
            required
            schema={Schema.firstName}
            errorMessagePrefix="First Name must contain"
            type="text"
            value={user.firstName}
        />
        <Input
            bind:this={lastNameInput}
            class="h-full w-full"
            inputId="last-name-input"
            label="Last Name"
            placeholder="Last Name"
            required
            schema={Schema.lastName}
            errorMessagePrefix="Last Name must contain"
            type="text"
            value={user.lastName}
        />
```

Slartibartfass2 (changes requested):
Nearly done!
There's still [this](https://github.com/SE-UUlm/snowballr-frontend/pull/320#discussion_r2056299009) open conversation.

The tests appear to be flaky. Use more retries to reproduce that.
Can you please add `failOnFlakyTests: !!process.env.CI,` to the playwright config? This will help us catch such cases in the CI.

Please also change `trace` in `use` to `"retain-on-failure"` so that we can better trace such errors.

It appears that they fail in the after hook, but I'm confused that testing logic is executed in there.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: You're right, i didn't see that. I changed it!

mowi12 (commented):
I have made all the changes. But the e2e tests seem to still fail. Let's investigate this!

Slartibartfass2 (approved):

PR #325, opened 2025-04-18
[Use Case]: Toggle shortcuts visibility

### LINKED ISSUE #10: [Use Case]: Toggle Shortcuts visibility

mowi12:
Closes #10 

## What I have made

- Added new shortcut settings section ("Shortcuts Visibility"), allowing users to change the visibility of shortcuts.
- Updated visible shortcuts to change visibility based on setting

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ No documentation needed
  - ~~[ ] unit tests~~ No basic components created
  - [x] integration tests
  - [x] end-to-end tests


mowi12:
The e2e tests for the checkbox are not working. I think it has something to do with [this](https://github.com/microsoft/playwright/issues/13470) bug. I do not have a fix for this issue.

src/lib/components/composites/navigation-bar/UserMenu.svelte
```diff
@@ -62,9 +63,11 @@
...
                             <item.icon class="mr-2 size-4" />
                             <span>{item.label}</span>
-                            <DropdownMenu.Shortcut>
-                                {item.shortcut}
-                            </DropdownMenu.Shortcut>
+                            {#if shortcutsVisibility.isActivated}
+                                <DropdownMenu.Shortcut>
+                                    {item.shortcut}
+                                </DropdownMenu.Shortcut>
+                            {/if}
```
Slartibartfass2: What are your thoughts on moving this to `dropdown-menu-shortcut.svelte`? We have the shadcn components in our codebase so that we can customize them if we'd like to and I think this is the case here.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,20 @@
+import { browser } from "$app/environment";
+
+/**
+ * Stores, whether shortcuts are visible for the current user or not.
+ *
+ * The default state is active, so if preference is not changed, the visibility of shortcuts is on.
+ */
+let shortcutsVisibilityState = $state(true);
```
Slartibartfass2: The state is never read from the local storage, i.e., when I close and reopen the browser, this is true if I set it to false before.
This doesn't seem like the functionality we want.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const shortcutsVisibility = {
+    get isActivated() {
```
Slartibartfass2: Can we think of another name than `shortcutsVisibility.isActivated`? It seems a bit long to me for just checking whether the shortcuts are shown or not.

Slartibartfass2 (changes requested):
Great work! I didn't check the tests. I will review these after the rebase.

There are some things to discuss.

src/lib/components/composites/navigation-bar/UserMenu.svelte
mowi12: Yes you're right! This also removes the second `if`-code part and applies the DRY (Don't repeat yourself) principle.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: Oh you're right. This should obviously persist over sessions. I updated the code accordingly so the initial state is only set to true if there is no value stored in the local storage. Else the state from local storage is loaded.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: Yes, we can do that. I made is shorter and more concise.
Do you like the new version or should we rather use one from the following:

- `shortcuts.areVisible`
- `shortcuts.visible`

I like the current version because it is short but the `is` keyword hints to a boolean type.

mowi12 (commented):
Thanks for the review. I think I changed everything to the correct version for all the parts that needed improvement.
I also updated the e2e tests accordingly.

tests/e2e/pom/home-page-model.ts
```diff
@@ -23,8 +23,17 @@ export class DevHomePage {
...
         await this.page.getByRole("link", { name: "Settings" }).click();
+        await this.page.getByRole("link", { name: "Account" }).click();
+    }
+
+    /**
+     * Opens the shortcuts settings page.
+     */
+    async openShortcutsSettings() {
+        await this.page.getByRole("button", { name: /[A-Z]{2}/ }).click();
+        await this.page.getByRole("link", { name: "Settings" }).click();
```
Slartibartfass2: What are your thoughts on extracting this to a separate method `openSettings`?

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
```diff
@@ -0,0 +1,38 @@
...
+```
+-->
+<SettingsSection sectionTitle="Shortcuts Visibility">
+    <div class="flex-start flex space-x-2">
+        <Checkbox id="shortcuts-visibility-checkbox" bind:checked={shortcuts.isVisible} />
+        <div class="grid gap-1.5 pt-0.25 leading-none">
+            <Label
+                class="text-base leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
+                >Display Shortcuts
+            </Label>
```
Slartibartfass2: What about using `for` so you can use the label to access it in the E2E tests.
```suggestion
            <Label
                class="text-base leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
                for="shortcuts-visibility-checkbox"
            >
                Display Shortcuts
            </Label>
```

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
```diff
@@ -0,0 +1,38 @@
...
+
+Usage:
+```svelte
+    <ShortcutsVisibilitySettings />
+```
+-->
+<SettingsSection sectionTitle="Shortcuts Visibility">
+    <div class="flex-start flex space-x-2">
+        <Checkbox id="shortcuts-visibility-checkbox" bind:checked={shortcuts.isVisible} />
+        <div class="grid gap-1.5 pt-0.25 leading-none">
```
Slartibartfass2: `leading-none` doesn't seem to make a difference.
```suggestion
        <div class="grid gap-1.5 pt-0.25">
```

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
```diff
@@ -0,0 +1,38 @@
...
+            <p class="text-description">
+                Shortcuts are displayed next to their respective element.
+            </p>
+            <div class="flex flex-col gap-3 pt-1.5 md:flex-row">
+                <Button class="text-md h-12.5">Button without Shortcut</Button>
+                <Button class="text-md h-12.5">
+                    Button with Shortcut
+                    <span class="text-md text-gray-400"> Ctrl+D </span>
+                </Button>
+            </div>
```
Slartibartfass2: At ~800px the right button is cut off and a horizontal scrollbar is added. Let's fix this.
image

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,45 @@
...
+ * The state is stored in local storage, so it persists across sessions.
+ */
+let shortcutsVisibilityState = $state(initialValue);
+
+export const shortcuts = {
+    get isVisible() {
+        return shortcutsVisibilityState;
+    },
+    set isVisible(value: boolean) {
+        const newValue = !!value;
```
Slartibartfass2, line 19: Why the double negation?

tests/e2e/pom/shortcuts-settings-model.ts
```diff
@@ -0,0 +1,23 @@
+import { type Locator, type Page } from "@playwright/test";
+
+export class DevShortcutsSettingsPage {
+    readonly page: Page;
+    readonly shortcutsVisibilityCheckbox: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.shortcutsVisibilityCheckbox = page.getByRole("checkbox");
```
Slartibartfass2: As already mentioned above, use the label here.

tests/integration/settings/user-settings/shortcuts-visibility-settings.test.ts
```diff
@@ -0,0 +1,31 @@
...
+        );
+        expect(shortcutsVisibilitySection).toBeInTheDocument();
+
+        expect(screen.queryByText("Shortcuts Visibility")).toBeInTheDocument();
+        expect(screen.queryByText("Display Shortcuts")).toBeInTheDocument();
+
+        expect(screen.getByRole("checkbox")).toBeInTheDocument();
+    });
+
+    test("When checkbox is toggled, then the global shortcuts visibility state changes", () => {
```
Slartibartfass2:
```suggestion
    test("When the checkbox is toggled, then the global shortcuts visibility state changes", () => {
```

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,45 @@
...
+ * The default state is active, so if the user has not set anything yet, the shortcuts are visible.
+ * The state is stored in local storage, so it persists across sessions.
+ */
+let shortcutsVisibilityState = $state(initialValue);
+
+export const shortcuts = {
+    get isVisible() {
+        return shortcutsVisibilityState;
+    },
+    set isVisible(value: boolean) {
```
Slartibartfass2, line 18: What about `areVisible`?

Slartibartfass2 (changes requested):
Great! We're nearly done here.

tests/e2e/pom/home-page-model.ts
mowi12: Yeah, why not. Looks very sensible!

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
mowi12: Makes sense. So we test what the user sees and are more specific as we do not reference any checkbox if there are more to come!

tests/e2e/pom/shortcuts-settings-model.ts
mowi12: Done

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: As discussed, we keep `isVisible`.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: The `!!` operator is a concise way to force any value into its corresponding boolean equivalent.
The first `!` takes a value, determines if it's "truthy" or "falsy", and returns the opposite boolean.
The second negates it again.

Why use it here?
While TypeScript gives a type hint, `!!` acts as a runtime safeguard and promises that the value will always be strictly true or false.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
mowi12: Fixed it.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
mowi12: You're right. I also removed `flex-start` because it also wasn't needed.

mowi12 (commented):
This time, everything should be working and all the issues you found have be improved.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
```diff
@@ -0,0 +1,39 @@
...
+-->
+<SettingsSection sectionTitle="Shortcuts Visibility">
+    <div class="flex space-x-2">
+        <Checkbox id="shortcuts-visibility-checkbox" bind:checked={shortcuts.isVisible} />
+        <div class="grid w-full max-w-100 gap-1.5 pt-0.25 md:max-w-200">
+            <Label
+                class="text-base leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
+                for="shortcuts-visibility-checkbox"
+                >Display Shortcuts
+            </Label>
```
Slartibartfass2: Just for the optics
```suggestion
            <Label
                class="text-base leading-none peer-disabled:cursor-not-allowed peer-disabled:opacity-70"
                for="shortcuts-visibility-checkbox"
            >
                Display Shortcuts
            </Label>
```

src/lib/global-state/shortcuts-visibility-state.svelte.ts
Slartibartfass2: Yes, but why the negation isn't this set by the isChecked binding of the checkbox?

tests/e2e/pom/shortcuts-settings-model.ts
```diff
@@ -0,0 +1,23 @@
+import { type Locator, type Page } from "@playwright/test";
+
+export class DevShortcutsSettingsPage {
+    readonly page: Page;
+    readonly shortcutsVisibilityCheckbox: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.shortcutsVisibilityCheckbox = page.getByLabel("shortcuts-visibility-checkbox");
```
Slartibartfass2: Shouldn't this be the following?
```suggestion
        this.shortcutsVisibilityCheckbox = page.getByLabel("Display Shortcuts");
```

tests/e2e/pom/home-page-model.ts
```diff
@@ -20,11 +20,26 @@ export class DevHomePage {
...
+    async openAccountSettings() {
+        this.openSettings();
+        await this.page.getByRole("link", { name: "Account" }).click();
+    }
+
+    /**
+     * Opens the shortcuts settings page.
+     */
+    async openShortcutsSettings() {
+        this.openSettings();
```
Slartibartfass2: You still need to await this
```suggestion
        await this.openSettings();
        await this.page.getByRole("link", { name: "Account" }).click();
    }

    /**
     * Opens the shortcuts settings page.
     */
    async openShortcutsSettings() {
        await this.openSettings();
```

Slartibartfass2 (changes requested):
Great work! Now we have to wait until the new API version is integrated.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
mowi12: Yeah, why not :)

tests/e2e/pom/shortcuts-settings-model.ts
mowi12: Oh, you're right. Funny, that the tests did work...

tests/e2e/pom/home-page-model.ts
mowi12: You're absolutely right.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: Yes, currently it is. But maybe there will be a case in the future, where this is not the case. This only acts as a safe guard.
I can remove it, if you think it is not necessary.

mowi12 (commented):
I hope, now, everything is to your liking.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,45 @@
...
+if (browser) {
+    const storedValue = localStorage.getItem(STORAGE_KEY);
+    if (storedValue !== null) {
+        try {
+            const parsedValue = JSON.parse(storedValue);
+            if (typeof parsedValue === "boolean") {
+                initialValue = parsedValue;
+            }
+        } catch (error) {
+            console.error("Error parsing ${STORAGE_KEY} from local storage: ", error);
```
Slartibartfass2: Use ` to interpolate strings
```suggestion
            console.error(`Error parsing ${STORAGE_KEY} from local storage:`, error);
```

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    set isVisible(value: boolean) {
+        const newValue = !!value;
+        shortcutsVisibilityState = newValue;
+
+        if (browser) {
+            try {
+                localStorage.setItem(STORAGE_KEY, JSON.stringify(newValue));
+            } catch (error) {
+                console.error(`Error saving ${STORAGE_KEY} to local storage: `, error);
```
Slartibartfass2: Minor: double space
```suggestion
                console.error(`Error saving ${STORAGE_KEY} to local storage:`, error);
```

tests/e2e/pom/home-page-model.ts
```diff
@@ -20,11 +20,26 @@ export class DevHomePage {
...
     /**
-     * Opens the account settings page.
+     * Opens the user settings page.
      */
-    async openAccountSettings() {
-        await expect(this.page.getByRole("heading", { name: "Settings" })).not.toBeVisible();
+    async openSettings() {
         await this.page.getByRole("button", { name: /[A-Z]{2}/ }).click();
         await this.page.getByRole("link", { name: "Settings" }).click();
     }
```
Slartibartfass2: Let's wait until the page has navigated to the settings page. It seems like the test error could be raised by that.

Slartibartfass2 (dismissed):
Only the tests left. Don't forget to rebase

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: I must have missed this. I will update this.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: You're right! I will remove the additional space.

tests/e2e/pom/home-page-model.ts
mowi12: You're right. This makes sense!

src/lib/global-state/shortcuts-visibility-state.svelte.ts
```diff
@@ -0,0 +1,45 @@
+import { browser } from "$app/environment";
+
+const STORAGE_KEY = "showShortcuts";
```
Merseleo, line 3: Since Felix is on holiday and I'm his replacement, I'll add my wishes to the review: Great idea with this constant, could you please add this for the review mode state as well (I know, my code, my fault, but it would be great, especially for maintainability).

Merseleo: Same applies for the check, whether value could be set to the local storage. If you think this doesn't belong here, please close this conversation on your own.

tests/e2e/pom/home-page-model.ts
Merseleo: Probably the way you wait will work, but is not recommended:
Bildschirmfoto 2025-05-08 um 20 35 41

Afaik you should use [`page.waitForURL`](https://playwright.dev/docs/navigations#waiting-for-navigation) instead. Same applies for your reload function etc.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
```diff
@@ -0,0 +1,40 @@
...
+                for="shortcuts-visibility-checkbox"
+            >
+                Display Shortcuts
+            </Label>
+            <p class="text-description">
+                Shortcuts are displayed next to their respective element.
+            </p>
+            <div class="flex w-full flex-col flex-wrap gap-3 pt-1.5 md:flex-row">
+                <Button class="text-md h-12.5">Button without Shortcut</Button>
+                <Button class="text-md h-12.5">
```
Merseleo: I thought we dont want to set the height of the button to explicitly 50px anymore?

Merseleo (changes requested):
So this review is open for a very long time now, lets finish it.

P.S. please rebase.

src/lib/components/composites/settings/user-settings/ShortcutsVisibilitySettings.svelte
mowi12: Yes, you're right. This is still old code, from the time where this wasn't the case.
I will remove it.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: Good idea, this makes the code much more cleaner!

tests/e2e/pom/home-page-model.ts
mowi12: But this is exactly the issue we are currently having. `waitForURL` does not work 100% because it only waits for hydration. But the website does take longer to load, therefore we literally have to wait until there is no communication.
At least in when i tried it, it does not work.
We also use it in 2 other places, because this was the most reliable way to test.

I think this is okay here until we want to rework all the tests.

mowi12 (commented):
So, I think everything is done now.

src/lib/global-state/shortcuts-visibility-state.svelte.ts
Merseleo: You did a really great job! However, please complete this by deleting the `onMount(...)` method in the global `+layout.svelte` where the review mode is still loaded the old way. I think you need to only remove it without adding anything.

tests/e2e/pom/home-page-model.ts
Merseleo: I read your code again and I don't quite understand why you want to wait for it explicitly at all. After navigation Playwright automatically waits until the page is loaded and the other elements are available and enabled?! Or is this the problem with "enabled although not fully hydrated" what you are solving in the other PR? Then let's move the customization to the other PR.

So my suggestion: Delete this `waitForLoadState`.

Merseleo: And in general I think your solution with `await page.getByRole("button", { name: /[A-Z]{2}/ }).click();` to open the user menu is flaky (at least locally on my machine). Why not explicitly set the name to the registered user? 

However, I think all the E2E problems we should solve in a different PR.

Merseleo (changes requested):
I think everything is fine expect for some unnecessary code that should be deleted to prevent confusion or so.

tests/e2e/pom/home-page-model.ts
Merseleo: I just figured out a solution to your problem that works without `waitForLoadState` and implement the best practices from playwright.

To make your E2E tests work after deleting `waitForLoadState` everywhere, you could add the following code snippets: 
- `await expect(page.getByRole("heading", {name: "SnowballR"})).toBeVisible();` in the `shortcuts-setting-fixture.ts` after you goto "/" to make sure that the homepage is completely loaded (because only if so, the navigation bar will be visible which you check with this statement)
- refactor your `reloadWait` method to not only get the page object but also an element to waited on after the reload:

```typescript
import { expect, type Locator, type Page } from "@playwright/test";

/**
 * Reloads the page and waits for it to be completely loaded.
 *
 * To ensure that the page has been fully loaded,
 * provide an element whose visibility should be waited for.
 * If possible, this element should be the last to become visible.
 *
 * @param page - The page to reload
 * @param element - The element to be waited for
 */
export async function reloadWait(page: Page, element: Locator) {
    await page.reload();
    await expect(element).toBeVisible();
}
``` 
that can used as follows:
```typescript
await reloadWait(page, page.getByRole("heading", { name: "Settings" }));
```

src/lib/global-state/shortcuts-visibility-state.svelte.ts
mowi12: You're right. This code is not needed anymore.

tests/e2e/pom/home-page-model.ts
mowi12: > I read your code again and I don't quite understand why you want to wait for it explicitly at all. After navigation Playwright automatically waits until the page is loaded and the other elements are available and enabled?! Or is this the problem with "enabled although not fully hydrated" what you are solving in the other PR? Then let's move the customization to the other PR.
> 
> So my suggestion: Delete this waitForLoadState.
> 

Yes exactly. That was the reason. But you found an elegant way to solve this!

Merseleo: Ok, so hopefully I didn't miss anything and we do not introduce bad code now.

Merseleo (approved):
In my opinion we are good to go!

PR #327, opened 2025-04-18
Add sveltekit-lighthouse-runner CLI tool

### LINKED ISSUE #326: [Feature]: Add lighthouse CLI tool

_No response_

There were some questions about the usage of lighthouse and I thought we could possibly make a CLI tool that helps with the usage of lighthouse.

A CLI tool that helps with selecting the route and provides several arguments.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #326

## What I have made

- I found this CLI tool and it looked like good way to use lighthouse for our project

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no logic added)

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -30,10 +30,11 @@ jobs:
             - name: Run frontend with dummy mock-backend
               run: |
                   echo "PORT=5173" > .env
+                  echo "RANDOMNESS_SEED=U25vd2JhbGxS" >> .env
                   docker compose --profile mock-local up -d --wait
 
             - name: Run Lighthouse
-              run: npm run lighthouse:all -- --params="projectId=1,paperId=1-1"
+              run: npm run lighthouse:all -- --params="projectId=11,paperId=11-0"
 
             - name: Archive coverage report
               uses: actions/upload-artifact@v4
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -18,34 +18,26 @@ jobs:
               with:
                   submodules: "recursive"
 
+            - name: Setup
+              uses: ./.github/workflows/setup/
+              with:
+                  node_version: ${{ env.node_version }}
+
             - name: Set up Docker Compose
               uses: docker/setup-compose-action@v1
 
             # We use the mock-local compose profile to run the frontend together with a dummy mock-backend
             - name: Run frontend with dummy mock-backend
               run: |
-                  echo "PORT=8080" > .env
+                  echo "PORT=5173" > .env
                   docker compose --profile mock-local up -d --wait
 
-            - name: Generate URLs
-              id: urls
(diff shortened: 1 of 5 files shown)


wiki/Contributing.md
```diff
@@ -169,16 +169,17 @@ For more advanced usage and documentation, refer to the official [Playwright doc
...
 
-To run a Lighthouse audit on the app, you can use (note we use the url of the preview server here):
+To run all available routes, use:
 
 ```bash
-npm run lighthouse -- http://localhost:4173/path
+npm run lighthouse:all
 ```
 
 The report will be saved in the `./lighthouse-reports` directory.
```
mowi12: I'm not quite sure if this is needed, but maybe add some hint how this could be served?
Not everyone has a VSC server extension installed or knows instantly how to view the report, although it is a `.html` file which could simply opened with the browser.

.github/workflows/lighthouse.yml
```diff
@@ -18,34 +18,26 @@ jobs:
...
-                          .map(x => x.replace('[paperId]', '1').replace('[projectId]', '1').replace(/\([a-z\-]+\)\//g, ''))
-                          .map(x => `http://localhost:8080/${x}`)
-                          .join('\n');
-                      console.log('generated URLs:', urls);
-                      core.setOutput('urls', urls);
-
-            - name: Audit URLs using Lighthouse
-              uses: treosh/lighthouse-ci-action@v12
+            - name: Run Lighthouse
+              run: npm run lighthouse:all -- --params="projectId=1,paperId=1-1"
```
mowi12: As discussed, this is not optimal as we do not know if these ids are valid. But at least, all other paths will be tested.
I assume you thought of other ways and found no better solution. Maybe we could force the mock-backend to always create a project with these ids?
This won't change this issue, but i wanted to write this idea down, so we can remember it!

mowi12 (changes requested):
As this is a very small review, there is not much to discuss. I found no problems with the code itself. Good job!

.github/workflows/lighthouse.yml
Slartibartfass2: I added a randomness seed functionality in https://github.com/SE-UUlm/snowballr-mock-backend/pull/26. This way we can choose a seed for which the IDs always refer to existing projects and papers.

wiki/Contributing.md
Slartibartfass2: I added more description.

Slartibartfass2 (commented):
Thanks for the review. Let's finish the update in the mock backend and then adjust this one.

mowi12 (approved):
Very good idea to handle the IDs with a seed in the backend!
Now everything seems fine.

.github/workflows/lighthouse.yml
mowi12: Very nice idea!

PR #328, opened 2025-04-21
[Use case] define keywords for review mode

### LINKED ISSUE #17: [Use Case]: Define Keywords For Review Mode

domienderle:
Closes #17 

## What I have made

Added a new section in the project review settings, that contains an input field for tags. The tag creation relies on the following rules and the created keyword tags are stored in the local storage: 

- the user can define multiple keywords
- the keywords must be unique
- the user can delete already existing keywords
- the maximum number of keywords is limited
- a keyword must be non-blank
- the maximum length of a keyword is limited
- a keyword can include spaces, i.e. it doesn't have to be just one word

On hover over an existing tag, the button to remove the tag appears.

## Checklist

  - ~~[] unit tests~~ not necessary, because there is no real logic to be tested
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -2,15 +2,28 @@
     import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
     import TagsInput from "svelte-unstyled-tags";
     import { toast } from "svelte-sonner";
+    import { onMount } from "svelte";
+
 
     interface Props {
+        projectId: string;
         maximumAmountOfTags: number;
         maximalTagLength: number;
     }
 
-    const { maximumAmountOfTags = 25, maximalTagLength = 15 }: Props = $props();
+    const { projectId, maximumAmountOfTags = 25, maximalTagLength = 15 }: Props = $props();
 
     let tags: string[] = $state([]);
+
+    /**
+     * Loads the already defined keyword tags for the selected project.
+     */
+    onMount(async () => {
+        const projectKeywords = localStorage.getItem(`project_${projectId}_keywords`);
+        if (projectKeywords) {
+            tags = JSON.parse(projectKeywords);
+        }
+    });
 
     /**
      * Validates the input for a tag. If the input is incorrect, the tag is not created.
@@ -34,6 +47,7 @@
                 tags.pop();
                 return toast(`Keyword name already exists!`);
             }
+            localStorage.setItem(`project_${projectId}_keywords`, JSON.stringify(tags))
             return tags;
         }
     }
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -74,13 +74,13 @@ length of a single tag are customizable.
 
 Usage:
 ```svelte
-    <KeywordSettings {maximumAmountOfTags} {maximalTagLength} />
+    <KeywordSettings {projectId} />
 ```
 -->
 <SettingsSection sectionTitle="Keywords">
     <div class="max-w-2xl">
         <ChipsInput
-            label="Define keywords that are highlighted in the abstract of a paper in the review mode."
+            label="Define keywords that are highlighted in the abstract of a paper when the review mode is activated."
             placeholder="Add keyword"
             validate={(newTag) => validateTags(newTag)}
             bind:items={tags}
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -73,7 +73,7 @@ Usage:
                 allTagsWrapperClasses="flex flex-row items-center gap-x-2 gap-y-2 flex-wrap"
                 inputClasses="focus:outline-none focus:ring-0 pl-2 py-1"
                 labelText="Define keywords that are highlighted in the abstract of a paper in the review mode."
-                removeTagButtonClasses="hidden group-hover:inline-block ml-2 text-black-500 hover:text-black-500 cursor-pointer"
+                removeTagButtonClasses="max-w-0 overflow-hidden group-hover:max-w-[20px] group-hover:ml-2 transition-all duration-200 text-black-500 hover:text-black-500 cursor-pointer"
                 showLabel={true}
                 tagWrapperClasses="group flex items-start bg-gray-200 text-black px-2 py-1.5 rounded-xl"
                 tagsInputWrapperClasses="flex items-start border-2 border-gray-200 py-2 px-3 rounded-md mt-2 h-[84px] overflow-auto"
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -8,8 +8,8 @@
         projectId: string;
     }
 
-    const maximumNumberOfTags = 50;
-    const maximumTagLength = 100;
+    const MAX_NUMBER_OF_TAGS = 50;
+    const MAX_TAG_LENGTH = 100;
 
     export function getProjectKeywordsKey(projectId: string): string | null {
         return localStorage.getItem(`project_${projectId}_keywords`);
@@ -37,27 +37,24 @@
      * @param newTag - validates the newly created tag
      * @returns - the validation result of the new tag
      */
-    function validateTags(newTag: string): ValidationResult {
-        let validationResult: ValidationResult = { success: true };
-        if (tags && tags.length > 0) {
-            if (newTag.trim().length == 0) {
-                validationResult = { success: false, error: "Empty keywords are not allowed!" };
-            } else if (newTag.trim().length > maximumTagLength) {
-                validationResult = {
-                    success: false,
-                    error: `Maximum keyword length of ${maximumTagLength} characters exceeded!`,
-                };
-            } else if (tags.length > maximumNumberOfTags) {
-                validationResult = { success: false, error: `Maximum number of keywords reached!` };
-            } else if (
-                Array.from(tags)
-                    .slice(0, tags.length - 1)
-                    .includes(newTag)
-            ) {
-                validationResult = { success: false, error: `Keyword name already exists!` };
-            }
+    function validateTag(newTag: string): ValidationResult {
+        if (tags.length >= MAX_NUMBER_OF_TAGS) {
+            return { success: false, error: `Maximum number of keywords reached!` };
         }
-        return validationResult;
+        const newTagLength = newTag.trim().length;
+        if (newTagLength === 0) {
+            return { success: false, error: "Blank keywords are not allowed!" };
+        }
+        if (newTagLength > MAX_TAG_LENGTH) {
+            return {
+                success: false,
+                error: `Maximum keyword length of ${MAX_TAG_LENGTH} characters exceeded!`,
+            };
+        }
+        if (tags.includes(newTag)) {
+            return { success: false, error: `Keyword name already exists!` };
+        }
+        return { success: true };
     }
 
     $effect(() => {
@@ -82,7 +79,7 @@ Usage:
         <ChipsInput
             label="Define keywords that are highlighted in the abstract of a paper when the review mode is activated."
             placeholder="Add keyword"
-            validate={(newTag) => validateTags(newTag)}
+            validate={(newTag) => validateTag(newTag)}
             bind:items={tags}
         />
     </div>
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -11,8 +11,8 @@
     const MAX_NUMBER_OF_TAGS = 50;
     const MAX_TAG_LENGTH = 100;
 
-    export function getProjectKeywordsKey(projectId: string): string | null {
-        return localStorage.getItem(`project_${projectId}_keywords`);
+    export function getProjectKeywordsKey(projectId: string) {
+        return `project_${projectId}_keywords`;
     }
 
     const { projectId }: Props = $props();
@@ -23,7 +23,7 @@
      * Loads the already defined keyword tags for the selected project.
      */
     onMount(async () => {
-        const projectKeywords = getProjectKeywordsKey(projectId);
+        const projectKeywords = localStorage.getItem(getProjectKeywordsKey(projectId));
         if (projectKeywords) {
             tags = JSON.parse(projectKeywords);
         }
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -58,7 +58,7 @@
     }
 
     $effect(() => {
-        localStorage.setItem(`project_${projectId}_keywords`, JSON.stringify(tags));
+        localStorage.setItem(getProjectKeywordsKey(projectId), JSON.stringify(tags));
     });
 </script>
 
--- a/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
@@ -1,16 +1,21 @@
 <script lang="ts">
     import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
-    import TagsInput from "svelte-unstyled-tags";
-    import { toast } from "svelte-sonner";
     import { onMount } from "svelte";
+    import ChipsInput from "$lib/components/composites/input/ChipsInput.svelte";
+    import type { ValidationResult } from "$lib/model/general";
 
     interface Props {
         projectId: string;
-        maximumAmountOfTags: number;
-        maximalTagLength: number;
     }
 
-    const { projectId, maximumAmountOfTags = 25, maximalTagLength = 15 }: Props = $props();
+    const maximumNumberOfTags = 50;
+    const maximumTagLength = 100
+
+    export function getProjectKeywordsKey(projectId: string): string |null {
+        return localStorage.getItem(`project_${projectId}_keywords`);
+    }
+
+    const { projectId }: Props = $props();
 
     let tags: string[] = $state([]);
 
@@ -18,7 +23,7 @@
      * Loads the already defined keyword tags for the selected project.
      */
     onMount(async () => {
-        const projectKeywords = localStorage.getItem(`project_${projectId}_keywords`);
+        const projectKeywords = getProjectKeywordsKey(projectId)
         if (projectKeywords) {
             tags = JSON.parse(projectKeywords);
         }
@@ -29,31 +34,34 @@
      * If a new tag input is registered, it is checked whether it is not empty, it does not exceed
      * the maximal tag length, the maximal number of tags is not reached or the tag name already
      * exists. In this case the tag is correct and is added to the tag input field.
+     * @param newTag - validates the newly created tag
+     * @returns - the validation result of the new tag
      */
-    function validateTags() {
+    function validateTags(newTag: string): ValidationResult {
+        let validationResult: ValidationResult = { success: true };
         if (tags && tags.length > 0) {
-            const newTag = tags[tags.length - 1];
             if (newTag.trim().length == 0) {
-                tags.pop();
-                return toast("Empty keywords are not allowed!");
-            } else if (newTag.trim().length > maximalTagLength) {
-                tags.pop();
-                return toast(`Maximal keyword length of ${maximalTagLength} characters exceeded!`);
-            } else if (tags.length > maximumAmountOfTags) {
-                tags.pop();
-                return toast(`Maximal amount of keywords reached!`);
+                validationResult = { success: false, error: "Empty keywords are not allowed!" };
+            } else if (newTag.trim().length > maximumTagLength) {
+                validationResult = { success: false, error: `Maximum keyword length of ${maximumTagLength} characters exceeded!` };
+            } else if (tags.length > maximumNumberOfTags) {
+                validationResult = { success: false, error: `Maximum number of keywords reached!` };
             } else if (
                 Array.from(tags)
                     .slice(0, tags.length - 1)
                     .includes(newTag)
             ) {
-                tags.pop();
-                return toast(`Keyword name already exists!`);
+                validationResult = { success: false, error: `Keyword name already exists!` };
             }
-            localStorage.setItem(`project_${projectId}_keywords`, JSON.stringify(tags));
-            return tags;
         }
+        return validationResult;
     }
+
+    $effect(() => {
+        localStorage.setItem(`project_${projectId}_keywords`, JSON.stringify(tags));
+    });
+
+
 </script>
 
 <!--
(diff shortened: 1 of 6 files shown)


src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+@component
+Component for adding tags. The component includes a text input field to enter the name of the tags.
+On pressing the enter button a tag gets created. Each tag has a button to remove the tag, this
+button appears if the user hovers over the tag. Both, the maximal amount of tags and the maximal
+length of a single tag are customizable.
+
+Usage:
+```svelte
+    <SettingsSection {maximumAmountOfTags} {maximalTagLength} />
+-->
```
Slartibartfass2: the closing `\`` is missing here

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+                allTagsWrapperClasses="flex flex-row items-center gap-x-2 gap-y-2 flex-wrap"
+                inputClasses="focus:outline-none focus:ring-0 pl-2 py-1"
+                labelText="Define keywords that are highlighted in the abstract of a paper in the review mode."
+                removeTagButtonClasses="max-w-0 overflow-hidden group-hover:max-w-[20px] group-hover:ml-2 transition-all duration-200 text-black-500 hover:text-black-500 cursor-pointer"
+                showLabel={true}
+                tagWrapperClasses="group flex items-start bg-gray-200 text-black px-2 py-1.5 rounded-xl"
+                tagsInputWrapperClasses="flex items-start border-2 border-gray-200 py-2 px-3 rounded-md mt-2 h-[84px] overflow-auto"
+                bind:tags
+                on:input={() => validateTags()}
+            />
```
Slartibartfass2: I'm not sure if we should use this library as we have no control over it.
When I tested it, some errors occurred. @Merseleo already implemented an input for this, called `ChipsInput`.
Let's use this instead.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+                return toast(`Maximal amount of keywords reached!`);
+            } else if (
+                Array.from(tags)
+                    .slice(0, tags.length - 1)
+                    .includes(newTag)
+            ) {
+                tags.pop();
+                return toast(`Keyword name already exists!`);
+            }
+            localStorage.setItem(`project_${projectId}_keywords`, JSON.stringify(tags));
```
Slartibartfass2: We only update the storage when we validate the tags, but when removing the last tag, the storage isn't updated. How about using `onEffect` for this?

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+        }
+    });
+
+    /**
+     * Validates the input for a tag. If the input is incorrect, the tag is not created.
+     * If a new tag input is registered, it is checked whether it is not empty, it does not exceed
+     * the maximal tag length, the maximal number of tags is not reached or the tag name already
+     * exists. In this case the tag is correct and is added to the tag input field.
+     */
+    function validateTags() {
```
Slartibartfass2: Why do we return anything here? The return value isn't used as far as I can tell.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+    const { projectId, maximumAmountOfTags = 25, maximalTagLength = 15 }: Props = $props();
+
+    let tags: string[] = $state([]);
+
+    /**
+     * Loads the already defined keyword tags for the selected project.
+     */
+    onMount(async () => {
+        const projectKeywords = localStorage.getItem(`project_${projectId}_keywords`);
```
Slartibartfass2: Let's export the key creation for the local storage so that it can be reused for fetching and updating the storage.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+button appears if the user hovers over the tag. Both, the maximal amount of tags and the maximal
+length of a single tag are customizable.
+
+Usage:
+```svelte
+    <SettingsSection {maximumAmountOfTags} {maximalTagLength} />
+-->
+<SettingsSection sectionTitle="Keywords">
+    <div class="items-top space-x-2">
+        <div class="grid max-w-2xl gap-1.5 pt-1 leading-none">
```
Slartibartfass2: It seems to me that most of these class attributes are either unnecessary or unused. What are the reasons for them and why are there two divs?

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
+<script lang="ts">
+    import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
+    import TagsInput from "svelte-unstyled-tags";
+    import { toast } from "svelte-sonner";
+    import { onMount } from "svelte";
+
+    interface Props {
+        projectId: string;
+        maximumAmountOfTags: number;
+        maximalTagLength: number;
```
Slartibartfass2: "maximal" is German", choose "maximum" instead.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
+<script lang="ts">
+    import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
+    import TagsInput from "svelte-unstyled-tags";
+    import { toast } from "svelte-sonner";
+    import { onMount } from "svelte";
+
+    interface Props {
+        projectId: string;
+        maximumAmountOfTags: number;
```
Slartibartfass2: "amount" is always for things that cannot be assigned to single units, e.g., liquids. For something that can be distinguished into units, we use "number".
```suggestion
        maximumNumberOfTags: number;
```

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+        maximalTagLength: number;
+    }
+
+    const { projectId, maximumAmountOfTags = 25, maximalTagLength = 15 }: Props = $props();
```
Slartibartfass2: Let's make these variables constants in this component, instead of props. Later, we probably use env variables for this. Until then, let's use a bit higher values, e.g., 50 keywords with a maximum length of 100.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,86 @@
...
+<!--
+@component
+Component for adding tags. The component includes a text input field to enter the name of the tags.
+On pressing the enter button a tag gets created. Each tag has a button to remove the tag, this
+button appears if the user hovers over the tag. Both, the maximal amount of tags and the maximal
+length of a single tag are customizable.
+
+Usage:
+```svelte
+    <SettingsSection {maximumAmountOfTags} {maximalTagLength} />
```
Slartibartfass2:
```suggestion
    <KeywordSettings {maximumAmountOfTags} {maximalTagLength} />
```

Slartibartfass2 (changes requested):
This looks great so far. Let's agree on the input component.
After this, I will review the tests, which may change if we change the component.

Good work!

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: True

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Ok I didn't see that. Well, than this is definitely the better option ...

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Ah ok.. Now I get the usage of the effect rune. Slowly, but steadily improving. That's a much better way to do that

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Should be fixed now

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Good point

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: True. I just copied it in the beginning from another component, but it is unnecessary.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: True

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Done

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Ok. I wondered how we would handle env variables. But this explaines it, so I will change it to const

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: True

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,89 @@
...
+<!--
+@component
+Component for adding tags. The component includes a text input field to enter the name of the tags.
+On pressing the enter button a tag gets created. Each tag has a button to remove the tag, this
+button appears if the user hovers over the tag. Both, the maximal amount of tags and the maximal
+length of a single tag are customizable.
+
+Usage:
+```svelte
+    <KeywordSettings {maximumAmountOfTags} {maximalTagLength} />
```
Slartibartfass2:
```suggestion
    <KeywordSettings {projectId} />
```

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,89 @@
...
+            } else if (
+                Array.from(tags)
+                    .slice(0, tags.length - 1)
+                    .includes(newTag)
+            ) {
+                validationResult = { success: false, error: `Keyword name already exists!` };
+            }
+        }
+        return validationResult;
+    }
```
Slartibartfass2: What are your thoughts on using guard clauses to reduce nesting.
```suggestion
    function validateTag(newTag: string): ValidationResult {
        if (tags.length >= maximumNumberOfTags) {
            return { success: false, error: `Maximum number of keywords reached!` };
        }

        const newTagLength = newTag.trim().length;
        if (newTagLength === 0) {
            return { success: false, error: "Blank keywords are not allowed!" };
        }

        if (newTagLength > maximumTagLength) {
            return {
                success: false,
                error: `Maximum keyword length of ${maximumTagLength} characters exceeded!`,
            };
        }

        const alreadyExistingTags = Array.from(tags).slice(0, tags.length - 1);
        if (alreadyExistingTags.includes(newTag)) {
            return { success: false, error: `Keyword name already exists!` };
        }

        return { success: true };
    }
```

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
Slartibartfass2: I thought of something like this
```ts
function getProjectKeywordsKey(projectId: string) {
    return `project_${projectId}_keywords`;
}
```
This way, you can use it for `localStorage.getItem` and `localStorage.setItem`.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,89 @@
...
+    import { onMount } from "svelte";
+    import ChipsInput from "$lib/components/composites/input/ChipsInput.svelte";
+    import type { ValidationResult } from "$lib/model/general";
+
+    interface Props {
+        projectId: string;
+    }
+
+    const maximumNumberOfTags = 50;
+    const maximumTagLength = 100;
```
Slartibartfass2: Currently, we write constants using SCREAMING_SNAKE_CASE. It's up to you if you abbreviate maximum or not.

tests/e2e/pom/project-review-settings-model.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Adds a tag with the given name.
+     *
+     * @param tagName - the new tag name
+     */
+    async addTag(tagName: string) {
+        await this.tagInputField.click();
+        await this.tagInputField.fill(tagName);
+        await this.tagInputField.focus();
```
Slartibartfass2: Is there a reason why we focus here? The button is focused when we click on it.

tests/e2e/pom/project-review-settings-model.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * Deletes the tag with the given name.
+     *
+     * @param tagName - tag name
+     */
+    async deleteTag(tagName: string) {
+        const tag = this.page.getByText(tagName);
+        await tag.hover();
```
Slartibartfass2: Why do we need to hover the tag?

tests/e2e/project-review-settings.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        mockBackendService
+            .createProject({ name: "Project 1" })
+            .then((project) => (projectId = project.response.id));
+    });
+
+    test.beforeEach(async ({ page }) => {
+        await page.goto(`/project/${projectId}/settings/review`);
+    });
+
+    test("When the user creates a new tag and deletes this tag, the tag is correctly added and deleted", async ({
```
Slartibartfass2:
```suggestion
    test("When the user creates a new tag and deletes this tag, then the tag is correctly added and deleted", async ({
```

tests/e2e/project-review-settings.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        projectReviewSettingsPage,
+    }) => {
+        await projectReviewSettingsPage.addTag("New Tag");
+        const newTag = page.getByText("New Tag");
+        await expect(newTag).toBeVisible();
+        await projectReviewSettingsPage.deleteTag("New Tag");
+        await expect(newTag).not.toBeVisible();
+    });
+
+    test("When the navigates to another page and comes back to the review settings, the previously defined keyword tags are still available", async ({
```
Slartibartfass2:
```suggestion
    test("When the user navigates to another page and comes back to the review settings, then the previously defined keyword tags are still available", async ({
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+            },
+        });
+    });
+
+    test("It is rendered correctly", async () => {
+        expect(screen.queryByText("Keywords")).toBeInTheDocument();
+        expect(screen.getByPlaceholderText("Add keyword")).toBeInTheDocument();
+    });
+
+    test("When the input is valid, a tag is created", async () => {
```
Slartibartfass2:
```suggestion
    test("When the input is valid, then a tag is created", async () => {
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        const tagsInputField = screen.getByLabelText(
+            "Define keywords that are highlighted in the abstract of a paper in the review mode.",
+        );
+        await userEvent.type(tagsInputField, "New Tag 1");
+        await userEvent.keyboard("{Enter}");
+        expect(screen.getByText("New Tag 1")).toBeInTheDocument();
+    });
+
+    test("When the tag name already exists, the name is to long or empty, then no tag is created", async () => {
```
Slartibartfass2:
```suggestion
    test("When the tag name already exists, the name is too long or blank, then no tag is created", async () => {
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        expect(screen.getAllByText("New Tag 1").length).toBe(1);
+
+        await userEvent.type(tagsInputField, " ");
+        await userEvent.keyboard("{Enter}");
+        expect(screen.queryByText(" ")).not.toBeInTheDocument();
+
+        const toLongString = "a".repeat(maximumTagLength);
+        await userEvent.type(tagsInputField, toLongString);
+        await userEvent.keyboard("{Enter}");
+        expect(screen.queryByText(toLongString)).not.toBeInTheDocument();
```
Slartibartfass2, line 53:
```suggestion
        const tooLongString = "a".repeat(maximumTagLength);
        await userEvent.type(tagsInputField, tooLongString);
        await userEvent.keyboard("{Enter}");
        expect(screen.queryByText(tooLongString)).not.toBeInTheDocument();
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    test("When the tag remove button is clicked, the according tag gets removed correctly", async () => {
```
Slartibartfass2:
```suggestion
    test("When the tag remove button is clicked, then the according tag gets removed correctly", async () => {
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        await userEvent.type(tagsInputField, "Tag to remove");
+        await userEvent.keyboard("{Enter}");
+        const tagToRemove = screen.getByText("Tag to remove");
+        expect(tagToRemove).toBeInTheDocument();
+        expect(screen.getByRole("button", { name: "×" })).toBeInTheDocument();
+        await userEvent.click(screen.getByRole("button", { name: "×" }));
+        expect(tagToRemove).not.toBeInTheDocument();
+    });
+
+    test("When the maximal amount of tags is reached, no other tag is created", async () => {
```
Slartibartfass2:
```suggestion
    test("When the maximum amount of tags is reached, then no other tag is created", async () => {
```

tests/integration/settings/project-settings/review/keyword-settings.test.ts
```diff
@@ -0,0 +1,80 @@
...
+            "Define keywords that are highlighted in the abstract of a paper in the review mode.",
+        );
+        for (let i = 0; i < maximumAmountOfTags; i++) {
+            await userEvent.type(tagsInputField, `New Tag ${i}`);
+            await userEvent.keyboard("{Enter}");
+            expect(screen.getByText(`New Tag ${i}`)).toBeInTheDocument();
+        }
+        await userEvent.type(tagsInputField, `New Tag ${maximumAmountOfTags}`);
+        await userEvent.keyboard("{Enter}");
+        expect(screen.queryByText(`New Tag ${maximumAmountOfTags}`)).not.toBeInTheDocument();
```
Slartibartfass2, line 86: Also check for the error message

tests/integration/settings/project-settings/review/keyword-settings.test.ts
Slartibartfass2: Also check for the error messages

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+Usage:
+```svelte
+    <KeywordSettings {maximumAmountOfTags} {maximalTagLength} />
+```
+-->
+<SettingsSection sectionTitle="Keywords">
+    <div class="max-w-2xl">
+        <ChipsInput
+            label="Define keywords that are highlighted in the abstract of a paper in the review mode."
```
Slartibartfass2: What are your thoughts on this? It sounds a bit better imo.
```suggestion
            label="Define keywords that are highlighted in the abstract of a paper when the review mode is activated."
```

Slartibartfass2 (changes requested):
We've come far, let's finish this and then wait until the new API version is integrated.

Great work!

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Done.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Maybe yes. The user may understand it better. Idm

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Looks like a prettier way to do that

tests/e2e/pom/project-review-settings-model.ts
domienderle: No I just forgot to delete it. Only focus is not enough, but with click it is unnecessary.

tests/e2e/pom/project-review-settings-model.ts
domienderle: No, it's survived from a previous test version..

tests/integration/settings/project-settings/review/keyword-settings.test.ts
domienderle: Done.

tests/e2e/project-review-settings.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        page,
+        projectReviewSettingsPage,
+    }) => {
+        await projectReviewSettingsPage.addTag("New Tag 1");
+        await expect(page.getByText("New Tag 1")).toBeVisible();
+        await projectReviewSettingsPage.addTag("New Tag 2");
+        await expect(page.getByText("New Tag 2")).toBeVisible();
+
+        await page.getByText("General").click();
+        await page.getByTestId("settings-tab-review").click();
```
Slartibartfass2: Use a locator, which is accessible to the user.

Slartibartfass2 (changes requested):
Great work, we're nearly done here.
There's one conversation left open, please answer it.
After that, rebase and then I will have a last look.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
domienderle: Now I got it. Yes, we can also do that.

tests/e2e/project-review-settings.test.ts
domienderle: I thought this would not work, because there are multiple elements containing the text "Review", but now it works.

domienderle: Thank you for your review and the time.
``` await this.tagInputField.waitFor({ state: "visible" });``` and ```await this.tagInputField.focus();``` are necessary, otherwise the e2e tests in chromium are flaky.

src/lib/components/composites/settings/project-settings/review/KeywordSettings.svelte
Slartibartfass2: Now you only have to reuse it for `localStorage.setItem`. Otherwise, it's obsolete.

tests/e2e/project-review-settings.test.ts
Slartibartfass2: That's fine.

Slartibartfass2 (dismissed):
Great!

Slartibartfass2 (approved):
🚀

PR #338, opened 2025-04-23
Update API submodule to version 0.2.0

### LINKED ISSUE #323: [Feature]: Integrate snowballR API v0.2.0

_No response_

At the moment we are missing the changes from the new API version 0.2.0, e.g. the local project paper id.

Update the submodule for the API to the new version and implement the new changes.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #323 

## What I have made

- I updated the git submodule for the snowballr API to version 0.2.0
- I re-generated the the gRPC client and implemented the missing functions and 
- I adapted the code using old, changed API calls and make use of the new / adapted calls

## Checklist

- [x] I have adapted the tests
  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests


src/lib/components/composites/paper-components/PaperInfo.svelte
```diff
@@ -8,19 +8,31 @@
...
     };
 
-    const { loadingPaper, class: className }: Props = $props();
+    const { loadingPaper, loadingPaperId = undefined, class: className }: Props = $props();
 </script>
 
-{#await loadingPaper}
+<!--
+@component
+Container for displaying basic paper information as
```
Slartibartfass2:
```suggestion
Container for displaying basic paper information such as:
```

src/lib/utils/common-helper.ts
```diff
@@ -134,22 +136,24 @@ function getStatusColor(status: PaperStatus, prefix: ColorPrefix = "text"): stri
...
+ * Maps the paper decision to the corresponding text.
  *
- * @param paper - the paper whose review status should be determined
+ * @param paperDecision - the paper decision
  * @returns the review status ("Accepted", "Declined", "Undecided" or "Not reviewed")
  */
-function getStatusText(paper: Project_Paper): PaperStatus {
-    switch (paper.decision) {
+function getStatusText(paperDecision: PaperDecision): PaperStatus {
+    switch (paperDecision) {
```
Slartibartfass2, line 153: Currently, we always use `getStatusText` to get the results of `getStatusColor`, e.g., `getStatusColor(getStatusText(decision))`. Do you think we can use the decision directly in `getStatusColor`?

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -59,6 +59,11 @@
 
     const loadingPaper = loadingPaperWrapper.then(asPaper);
     const loadingPaperId = loadingPaper.then((paper) => paper.id);
+    // as the navigation bar shows either the paper id or the local / relative id, if the paper
+    // is a project paper, the id for the navigation bar must be handles differently
```
Slartibartfass2:
```suggestion
    // is a project paper, the id for the navigation bar must be handled differently
```

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -11,24 +11,29 @@
...
+        } else {
+            // navigate to paper
+            goto(`/paper/${paperId}`);
+        }
     };
 
     const { paper, projectId, onClick = navigateToPaperView }: PaperListEntryProps = $props();
 
+    const paperId = $derived(
+        isProjectPaper(paper) ? asProjectPaper(paper)!.localId : asPaper(paper).id,
```
Slartibartfass2: The same logic is in `PaperView`. Let's extract this logic and document the reason for it.

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -1,38 +1,47 @@
...
+            cause: "NoDate",
+        });
+    }
+    const startDate = new Date(Number(projectInformation.creationDate.seconds) * 1000);
+    if (!projectInformation.lastStageStarted) {
+        throw new Error("Couldn't load start date of latest stage.", {
+            cause: "NoDate",
+        });
+    }
+    const startDateForStage = new Date(Number(projectInformation.lastStageStarted.seconds) * 1000);
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]: These two checks are almost the same. How about making it a function?
```ts
function parseTimestamp(timestamp?: Timestamp, errorMessage: string = ""): Date {
    if (!timestamp) {
        throw new Error(errorMessage, { cause: "NoDate" });
    }

    return new Date(Number(timestamp.seconds) * 1000);
}
```

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -1,38 +1,47 @@
...
+    const daysInStage = Math.round(
+        Math.abs(new Date().getTime() - startDateForStage.getTime()) / 8.64e7,
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]:
```suggestion
        Math.abs(Date.now() - startDateForStage.getTime()) / 8.64e7,
```
Additionally, please consider using a constant for that magic number.

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -1,38 +1,47 @@
...
+        (acc, currentValue) =>
+            [PaperDecision.IN_REVIEW, PaperDecision.UNREVIEWED].includes(currentValue.decision)
+                ? acc + Number(currentValue.count)
+                : acc,
+        0,
+    );
+    const numberOfPapers = decisionStatistics.statistics.reduce(
+        (acc, currentValue) => acc + Number(currentValue.count),
         0,
     );
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts:58]: Optional: These could be replaced by a sum function

Slartibartfass2 (changes requested):
Only some major changes left.

Currently, on the homepage and on the project dashboard the papers all have the ID 0.
This should be fixed.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts:58]: I know that it could be done with a sum, but imo this solution is very readable. Or what solution with a sum function do you suggest? Because I do not know any that is less complicated, but my javascript knowledge is not best.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: Good idea, I changed it.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: Good idea, if we plan to change the cause of something else. I will implement your suggestion.

src/lib/utils/common-helper.ts
Merseleo: You are right, actually it is unnecessary to not use the paper decision directly in the `getStatusColor` function. I replaced the parameter by the paper decision. Did I miss anything?

src/lib/components/composites/paper-components/PaperListEntry.svelte
Merseleo: Actually it is not the same, but I will extract some logic and add a comment.

Merseleo (commented):
> Only some major changes left.
> 
> Currently, on the homepage and on the project dashboard the papers all have the ID 0. This should be fixed.

I added a fix in the mock backend. Thank you for the quick response and good ideas and suggestions.

src/routes/project/[projectId]/dashboard/+page.ts
```diff
@@ -1,38 +1,64 @@
...
+        .response;
+    if (!projectInformation.creationDate) {
+        throw new Error("Couldn't load start date of project.", {
+            cause: "NoDate",
+        });
+    }
+    const startDate = parseTimestamp(
+        projectInformation.creationDate,
+        "Couldn't load start date of project.",
+    );
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/dashboard/+page.ts]: Now that we use the parsing method, we don't require the check.
```suggestion
    const projectInformation = await backendService.getProjectInformation({ projectId: project.id })
        .response;
    const startDate = parseTimestamp(
        projectInformation.creationDate,
        "Couldn't load start date of project.",
    );
```

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/dashboard/+page.ts:58]: That's fine for me.

Slartibartfass2 (changes requested):
Just this small change left.

src/routes/project/[projectId]/dashboard/+page.ts
Comment (Merseleo) [reply src/routes/project/[projectId]/dashboard/+page.ts]: I am so lost, i forgot to delete it.

Slartibartfass2 (approved):

PR #351, opened 2025-05-02
Add/remove paper from reading list

### LINKED ISSUE #99: [Use Case]: Add Paper To Reading List

### LINKED ISSUE #100: [Use Case]: Remove Paper From Reading List

mowi12:
Closes #99
Closes #100 

## What I have made

- I added the basic functionality to the `PaperBookmarkButton`.
  - Based on this logic, I updated corresponding component accordingly

## Checklist

  - [ ] ~~unit tests~~ no new component added
  - [x] integration tests
  - [x] end-to-end tests


src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
+     * Uses a cleanup function to handle cases where the paper ID changes before the fetch completes.
      */
-    function toggleBookmarkState() {
-        if (isBookmarked) {
-            removePaperFromReadingList();
+    $effect(() => {
+        const currentPaperId = paperId.value;
+        let cancelled = false;
+
+        if (typeof currentPaperId === "string") {
```
Merseleo:
```suggestion
        if (currentPaperId) {
```
Why do you want to check the type of the paperId.value? It is either a string or undefined and so use can use normal truthy checks?

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
+                .catch((error) => {
+                    if (cancelled) return;
+
+                    console.error(`Failed to fetch bookmark status for ${currentPaperId}:`, error);
+                })
+                .finally(() => {
+                    if (cancelled) return;
+
+                    isUpdatingBookmarkStatus = false;
+                    // If cancelled, isLoadingStatus should already be false from cleanup
```
Merseleo:
```suggestion
                    // If cancelled, `isUpdatingBookmarkStatus` should already be false from cleanup
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
         }),
     );
 
+    let isUpdatingBookmarkStatus = $state(false);
+
     const onMouseEnter = () => (isHovered = true);
     const onMouseLeave = () => (isHovered = false);
 
     /**
-     * Adds the paper to the reading list, if it is not added yet, otherwise removes it.
```
Merseleo, line 30: Why did you delete this comment completely? Now `toggleBookmarkStatus` is not documented anymore.

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
-    }
+    function toggleBookmarkStatus() {
+        if (isUpdatingBookmarkStatus) return;
+        isUpdatingBookmarkStatus = true;
 
-    onMount(() => {
-        if (isBookmarkedDefault) {
-            addPaperToReadingList();
+        if (paperId.value === undefined) {
+            console.error("Paper ID is undefined");
```
Merseleo: In general I thought we agreed on writing "Id" instead of "ID"? Applies for some other comments as well.

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
+                });
+        } else {
+            backendService
+                .addPaperToReadingList({ id: paperId.value })
+                .response.then(() => {
+                    isBookmarked = true;
+                    onPaperChangedBookmarkStatus?.();
+                })
+                .catch((error) => {
+                    console.error("Error adding paper to reading list", error);
```
Merseleo:
```suggestion
                    console.error("Error removing paper from reading list:", error);
                })
                .finally(() => {
                    isUpdatingBookmarkStatus = false;
                });
        } else {
            backendService
                .addPaperToReadingList({ id: paperId.value })
                .response.then(() => {
                    isBookmarked = true;
                    onPaperChangedBookmarkStatus?.();
                })
                .catch((error) => {
                    console.error("Error adding paper to reading list:", error);
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
 <!--
 @component
 Button to add a paper to or remove a paper from the reading list.
 
 This component will handle the API calls itself.
-According to bookmark state of the paper (isBookmarked), the button will change its appearance.
-
-By setting the `isBookmarkedDefault` property, the default state for the button can be set, i.e.
-if set to true, the paper with the given id is added to the reading list per default.
+If isBookmarkedDefault is set, it will show the initial state of the bookmark button.
```
Merseleo:
```suggestion
If `isBookmarkedDefault` is set, it will show the initial state of the bookmark button.
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
+After the initial state is set, the component will fetch the bookmark status of the paper from the backend.
 
 Usage:
 ```svelte
-<PaperBookmarkButton loadingPaperId={Promise.resolve("42")} isBookmarkedDefault={true} />
+    <PaperBookmarkButton
+        loadingPaperId={Promise.resolve(paperId)}
+        isBookmarkedDefault={true}
+        onPaperChangedBookmarkStatus={onPaperChangedBookmarkStatus}
```
Merseleo: Maybe I could be better to provide a dummy function as example as logging something or so? Otherwise you would use `{onPaperChangedBookmarkStatus}` afaik.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -31,9 +36,13 @@ Furthermore this component is clickable and navigates to the corresponding paper
 if the onClick() event handler is not overridden. Otherwise it executes the custom event handler
 on a single click. A double click always causes the navigation to the paper view.
 
+In addition, in order to propagate what happens
+when the bookmark status of the paper changes, the component accepts a callback function
+onPaperChangedBookmarkStatus() which is executed when the bookmark status changes.
```
Merseleo: This sounds a bit weird to me. 

Something likes this: "In addition, it is possible to provide a callback function `onPaperChangedBookmarkStatus()`
that is propagated to the `PaperBookmarkButton` and defines the behavior 
what should happen when the bookmark status of the paper changes.
This callback is executed when the bookmark status changes." Or do you have any other ideas?

src/routes/readinglist/helper.ts
Merseleo, line 1: Good idea, but why do you dont reuse this code in the `+page.ts`?

src/routes/readinglist/+page.svelte
Merseleo, line 1: Why do you use this concept? Couldn't you just invalidate the load data in the `+page.ts` and derive the filterReadingList from the `loadingReadingList`?

tests/e2e/fixtures/reading-list-page-fixture.ts
```diff
@@ -11,6 +12,10 @@ type ReadingListPage = {
  */
 export const test = base.extend<ReadingListPage>({
     readingListPage: async ({ page }, use) => {
+        await page.goto("/");
+        await page.waitForLoadState("networkidle");
```
Merseleo: Please update this to not use `waitForLoadState`.

tests/e2e/pom/home-page-model.ts
```diff
@@ -20,11 +20,22 @@ export class DevHomePage {
...
-        await this.page.getByRole("link", { name: "Settings" }).click();
+        await this.page.getByRole("link", { name: linkName }).click();
+    }
+
+    /**
+     * Opens the user settings page and clicks on a specific setting.
+     *
+     * @param settingName - The name of the setting to be clicked on the user settings page.
+     */
+    async openUserSetting(settingName: string) {
```
Merseleo: To make this more robust: what do you think of make `linkName` and `settingName` not of type string but of a union type that limit the possible names to the settings or links we have?

tests/e2e/pom/home-page-model.ts
```diff
@@ -20,11 +20,22 @@ export class DevHomePage {
...
+    async openLinkInUserMenuDialog(linkName: string) {
         await this.page.getByRole("button", { name: /[A-Z]{2}/ }).click();
-        await this.page.getByRole("link", { name: "Settings" }).click();
+        await this.page.getByRole("link", { name: linkName }).click();
+    }
+
+    /**
+     * Opens the user settings page and clicks on a specific setting.
+     *
+     * @param settingName - The name of the setting to be clicked on the user settings page.
```
Merseleo:
```suggestion
     * Opens the user settings page and clicks on a specific settings tab, e.g. "Account".
     *
     * @param settingName - The name of the settings tab to be clicked on the user settings page.
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -23,57 +28,115 @@
...
 
     /**
-     * Adds the paper to the reading list, if it is not added yet, otherwise removes it.
+     * Fetches the actual bookmark status when the paper ID becomes available.
+     * Uses a cleanup function to handle cases where the paper ID changes before the fetch completes.
      */
-    function toggleBookmarkState() {
-        if (isBookmarked) {
-            removePaperFromReadingList();
+    $effect(() => {
```
Merseleo: Only for interest: why did you make this an effect and not onMount? Where is it possible that this state while the component is mounted? On the reading list page when you remove an entry?

tests/e2e/pom/reading-list-page-model.ts
```diff
@@ -1,26 +1,50 @@
 import { expect, type Locator, type Page } from "@playwright/test";
 
+export const PREDICATBLE_PAPER_TITLE_PREFIX = "Test Paper";
```
Merseleo: I think this a spelling mistake: It should be `PREDICTABLE` and not `PREDICATBLE`

tests/e2e/pom/reading-list-page-model.ts
```diff
@@ -1,26 +1,50 @@
...
+     */
+    async openPaperListEntry(index: number) {
+        const entry = this.readingListEntries.filter({
+            has: this.page.getByRole("heading", {
+                name: `${PREDICATBLE_PAPER_TITLE_PREFIX} ${index}`,
+                exact: true,
+            }),
         });
+        await entry.scrollIntoViewIfNeeded();
+        await entry.click();
```
Merseleo: Clever solution, great work (close on your own).

tests/e2e/pom/reading-list-page-model.ts
```diff
@@ -1,26 +1,50 @@
...
         this.readingListEntries = page.getByRole("listitem");
         this.searchBarInput = page.getByPlaceholder("Search");
-        this.readingListEntry0 = page.getByRole("listitem").filter({
-            has: page.getByRole("heading", { name: "Paper 0 on reading list", exact: true }),
+    }
+
+    /**
+     * Opens the paper list entry in the reading list.
+     *
+     * @param index - the index in the name of the paper in the reading list
```
Merseleo: Maybe add to this comment in which range the indices are? (same for second method below)

tests/integration/paper-components/paper-view/paper-view.test.ts
```diff
@@ -11,8 +11,13 @@ import {
 } from "../../../model-builder";
 import { reviewMode } from "$lib/global-state/review-mode-state.svelte";
 import { waitForComponentLoading } from "$tests/integration/test-helper";
+import { mockApiCall } from "$tests/setupTest";
 
 describe("PaperView", () => {
+    beforeAll(() => {
+        mockApiCall("isPaperOnReadingList", { value: false });
+    });
```
Merseleo: Cannot you simplify this and set this default return of false in the `setupTest.ts`? Because imo it is ok to have some default implementation there and returning false is a default, especially since you set it here without using it in this test? (same for `reading-list-entry.test.ts`)

tests/integration/paper-bookmark-button.test.ts
```diff
@@ -18,7 +20,8 @@ describe("PaperBookmarkButton", () => {
         expect(button).toHaveAttribute("aria-label", "Add to reading list");
     });
 
-    test("When paper is bookmarked, then bookmark button has 'Remove from reading list' tooltip", () => {
+    test("When the paper is bookmarked, then bookmark button has 'Remove from reading list' tooltip", () => {
```
Merseleo: You add an article for "paper" but not for bookmark button or the tooltip? Any specific reason?

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
 import { createPaper } from "$tests/model-builder";
 import type { Project } from "$lib/model/api/project";
+import {
+    EXTRA_PAPER_TITLE,
+    NUM_PAPERS_DEFAULT,
+    PREDICATBLE_PAPER_TITLE_PREFIX,
+} from "./pom/reading-list-page-model";
 
-test.describe("Add, view or remove papers to / from the reading list", () => {
+test.describe("Reading List Functionality", () => {
```
Merseleo, line 14: Whats wrong with my name ;) (just kidding)

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
-    const paperIds: string[] = [];
-
-    const NUMBER_OF_PAPERS = 11;
+    let defaultPaperIds: string[] = [];
+    let extraPaperId: string = "";
 
     test.beforeAll(async ({ mockBackendService }) => {
         const project: Project = await mockBackendService.createProject({
-            name: "Project reading list",
+            name: "Test Project",
```
Merseleo: The reason why I used this name, is that this way the project names are more likely unique. If you call it "Test Project" and some else in another test as well, one project cannot be created due to the same name. So do you really the name "Test Project"?

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
+                currentList.papers.map((p) =>
+                    mockBackendService.removePaperFromReadingList({ id: p.id }),
+                ),
+            );
+        }
+
+        await Promise.all(
+            defaultPaperIds.map((id) => mockBackendService.addPaperToReadingList({ id })),
+        );
+
```
Merseleo: Is this really necessary because it seems like a lot of overhead and besides ignores the extra paper id?

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
+                page.getByText(`${PREDICATBLE_PAPER_TITLE_PREFIX} ${i}`, { exact: true }),
+            ).toBeVisible();
+        }
+
+        // Check that the extra paper (not added in beforeEach) is NOT visible
+        await expect(page.getByText(EXTRA_PAPER_TITLE, { exact: true })).toBeHidden();
     });
 
-    test("When the user searches in the reading list, then only papers matching the search input are shown.", async ({
+    test("Filters papers based on search input", async ({ page, readingListPage }) => {
```
Merseleo: This ignores the the "When ... then ..." pattern.

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
+        await readingListPage.searchBarInput.fill(PREDICATBLE_PAPER_TITLE_PREFIX);
+        await readingListPage.expectNumberOfEntries(NUM_PAPERS_DEFAULT);
+
+        // Search for something not present
+        await readingListPage.searchBarInput.fill("NonExistentSearchTerm");
+        await readingListPage.expectNumberOfEntries(0);
+        await expect(page.getByText("Your reading list is empty.")).toBeVisible(); // Adjust text if needed
+
+        // Clear search
+        await readingListPage.searchBarInput.press("Escape"); // Or clear button if available
```
Merseleo: Why the comments after the statements?

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
+        // Check content on the paper page
+        await expect(
+            page.getByRole("navigation").getByText(expectedTitle, { exact: false }),
+        ).toBeVisible();
+    });
 
-        // clear search
-        await readingListPage.searchBarInput.press("Escape");
-        await readingListPage.expectNumberOfEntries(NUMBER_OF_PAPERS);
+    test("When the user removes a paper from the reading list, then this paper will not be displyed anymore.", async ({
```
Merseleo:
```suggestion
    test("When the user removes a paper from the reading list, then this paper will not be displayed anymore.", async ({
```

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,198 @@
...
+        const paperIdToRemove = defaultPaperIds[paperIndexToRemove];
+        const titleToRemove = `${PREDICATBLE_PAPER_TITLE_PREFIX} ${paperIndexToRemove}`;
 
-        await readingListPage.openPaperListEntry();
+        await readingListPage.expectNumberOfEntries(NUM_PAPERS_DEFAULT);
 
-        // paper view was opened
+        // Go to the paper's page
+        await page.goto(`/paper/${paperIdToRemove}`);
+        await page.waitForLoadState("networkidle");
```
Merseleo: Please delete `waitForLoadState`

Merseleo (changes requested):
I tested it and found no error, even with slow response time of the backend everything seems to work correctly, great job! However, I think there are some implementation details that could be improved, I am looking forward to your comments and reactions.

I know this was not part of the issues you implemented but what do you think of displaying "No search results" instead of "Empty reading list" when to search something and no matches could be found?

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: Yes you're right. A simple truthy check is sufficient and much cleaner!

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: You are completely right! I update all the comments.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
mowi12: You're right. This can be rephrased.

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: You're right, i could have just kept this comment for this function as the functionality is the same.

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: Yes you're right. I added a simply logging function.

src/routes/readinglist/helper.ts
mowi12: Good catch, I will definitely reuse this part of my code!

src/routes/readinglist/+page.svelte
mowi12: With this concept, I can work around the loading state in the reading list. Meaning, after I click a bookmark button in the reading list, removing a paper from the list, normally there would be a flickering due to the reading list updating. With this approach I simply update the list after the fetching, so the list is only updated after the fetching is complete.

tests/e2e/pom/home-page-model.ts
mowi12: Great idea! Take a look at my implementation if you think this is better.

tests/e2e/pom/reading-list-page-model.ts
mowi12: Yes you're right, good catch!

tests/e2e/pom/reading-list-page-model.ts
mowi12: Thanks!

tests/e2e/pom/reading-list-page-model.ts
mowi12: Yeah, why not.

tests/e2e/reading-list.test.ts
mowi12: It's a bit much on the nose, isn't it?

tests/e2e/reading-list.test.ts
mowi12: You're right again. This makes sense!
Edit: After running the tests again, I remembered why I changed it. The locator in the homepage.pom file searches for "Reading List" when trying to open the reading list but also finds this project name. But I will change it into something better.

tests/e2e/reading-list.test.ts
mowi12: You're right, this is a lot of overhead. But it ensures that each test starts with the same conditions and one doesn't need to restore the previous setup. I also think that the `beforeEach` is exactly made for this purpose.
And I do not think that this will make more than a few seconds of difference.

tests/e2e/reading-list.test.ts
mowi12: You're right. I will change it back.

tests/e2e/reading-list.test.ts
mowi12: Those are truly awful remnants of generated code ... these shouldn't be here.

tests/integration/paper-components/paper-view/paper-view.test.ts
mowi12: You're right.

tests/integration/paper-bookmark-button.test.ts
mowi12: No. Just bad grammar.

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: When you're in the reading list and click on a paper. Then the `PaperBookmarkButton` only updates if this is an `$effect` and not `mount`.

mowi12 (commented):
I'm happy that my code works and only some refactoring needed to be done.
I hope everything is still working and to your satisfaction.

tests/e2e/reading-list.test.ts
Merseleo: Yep, it is, I am ok with your improvement.

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: But then without spelling mistake ;) "otherweise" should only be "otherwise".

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: Great example. helps a lot imo.

However, if you name your function `onPaperChangedBookmarkStatus` as well you can use `{onPaperChangedBookmarkStatus}` and I think this would be even automatically formatted this way.

src/routes/readinglist/helper.ts
Merseleo: I think this could be even more simplified: 
```typescript
return {
        loadingReadingList: loadReadingList(),
    };
```

But you can also keep it as it is.

tests/e2e/pom/home-page-model.ts
Merseleo: Yep, I think this way it is safer to use, thank you for that work!

tests/e2e/pom/home-page-model.ts
Merseleo: Your solution is even better, great!

tests/e2e/pom/reading-list-page-model.ts
Merseleo: A catch from Webstorm like the other spelling mistakes ^^.

tests/integration/paper-bookmark-button.test.ts
Merseleo: Ok

tests/e2e/reading-list.test.ts
Merseleo: Ok 😂

tests/e2e/reading-list.test.ts
Merseleo: I understand your reasoning and would suggest a change. What do you think of moving this code into the reading list fixture (because there you can define code that runs automatically before each test afaik. Imo this would be a great place as this fixture simulates a correct reading list page and this should contain a consistent reading list?

tests/e2e/reading-list.test.ts
Merseleo: The locator on homepage.pom is this one, right: `await expect(this.page.getByRole("heading", { name: linkName })).toBeVisible();` Couldn't you just add `exact: true`? Then "Reading list project" should not be found anymore?! Because imo the name "Reading Testlist Project" is not better ^^.

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: Yeah i tested it. I do not quite understand why this component is mounted when I navigate to the paper view, but it seems it is and only effect solves the problem. Good work, my mistake.

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,76 +4,150 @@
...
-        // TODO: Will be implemented in #100
-        isBookmarked = false;
-        console.log(`Removed paper with id ${paperId.value} from reading list`);
-    }
+    /**
+     * Adds the paper to the reading list, if it is not already added yet, otherweise removes it.
+     */
+    function toggleBookmarkStatus() {
+        if (isUpdatingBookmarkStatus) return;
+        isUpdatingBookmarkStatus = true;
```
Merseleo: One last idea that came to my mind: what about disabling the tooltip while loading and not returning without any change? The `Tooltip` component has the `disabled` attribute and this could be set as following, didnt it: 
```typescript
aria-label={tooltipText}
disabled={isUpdatingBookmarkStatus}
onclick={toggleBookmarkStatus}
```

src/routes/readinglist/+page.svelte
Merseleo: I thought using invalidate prevent this flickering and only reloads the data and updates the list if necessary? So exactly what you do manually?

Merseleo (changes requested):
Ok despite of some open discussion on how to reload the data in the reading list, I think there are only minor changes left, really! Great job.

PS: Your E2E tests are failing again, so maybe not only a problem of our flaky tests?!

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: My bad ...

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: You're probably right. I changed it to another name instead of removing the part in front of the `=`.

src/routes/readinglist/helper.ts
mowi12: Ahh, thats how you do it.

I tried to do it like this:

```ts
return {
        loadReadingList(),
};
```

so dumb ...

src/routes/readinglist/+page.svelte
mowi12: With flickering, I mean, that there will be a short frame or to in which the skeletons will be visible.
And to prevent this, I only set the `filteredReadingList` after the `currentFullReadingList` is done fetching.
Therefore, no loading state will be seen.
I think Felix also used this in another part of the project. I think it is the removing of members in the project.
Normally the list would update with showing the skeletons while fetching.

tests/e2e/reading-list.test.ts
mowi12: The problem with exact is, that we still have the shortcut visible on the right of the text `Reading List`. So `exact` would not help us in this case.

This is just a very unfortunate case in the testing library, that there are two identical text parts which are both headings. And the name of the project doesn't really matter, does it?
Although I'll agree with you, this project name isn't great at all.

I'm open to suggestions.

tests/e2e/reading-list.test.ts
mowi12: Yes, good idea. Done.

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: Good idea, I don't think this will change much, but it definitely makes sense.

mowi12 (commented):
Soo, let me know what you think about the changes.

I also added a commit which fixes/refactors all e2e tests.
I know this shouldn't be here, but as this is probably the next PR going live, why not.
They should be a lot more stable now.

src/routes/readinglist/helper.ts
Merseleo: No you still return an object but what you tried only works when the key and value have the same name, which is not possible for functions afaik.

src/lib/components/composites/PaperBookmarkButton.svelte
Merseleo: If you add the `disabled` attribute you can delete the lines I marked as they can never be triggered. At least the if condition cannot be met.

tests/e2e/reading-list.test.ts
Merseleo: Ah ok, i understand the problem, didnt consider this! In my opinion your new name should be ok.

src/routes/readinglist/+page.svelte
Merseleo: I understand. I just looked at Felix code and you are right, he implemented it the same way. And probably you are right and it is cleaner if you do not see the skeletons again. So lets keep it as it is and finish this PR.

tests/e2e/fixtures/reading-list-page-fixture.ts
```diff
@@ -0,0 +1,42 @@
...
+    homepage: DevHomePage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - reading list page
+ */
+export const test = base.extend<{ readingListPageAndHomepagePage: ReadingListPageAndHomepagePage }>(
+    {
+        // Clear current version of reading list and add default papers
```
Merseleo: This comment could move up to the fixture comment as this comment is not visible if you hover test but only if you open this file. But this is not necessary, you can also close this conversation.

Merseleo (changes requested):
> I know this was not part of the issues you implemented but what do you think of displaying "No search results" instead of "Empty reading list" when to search something and no matches could be found?

What do you think? (you ignored this comment in the first round)

> I also added a commit which fixes/refactors all e2e tests. I know this shouldn't be here, but as this is probably the next PR going live, why not. They should be a lot more stable now.

Seems not as they are still failing. Do you know the reason?

But in general good work, I am glad this is done now (only one minor removal left).

src/lib/components/composites/PaperBookmarkButton.svelte
mowi12: You're right. This code can definitely be removed.

tests/e2e/fixtures/reading-list-page-fixture.ts
mowi12: No, you're right. This comment makes much more sense in the doc part above!

mowi12 (commented):
Perfect!

Regarding the change in the reading list. Generally i think this is a very good idea, but as this also would mean to update the e2e tests, i think this should be a new issue.
I really want this one to be finished as it already updates the e2e tests for everyone.

And regarding the tests. No, i have no f*ing idea why they do not work.

I will look into it, but until then you can look at the last to changes.

Merseleo (changes requested):
So the only thing left is that some error cases in the `PaperBookmarkButton` are not tested in the integration tests that are testable and should be tested imo.

mowi12 (commented):
With the help of @Slartibartfass2, I refactored the logic inside the paper bookmark button as you tried to tell me in the earlier messages.

I added some code in the integration tests which test if the given method has been called. This should remove some line from untested lines. As well as testing failed API calls.

I also added two other small fixes, all in separate commits, just how you like it!

tests/e2e/fixtures/reading-list-page-fixture.ts
```diff
@@ -0,0 +1,44 @@
...
+    readingListPage: DevReadingListPage;
+    homepage: DevHomePage;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - reading list page
+ *
+ * Clears current version of reading list and add default papers to keep a consistent
+ * starting state for each test.
```
Merseleo:
```suggestion
 * Extends the default **custom** fixture by providing the page object models for the
 * - reading list page and
 * - home page
 *
 * It navigates to the reading list before each test, clears the current version of the 
 * reading list and add default papers to keep a consistent
 * starting state for each test.
```

tests/integration/download-button.test.ts
```diff
@@ -23,7 +23,7 @@ describe("DownloadButton", () => {
         expect(tooltip).toBeInTheDocument();
     });
 
-    test("When the download button is clicked, then the download function is executed (currently signaled by an event dispatch)", async () => {
+    test("When the download button is clicked, then the download function is executed (currently signaled by an event dispatch", async () => {
```
Merseleo: I think this commit should be deleted or reverted: you have one opening bracket but no closing!

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,87 +4,140 @@
...
 
-    let isBookmarked = $state(isBookmarkedDefault);
-    let isHovered = $state(false);
-    const tooltipText = $derived(isBookmarked ? "Remove from reading list" : "Add to reading list");
-    const paperId = $derived(
-        resource<string, string | undefined>(loadingPaperId, {
-            initialValue: undefined,
-            resourceName: "paper ID",
-        }),
+    // `isUpdatingBookmarkStatus` is initialy set to `true` to represent the loading state
```
Merseleo:
```suggestion
    // `isUpdatingBookmarkStatus` is initially set to `true` to represent the loading state
```

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,199 @@
...
+export let defaultPaperIds: string[] = [];
 
-    const NUMBER_OF_PAPERS = 11;
+test.describe("Reading List Functionality", () => {
+    let projectId: string = "";
+    let extraPaperId: string = "";
 
+    /**
+     * Create a project and papers for the reading list.
+     * Do not add papers to the reading list, since this is done in the `beforeEach` hook.
```
Merseleo:
```suggestion
     * Do not add papers to the reading list, since they are provided by the reading list page fixture.
```

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,199 @@
...
     test("When the user opens the reading list, then all papers of the user's reading list are shown.", async ({
         page,
-        readingListPage,
+        readingListPageAndHomepagePage,
     }) => {
-        await readingListPage.expectNumberOfEntries(NUMBER_OF_PAPERS);
+        const { readingListPage } = readingListPageAndHomepagePage;
+        await readingListPage.expectNumberOfEntries(NUM_PAPERS_DEFAULT);
+
+        // Check for specific papers added in beforeEach
```
Merseleo:
```suggestion
        // Check for specific papers added before the test
```

tests/e2e/reading-list.test.ts
```diff
@@ -1,117 +1,199 @@
...
+        for (let i = 0; i < NUM_PAPERS_DEFAULT; i++) {
+            await expect(
+                page.getByText(`${PREDICTABLE_PAPER_TITLE_PREFIX} ${i}`, { exact: true }),
+            ).toBeVisible();
+        }
 
-        await expect(page.getByText(`Paper 0 on reading list`)).toBeVisible();
-        await expect(page.getByText(`Paper ${NUMBER_OF_PAPERS} on reading list`)).toBeHidden();
+        // Check that the extra paper (not added in `beforeEach`) is NOT visible
```
Merseleo:
```suggestion
        // Check that the extra paper (not added in before the test) is NOT visible
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,87 +4,140 @@
...
+    let isUpdatingBookmarkStatus = $state(true);
+
+    const loadingBookmarkStatus = loadingPaperId.then((id) => {
+        return checkInitialBookmarkStatus(id);
+    });
```
Merseleo:
```suggestion
    const loadingBookmarkStatus = loadingPaperId.then((id) => checkInitialBookmarkStatus(id));
```

src/lib/components/composites/PaperBookmarkButton.svelte
```diff
@@ -4,87 +4,140 @@
...
 
     const onMouseEnter = () => (isHovered = true);
     const onMouseLeave = () => (isHovered = false);
 
+    async function checkInitialBookmarkStatus(id: string) {
+        return await backendService
+            .isPaperOnReadingList({ id })
+            .response.then((response) => {
+                return response.value;
+            })
```
Merseleo:
```suggestion
            .response.then((response) => response.value)
```

Merseleo (changes requested):
One general question: is it intended in your commit message that you write "fix: fix ..." or "refactor: Refactor ..."? 

Only some minor left overs from debugging or spelling mistakes to be fixed, then I think it is ready now (finally). Good job with the refactoring!

tests/integration/download-button.test.ts
mowi12: oh you're right, oops

mowi12 (commented):
Yes, if I really only fix something in a commit, then I start with fix: fix ... because thats what I did (e.g.) with the button size and position.

And I really hope that we are now done : )

Merseleo (approved):

PR #352, opened 2025-05-02
Rename button in account settings is not correctly aligned

### LINKED ISSUE #346: [Bug]: Rename button in account settings is not correctly aligned

#58

### What happened?

When I open the account settings and want to change my name, the "Rename" button to apply my changes is not aligned with the input fields:

Image

### Steps to reproduce

1. Go to the user settings
2. Open the account settings
3. Look at the "Rename" button in the section "Change name"

### What did you expect?

I would expect that this button is vertically aligned, i.e. on the same baseline, as the input fields.

### Anything else?

_No response_

mowi12:
Closes #346 

## What I have made

- Added correct height to correct component

## Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ not needed
  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests

### DIFF
--- a/src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
+++ b/src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
@@ -78,7 +78,7 @@ Usage:
     >
         <Input
             bind:this={firstNameInput}
-            class="h-full w-full"
+            class="w-full"
             errorMessagePrefix="First Name must contain"
             inputId="first-name-input"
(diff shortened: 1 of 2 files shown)


domienderle (dismissed):

mowi12:
@domienderle Please check if all input/search bars and buttons have the same size. They should be!

domienderle (approved):
Yes, for me they are all the same size.

Issue #12, opened 2024-10-28
[Use Case]: Change Default Review Criteria


Slartibartfass2:
### Who is the primary actor in this use case?

User

### Describe the use case

As user, I want to change the review criteria that are used when I create a project.

### Acceptance criteria

- ability to add exclusion, soft exclusion and inclusion criteria to the default review criteria

### Anything else?

Part of #11

PR #353, opened 2025-05-03
Search for papers

### LINKED ISSUE #37: [Use Case]: Search For Papers

mowi12:
Closes #37 

## What I have made

- Added search functionality to papers overview page of a project
  - Updated `StageEntry` component.
  - If the first character is a '#' then only ids are searched

## Checklist

  - [ ] ~~unit tests~~
  - [ ] ~~integration tests~~ handled by searchbar tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/select/StageEntry.svelte
+++ b/src/lib/components/composites/select/StageEntry.svelte
@@ -5,15 +5,28 @@
     import type { Project_Paper } from "$lib/model/api/project";
     import type { Stage } from "$lib/model/general";
     import { pluralize } from "$lib/utils/common-helper";
+    import { filterProjectPapers } from "$lib/utils/filters";
     import CirclePlus from "lucide-svelte/icons/circle-plus";
 
     interface Props {
         projectId: string;
         stage: Stage;
         selectedPaper?: Project_Paper;
+        searchText?: string;
     }
 
-    let { projectId, stage, selectedPaper = $bindable(undefined) }: Props = $props();
+    let {
+        projectId,
+        stage,
+        selectedPaper = $bindable(undefined),
+        searchText = "",
+    }: Props = $props();
+
+    let filteredPapers = $derived(
+        searchText ? filterProjectPapers(stage.papers ?? [], searchText) : (stage.papers ?? []),
+    );
+
+    let totalPaperCount = $derived(stage.papers?.length ?? 0);
 </script>
 
 <Accordion.Item value={`stage-${stage.stageIndex}`}>
@@ -21,20 +34,36 @@
     <Accordion.Trigger data-testid="stage-entry-trigger">
         <div class="flex w-full flex-row justify-between">
             <span>Stage {stage.stageIndex}</span>
-            <span>({stage.papers.length} {pluralize(stage.papers.length, "paper", "papers")})</span>
+            {#if searchText}
+                <span>
+                    ({filteredPapers.length} / {totalPaperCount}
+                    {pluralize(totalPaperCount, "paper", "papers")})
+                </span>
+            {:else}
+                <span>({totalPaperCount} {pluralize(stage.papers.length, "paper", "papers")})</span>
+            {/if}
         </div>
     </Accordion.Trigger>
     <Accordion.Content>
         <div class="flex flex-col gap-4 pl-5">
-            {#each stage.papers as paper (paper.id)}
-                <PaperListEntry
-                    onClick={() => {
-                        selectedPaper = paper;
-                    }}
-                    {paper}
-                    {projectId}
-                />
-            {/each}
+            {#if filteredPapers.length > 0}
+                {#each filteredPapers as paper (paper.id)}
+                    <PaperListEntry
+                        onClick={() => {
+                            selectedPaper = paper;
+                        }}
+                        {paper}
+                        {projectId}
+                    />
+                {/each}
+            {:else if searchText && totalPaperCount > 0}
+                <span class="text-hint italic">
+                    No papers match your search criteria in this stage.
+                </span>
+            {:else if totalPaperCount === 0}
+                <span class="text-hint italic">No papers currently in this stage.</span>
+            {/if}
+
             <Button
                 onclick={() => {
                     // TODO: This is done in https://github.com/SE-UUlm/snowballr-frontend/issues/35
--- a/src/lib/components/composites/select/StageEntry.svelte
+++ b/src/lib/components/composites/select/StageEntry.svelte
@@ -29,6 +29,17 @@
     let totalPaperCount = $derived(stage.papers?.length ?? 0);
 </script>
 
+<!--
+@component
+Accordion entry for a stage in the project paper list. It displays the stage number and the number of papers in the stage.
+It also displays the number of papers that match the search criteria if a search text is provided.
+It contains a button to add a paper to the stage and a list of papers in the stage.
+
+Usage:
+```svelte
+    <StageEntry {projectId} {stage} {searchText} bind:selectedPaper />
+```
+-->
 <Accordion.Item value={`stage-${stage.stageIndex}`}>
     <!-- TODO: Add amount of filtered/all e.g. (3/7) -->
     <Accordion.Trigger data-testid="stage-entry-trigger">
--- a/src/routes/project/[projectId]/papers/+page.svelte
+++ b/src/routes/project/[projectId]/papers/+page.svelte
@@ -43,6 +43,7 @@
     const loadingStageCount = loadingProject.then((project) => project.maxStage);
 
     let showFilters = $state(true);
+    let searchText = $state("");
 
     interface PapersFilters {
         stages: string[];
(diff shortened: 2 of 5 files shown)


src/lib/components/composites/select/StageEntry.svelte
domienderle, line 1: There are unused imports in this file.

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -104,13 +105,11 @@
...
-                    onSearch={(searchText) => {
-                        // TODO: build filters from search text
-                        // This is done in https://github.com/SE-UUlm/snowballr-frontend/issues/37
-                        // and https://github.com/SE-UUlm/snowballr-frontend/issues/38
-                        console.log(searchText);
+                    onSearch={(text) => {
+                        searchText = text;
                     }}
                     placeholderText="Search paper"
+                    timeoutInMs={0}
```
Comment (domienderle) [inline src/routes/project/[projectId]/papers/+page.svelte:111]: Is this ```timeoutInMs={0}``` really neccessary?

src/lib/components/composites/select/StageEntry.svelte
```diff
@@ -5,36 +5,71 @@
...
+    let {
+        projectId,
+        stage,
+        selectedPaper = $bindable(undefined),
+        searchText = "",
+    }: Props = $props();
+
+    let filteredPapers = $state<Project_Paper[]>(stage.papers ?? []);
+
+    $effect(() => {
```
domienderle: At the moment the ```$effect``` rune is run three times each time the search paper is updated. Even though this does not affect the functionality it should be changed to run only a single time.

src/routes/project/[projectId]/papers/+page.svelte
Comment (domienderle) [inline src/routes/project/[projectId]/papers/+page.svelte:1]: There are unused imports in this file as well.

src/lib/components/composites/select/StageEntry.svelte
```diff
@@ -5,36 +5,71 @@
...
+        if (searchText) {
+            filteredPapers = filterProjectPapers(stage.papers ?? [], searchText);
+        } else {
+            filteredPapers = stage.papers ?? [];
+        }
+    });
+
+    let totalPaperCount = $derived(stage.papers?.length ?? 0);
 </script>
```
domienderle, line 31: Add a comment how to use the component
```suggestion
<!--
@component
Accordion-based stage entry displaying a list of papers, with optional search filtering
and the ability to select a paper or add a new one.

Usage:
```svelte
<StageEntry 
    {projectId}
    {searchText} 
    {stage} 
    bind:selectedPaper />
/>```
-->

File: tests/e2e/pom/project-papers-page-model.ts [hunk 4]
```diff
@@ -1,44 +1,97 @@
...
+    async openPaperView(stageIndex: number, paperIndexInStage: number) {
+        await this.openStage(stageIndex);
+        const paperLocator = this.getPaper(stageIndex, paperIndexInStage);
+        await paperLocator.dblclick();
+    }
+
+    /** Performs a search */
+    async search(text: string) {
+        await this.searchBarInput.fill(text);
+        await this.page.waitForTimeout(300);
```
domienderle: Do we need this timeout here? Did it cause any problems without it?

tests/e2e/pom/project-papers-page-model.ts
```diff
@@ -1,44 +1,97 @@
...
+    }
+
+    /** Clears the search using the Escape key */
+    async clearSearchViaEscape() {
+        await this.searchBarInput.focus();
+        await this.page.keyboard.press("Escape");
+        await this.page.waitForTimeout(300);
```
domienderle: Same here. Do we need this timeout?

tests/e2e/papers-overview.test.ts
```diff
@@ -46,86 +60,196 @@ test.describe("View all papers of a project", () => {
...
+        await expect(projectPapersPage.getPaperByTitle(fullTitleSearch)).toBeVisible();
+        await expect(projectPapersPage.getPaperByTitle(fullTitleHidden0)).toBeHidden();
+
+        // Stage 1 checks (no matches)
+        await projectPapersPage.expectStageCounts(1, `(0 / ${NUM_PAPERS_PER_STAGE} papers)`);
+        await expect(projectPapersPage.getPaperByTitle(fullTitleHidden1)).toBeHidden();
+        await expect(projectPapersPage.getNoSearchResultsText().last()).toBeVisible();
+    });
+
+    test("When the user searches for a specific author, then only papers with matching others are shown in open stages.", async ({
```
domienderle:
```suggestion
test("When the user searches for a specific author, then only papers with matching authors are shown in open stages.", async ({

tests/e2e/papers-overview.test.ts
domienderle, line 1: Add one test, that tests if the according paper is shown when the user enters a specific paper id.

domienderle (changes requested):
Looks good, but some changes are required. One acceptance criteria is still not implemented: 
- the matching part of the paper title is highlighted
Please implement this.

src/routes/project/[projectId]/papers/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/papers/+page.svelte:111]: Yes, in order to get immediate feedback while typing, this is necessary.
The `SearchBar` component in the reading list has the same value and therefore this search bar should have the same timeout, imo.

src/lib/components/composites/select/StageEntry.svelte
mowi12: First I wanted to use `$derived()` but I didn't get it to work. But now I tried again and found out what I did wrong. Thanks!

src/routes/project/[projectId]/papers/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/papers/+page.svelte:1]: I still have no in my view ... Could you also show me which are unused here?

src/lib/components/composites/select/StageEntry.svelte
mowi12: You're right. I did not create this component in the first place, but this doesn't mean i shouldn't add documentation to previous code.

tests/e2e/pom/project-papers-page-model.ts
mowi12: No not really. Since we are using the 0 milliseconds refresh for the search bar, this can be removed. I just thought it would make sense. To be honest, maybe code pilot added these lines and for it made sense at that moment.

tests/e2e/pom/project-papers-page-model.ts
mowi12: Same here.

tests/e2e/papers-overview.test.ts
mowi12: Good catch, makes more sense like this!

tests/e2e/papers-overview.test.ts
mowi12: I will do!

src/lib/components/composites/select/StageEntry.svelte
mowi12: I do not see any unused imports. Could you specify which?

mowi12 (commented):
I implemented all you requests.

I actually did not see the point about the highlighting.
I researched a little bit and I think that this feature should not be included in this issue.
It would make more sense as a separate issue and then try to add the highlighting to all searchable areas, e.g. the reading list.

What do you think?

I you agree, then I would update the description of the issue and create a new one for the future.

mowi12:
Just for future reference so we do not forget: https://github.com/metonym/svelte-fuzzy

mowi12:
After the meeting today, we decided to add a new functionality to the search.
If the first character is a '#' then only ids are searched.
They should match from the start, so '#2' matches all paper ids that start with a 2 but not '#12', even if it includes '#' and '2' in the correct order.

src/lib/components/composites/select/StageEntry.svelte
domienderle: Never mind. It is just my WebStorm that thinks this that one import is unnecessary, which is not the case. You are right.

src/routes/project/[projectId]/papers/+page.svelte
Comment (domienderle) [reply src/routes/project/[projectId]/papers/+page.svelte:1]: Forget it. It's a weird bug of WebStorm but the imports are just fine.

src/routes/project/[projectId]/papers/+page.svelte
Comment (domienderle) [reply src/routes/project/[projectId]/papers/+page.svelte:111]: You are right. This is plausible.

domienderle (changes requested):
Loks good so far, but there are a lot of e2e tests, that still fail. Fix them, then we should be good to merge.

domienderle (approved):

PR #361, opened 2025-05-09
fix: reload page data after navigation

### LINKED ISSUE #217: [Bug]: No reload of page data after navigation

#42

### What happened?

Wenn the user clicks (or double clicks) on a paper in the references / citations list and navigates to this paper page, the navigation is successful (can be seen in the updated URL), but the page is not loaded again and the loaded page data don't update, so the wrong title, authors etc. are shown.

### Steps to reproduce

1. Go to a paper page (e.g. /paper/0)
2. Click on one reference from the "References" list
3. Check the URL updated

### What did you expect?

The new paper page should display the information from the new paper and not from the old paper.

### Anything else?

Image

domienderle:
Closes #217 

## What I have made

I have added derived functions in the ```PaperView.svelte``` and ```paper/[paperId]/+page.svelte``` to translate the updates of the paper if the url is changed.

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ not necessary, because I only inserted the ```$derived``` rune 
  - ~~[ ] unit tests~~ already existed
  - ~~[ ] integration tests~~ already existed
  - ~~[ ] end-to-end tests~~ already existed

### DIFF
--- a/tests/e2e/paper-navigation.test.ts
+++ b/tests/e2e/paper-navigation.test.ts
@@ -1,0 +1,31 @@
+import { test } from "./fixtures/paper-view-fixture";
+import { createPaper } from "$tests/model-builder";
+import { expect } from "@playwright/test";
+import type { Paper } from "$lib/model/api/paper";
+
+test.describe("Navigate to referenced paper in paper view", () => {
+    const paperIds: string[] = [];
+    let papers: Paper[] = [];
+
+    test.beforeAll(async ({ mockBackendService }) => {
+        papers = await Promise.all(
+            Array.from(
+                { length: 3 },
+                (_, i) =>
+                    mockBackendService.createPaper(
+                        createPaper({ title: `Paper ${i} to be referenced` }),
+                    ).response,
+            ),
+        );
+        papers.forEach((paper) => paperIds.push(paper.id));
+    });
+
+    test("When clicking on a referenced paper in the paper view (no project paper) the paper view of the referenced paper is shown", async ({
+        page,
+        paperViewPage,
+    }) => {
+        await paperViewPage.openPaperView(paperIds[1]);
+        await paperViewPage.navigateToReferencedPaper();
+        await expect(page.getByText("Paper 0 to be referenced").first()).toBeVisible();
+    });
+});
--- a/tests/e2e/pom/paper-view-model.ts
+++ b/tests/e2e/pom/paper-view-model.ts
@@ -1,0 +1,34 @@
+import { expect, type Locator, type Page } from "@playwright/test";
+
+export class DevPaperViewPage {
+    readonly page: Page;
+    readonly allListEntries: Locator;
+    readonly referenceListEntry0: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.allListEntries = page.getByRole("listitem");
+        this.referenceListEntry0 = this.allListEntries
+            .getByRole("button", {
+                name: "Paper 0 to be referenced",
+            })
+            .first();
+    }
+
+    /**
+     * Navigates to a certain paper view and ensures the page is loaded.
+     *
+     * @param paperId - The local project paper id
+     */
+    async openPaperView(paperId: string) {
+        await this.page.goto(`/paper/${paperId}`);
+        await expect(this.referenceListEntry0).toBeVisible();
+    }
+
+    /**
+     * Navigates to the first referenced paper and ensures the page is loaded.
+     */
+    async navigateToReferencedPaper() {
+        await this.referenceListEntry0.click();
+    }
+}
--- a/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperView.svelte
@@ -44,6 +44,7 @@
     export type PaperViewProps = IndependentPaperViewProps &
         (ProjectPaperViewProps | NonProjectPaperViewProps);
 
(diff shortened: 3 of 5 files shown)


domienderle (commented):
Can be reviewed. The failing pipeline originates from the flaky tests.

mowi12 (changes requested):
Your changes look very good. I mean, there is not much to look at, which is good!
But I have two requests.

The first thing that came to mind is, that the page always quickly shows the skeletons while reloading. This is kinda annoying.
The problem is, that we actually redirect to a new page (i think), since we are actually changing the URL, although we are still in the same svelte file.

Try having a look at the code in branch feat/100. If I remove a paper from the reading list, there is no reload, and no skeletons are shown. This is because I only update the UI after the new papers are loaded.

Felix did the same here (he sent me those some time ago):

Da wird das neu geladen:
https://github.com/SE-UUlm/snowballr-frontend/blob/9130e0e12aa74e4e2279f2c4b4242fcd784bc51c/src/routes/project/%5BprojectId%5D/settings/members/%2Bpage.svelte#L30
Da ruf ich den Callback auf:
https://github.com/SE-UUlm/snowballr-frontend/blob/9130e0e12aa74e4e2279f2c4b4242fcd784bc51c/src/lib/components/composites/settings/project-settings/members/InviteUsersDialog.svelte#L57


Second, you checked the checkmark, that the e2e test already exists. But I think, there is no e2e test for clicking on the references of a paper.
Add or create a new use case for clicking on things in the paper view, which in your case, only needs to test the switching of the paper view.

domienderle (commented):
The first change is not possible. The entries in the referenced papers are of ```PaperListEntry```. This entries have to implement a ```goto()``` method, because in other lists where ```PaperListEntry``` elemts are used it is required to change the page to the requested ```URL```. But with #362 the time of the loading should be as minimal as possible, because then the ressources are preloaded (as long as single and double click on the referenced paper entries are the same).

domienderle (commented):
You are right. I didn't really look into the e2e tests. Now I created a test, that checks if the navigation works properly, by clicking on a referenced paper (which should cover all of this fix issue).

tests/e2e/pom/paper-view-model.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class DevPaperViewPage {
+    readonly page: Page;
+    readonly allListEntries: Locator;
+    readonly referenceListEntry0: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.allListEntries = page.getByRole("listitem");
+        this.referenceListEntry0 = page
```
mowi12: I think you wanted to use the `allListEntries` var, correct?

```suggestion
        this.referenceListEntry0 = allListEntries
```

tests/e2e/pom/paper-view-model.ts
```diff
@@ -0,0 +1,34 @@
...
+            .getByRole("button", {
+                name: /Paper 0 to be referenced/,
+            })
+            .nth(0);
```
mowi12: Imo, `first()` looks a bit cleaner if you only want the first item. What do you think?

tests/e2e/pom/paper-view-model.ts
```diff
@@ -0,0 +1,34 @@
...
+    readonly page: Page;
+    readonly allListEntries: Locator;
+    readonly referenceListEntry0: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.allListEntries = page.getByRole("listitem");
+        this.referenceListEntry0 = page
+            .getByRole("button", {
+                name: /Paper 0 to be referenced/,
```
mowi12: You can actually just use a string instead of the regex, since you do not use any regex parts.

```suggestion
                name: "Paper 0 to be referenced",
```

If you had used `name: "Paper 0 to be referenced", exact: true`, then this would make a difference, but without it, it does not need to be exact.

tests/e2e/paper-navigation.test.ts
```diff
@@ -0,0 +1,31 @@
...
+        papers.forEach((paper) => paperIds.push(paper.id));
+    });
+
+    test("When clicking on a referenced paper in the paper view (no project paper) the paper view of the referenced paper is shown", async ({
+        page,
+        paperViewPage,
+    }) => {
+        await paperViewPage.openPaperView(paperIds[1]);
+        await paperViewPage.navigateToReferencedPaper();
+        await expect(page.getByText("Paper 0 to be referenced").nth(0)).toBeVisible();
```
mowi12: Same here, I think `first()` is a bit cleaner if only the first element is relevant.

```suggestion
        await expect(page.getByText("Paper 0 to be referenced").first()).toBeVisible();
```

mowi12 (changes requested):
Nice, now everything looks good. I only have some minor suggestions for you to take a look at. But then we are done!

tests/e2e/pom/paper-view-model.ts
domienderle: I guess, that is true.

tests/e2e/pom/paper-view-model.ts
domienderle: Yes. That is just a remnant of previous tries.

mowi12 (approved):
Now everything looks good!

One small thing. In the future let the reviewer resolve the conversations, he is the one who has to decide if your changes are good, or not. In this case I already gave you suggestions, but you know : )

PR #362, opened 2025-05-09
feat: optionally wrap list entries in link tags

### LINKED ISSUE #324: [Feature]: Optionally wrap list entries in link tags

_No response_

The components `ProjectListEntry`, `PaperListEntry` and `ReadingListEntry` are all customized buttons to allow defining a single and double-click behavior. However, sometimes these components are used only with the single-click behavior, so the option to define the double-click behavior is unnecessary. Moreover, it is then unnecessary to "wrap" these components in a `button`. Instead they should be wrapped by an `a` tag as this would allow preloading in Svelte.

Extend the named components, so if the `onClick` event for the double-click is not defined, then an `a` tag is used instead of an `button` tag.

Alternatively, we could introduce a new boolean prop to select between a button and a link.

### Anything else?

Maybe [svelte:element](https://svelte.dev/docs/svelte/svelte-element) could help solving this problem.

domienderle:
Closes #324 

## What I have made

I have replaced the button element in the ReadingListEntry/PaperListEntry and the link element in the ProjectListEntry to a svelte:element, which is either a link or a button depending upon the onClick handler for single click is set or not.

## Checklist

  - ~~[ ] unit tests~~ not necessary
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ no new use case, but fixed failing tests

### DIFF
--- a/src/lib/components/composites/paper-components/PaperListEntry.svelte
+++ b/src/lib/components/composites/paper-components/PaperListEntry.svelte
@@ -17,20 +17,11 @@
     type PaperListEntryProps = PaperListEntryInterface & {
         onClick?: () => void;
     };
-
-    const navigateToPaperView = () => {
-        if (isProjectPaper(paper)) {
-            // navigate to project paper
-            goto(`/project/${projectId}/paper/${paperId}`);
-        } else {
-            // navigate to paper
-            goto(`/paper/${paperId}`);
-        }
-    };
-
-    const { paper, projectId, onClick = navigateToPaperView }: PaperListEntryProps = $props();
-
+    const { paper, projectId, onClick }: PaperListEntryProps = $props();
     const paperId = $derived(getDisplayPaperId(paper));
+    const href = $derived(
+        isProjectPaper(paper) ? `/project/${projectId}/paper/${paperId}` : `/paper/${paperId}`,
+    );
 
     async function getReviewUserById(id: string): Promise<User | undefined> {
         let reviewingUser: undefined | User = undefined;
@@ -60,15 +51,16 @@ on a single click. A double click always causes the navigation to the paper view
 
 Usage:
 ```svelte
-    <PaperListEntry paper={paper} projectId={"1"} />
+    <PaperListEntry {paper} projectId={"1"} {onClick} />
 ```
 -->
-<button
+<svelte:element
+    this={!onClick ? "a" : "button"}
     class="border-container-border-grey highlight-on-hover group/paper-list-entry flex w-full flex-row items-center justify-end gap-3 rounded-md border pe-3"
     class:border-l-0={!reviewMode.isActivated}
     data-testid="paper-list-entry"
-    onclick={handleSingleOrDoubleClick(onClick, navigateToPaperView)}
-    type="button"
+    onclick={handleSingleOrDoubleClick(onClick ?? (() => {}), () => goto(href))}
+    {...!onClick ? { href: href } : { type: "button" }}
 >
     <div
         class={cn(
@@ -90,4 +82,4 @@ Usage:
             {/await}
         {/each}
     {/if}
-</button>
+</svelte:element>
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -12,16 +12,14 @@
         onPaperChangedBookmarkStatus?: () => void;
     }
 
-    const navigateToPaperView = () => {
-        goto(`/paper/${id}`);
-    };
-
     const {
         paper,
-        onClick = navigateToPaperView,
+        onClick,
         onPaperChangedBookmarkStatus = undefined,
     }: ReadingListEntryProps = $props();
     const { id, ...paperWithoutId } = paper;
+
+    const href = `/paper/${id}`;
 </script>
 
 <!--
@@ -43,20 +41,21 @@ This callback is executed when the bookmark status changes.
 
 Usage:
 ```svelte
-    <ReadingListEntry paper={paper} {onPaperChangedBookmarkStatus} />
+    <ReadingListEntry {paper} {onClick} {onPaperChangedBookmarkStatus} />
 ```
 -->
 <div
     class="border-container-border-grey highlight-on-hover flex w-full flex-row items-center gap-16 rounded-md border px-3 py-2"
 >
-    <button
+    <svelte:element
+        this={!onClick ? "a" : "button"}
         class="flex flex-auto"
         aria-label="Paper info for reading list entry"
-        onclick={handleSingleOrDoubleClick(onClick, navigateToPaperView)}
-        type="button"
+        onclick={handleSingleOrDoubleClick(onClick ?? (() => {}), () => goto(href))}
+        {...!onClick ? { href: href } : { type: "button" }}
     >
         <PaperInfo class="gap-1" loadingPaper={Promise.resolve(paperWithoutId)} />
-    </button>
+    </svelte:element>
     <div class="flex flex-row items-center gap-4">
         <PaperBookmarkButton
             isBookmarkedDefault={true}
--- a/src/lib/components/composites/project-components/ProjectListEntry.svelte
+++ b/src/lib/components/composites/project-components/ProjectListEntry.svelte
@@ -1,10 +1,17 @@
 <script lang="ts">
     import { Progress } from "$lib/components/primitives/progress";
-    import { getNames } from "$lib/utils/common-helper";
+    import { getNames, handleSingleOrDoubleClick } from "$lib/utils/common-helper";
     import { ProjectStatus } from "$lib/model/api/project";
     import type { ProjectListEntryInterface } from "$lib/model/component-interfaces";
+    import { goto } from "$app/navigation";
 
-    const { project, membersList, information }: ProjectListEntryInterface = $props();
+    type ProjectListEntryProps = ProjectListEntryInterface & {
+        onClick?: () => void;
+    };
+
+    const { project, membersList, information, onClick }: ProjectListEntryProps = $props();
+
+    const href = `/project/${project.id}/dashboard`;
 </script>
 
 <!--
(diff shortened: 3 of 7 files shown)


src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -63,12 +65,14 @@ Usage:
     <PaperListEntry paper={paper} projectId={"1"} />
```
mowi12: I know, you haven't added `onClick`, but could you please update the doc?
You can have a look at the `PaperBookmarkButton`, there is also an example for optional functions.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -42,6 +41,9 @@
 
         return reviewingUser;
     }
+    const href = $derived(
+        isProjectPaper(paper) ? `/project/${projectId}/paper/${paperId}` : `/paper/${paperId}`,
+    );
```
mowi12: Maybe move these lines up below `const paperId ...`, so all reactive states are at the top?

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -63,12 +65,14 @@ Usage:
     <PaperListEntry paper={paper} projectId={"1"} />
 ```
 -->
-<button
+
```
mowi12: Please remove this empty line.

src/lib/components/composites/paper-components/PaperListEntry.svelte
```diff
@@ -28,8 +28,7 @@
         }
     };
```
mowi12: The `navigateToPaperView` function executes the same check as the `const href` part.
Can't you simply reuse the `href` variable?

```suggestion
    const navigateToPaperView = () => goto(href);
```

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -39,14 +39,15 @@ Usage:
...
 >
-    <button
+    <svelte:element
+        this={!onClick ? "a" : "button"}
         class="flex flex-auto"
         aria-label="Paper info for reading list entry"
-        onclick={handleSingleOrDoubleClick(onClick, navigateToPaperView)}
-        type="button"
+        onclick={handleSingleOrDoubleClick(onClick ?? (() => {}), navigateToPaperView)}
+        {...!onClick ? { href: `/paper/${id}` } : { type: "button" }}
```
mowi12: Same here, maybe define `href` as `const` at the top and reuse it here and in the `navigateToPaperView`.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -39,14 +39,15 @@ Usage:
 <div
```
mowi12, line 47: Please update the doc to include the `onclick`. Same as in `PaperListEntry.svelte`.

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -24,11 +32,13 @@ Usage:
     <ProjectListEntry project={demoProject} membersList={memberUserSpecArray} information={{projectProgress: 0.3}} />
 ```
 -->
```
mowi12, line 35: Please update the doc to include the `onClick`.

src/lib/components/composites/project-components/ProjectListEntry.svelte
```diff
@@ -24,11 +32,13 @@ Usage:
...
-<a
+<svelte:element
+    this={!onClick ? "a" : "button"}
     class="border-container-border-grey highlight-on-hover flex h-fit w-full flex-col justify-between gap-2 rounded-md border px-5
     py-2 lg:flex-row lg:items-center lg:gap-10"
     class:opacity-25={project.status === ProjectStatus.ARCHIVED}
-    href={`/project/${project.id}/dashboard`}
+    onclick={handleSingleOrDoubleClick(onClick ?? (() => {}), navigateToProjectDashboard)}
+    {...!onClick ? { href: `/project/${project.id}/dashboard` } : { type: "button" }}
```
mowi12: same here. Move `href` to the top, and reuse it here and in the `navigateToProjectDashboard`.

tests/e2e/pom/paper-view-model.ts
mowi12, line 1: I will look at this test, after you rebase since there will be some small changes.

tests/integration/paper-components/paper-list-entry.test.ts
```diff
@@ -27,6 +77,7 @@ describe("PaperListEntryComponent", () => {
         expect(screen.getByText("John Doe, Bob Johnson")).toBeInTheDocument();
 
         // border does not indicate review status
+        // It is a link, because no onClick handler was provided
```
mowi12: This comment makes no sense, since you use a button in the line below.

tests/integration/paper-components/paper-list-entry.test.ts
```diff
@@ -54,6 +106,7 @@ describe("PaperListEntryComponent", () => {
...
```
mowi12: Same here.

mowi12 (changes requested):
Your implementation looks good. The hint with the `svelte:element` from @Merseleo looks really nice.

I only have some small code duplication critics and missing docs you need to add.

Also, please rebase so I can review your e2e tests as well, as they will change a little bit (i think).

src/lib/components/composites/paper-components/PaperListEntry.svelte
domienderle: Good suggestion. I will do that.

src/lib/components/composites/paper-components/PaperListEntry.svelte
domienderle: Yes you are right. In the element usage it is missing. The rest should be documented. Is it ok, as it is now?

src/lib/components/composites/paper-components/PaperListEntry.svelte
domienderle: Good point. Thank you.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
domienderle: The refactoring of the href is a good idea. Done.

src/lib/components/composites/project-components/ProjectListEntry.svelte
domienderle: Good idea. Is done.

tests/e2e/pom/paper-view-model.ts
domienderle: I have rebased. You can take a look now.

tests/integration/paper-components/paper-list-entry.test.ts
domienderle: Yes you are right. Copy paste mistake. Sorry

domienderle (commented):
It is already rebased. So you can take a look now.

src/lib/components/composites/paper-components/PaperListEntry.svelte
mowi12: Yes that's better!

But I missed something the last time. `paper={paper}` should be `{paper}` since the formatter would also format it that way.

src/lib/components/composites/paper-components/PaperListEntry.svelte
mowi12: In my code, this empty line is still there.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -43,20 +42,21 @@ This callback is executed when the bookmark status changes.
 
 Usage:
 ```svelte
-    <ReadingListEntry paper={paper} {onPaperChangedBookmarkStatus} />
+    <ReadingListEntry paper={paper} {onClick} {onPaperChangedBookmarkStatus} />
```
mowi12: Same here. `{paper}` is enough.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
```diff
@@ -43,20 +42,21 @@ This callback is executed when the bookmark status changes.
...
         class="flex flex-auto"
         aria-label="Paper info for reading list entry"
-        onclick={handleSingleOrDoubleClick(onClick, navigateToPaperView)}
-        type="button"
+        onclick={handleSingleOrDoubleClick(onClick ?? (() => {}), navigateToPaperView)}
+        {...!onClick ? { href: href } : { type: "button" }}
     >
         <PaperInfo class="gap-1" loadingPaper={Promise.resolve(paperWithoutId)} />
-    </button>
+    </svelte:element>>
```
mowi12: Here is `>` duplicated.

tests/e2e/pom/paper-view-model.ts
```diff
@@ -31,4 +27,26 @@ export class DevPaperViewPage {
...
         await this.referenceListEntry0.click();
     }
+
+    /**
+     * Depending upon the list entries are links or buttons,
+     * this function returns the first list entry of the paper view.
+     *
+     * @returns The first list entry of the paper view.
+     */
+    getFirstListEntry(): Locator {
```
mowi12: Since you know/expect the type to be either "link" or "button" you can simply use `role: "link" | "button"` as function parameter. This is compatible with the role parameter expected by the `getByRole()` function.

mowi12 (changes requested):
First, I think the you commits are not correct. The first commit is not only adjusting the e2e tests but also adding the new functionality. 

I think the next time, this PR should be done. Only really minor fixes needed.

src/lib/components/composites/paper-components/PaperListEntry.svelte
domienderle: I guess you cannot delete this. I'm not sure which, but one of the automatic code stylers adds this line add the end of every file. Have a look on any other file, they all have this, so I think it shouldn't be a problem.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
domienderle: True.

src/lib/components/composites/paper-components/ReadingListEntry.svelte
domienderle: Good catch.

tests/e2e/pom/paper-view-model.ts
domienderle: Thank you. That was the solution I was looking for. Could have been so easy ...

domienderle (commented):
The commits should be fixed now.

src/lib/components/composites/paper-components/PaperListEntry.svelte
mowi12: I looked at several components, none have this line.
You could simply use nano or vim and remove this line from there.

tests/e2e/pom/paper-view-model.ts
mowi12: Lines 11-13 are a bit chunky since you have to execute the link part two times in the worst case, but since AI couldn't give me any better solution, this will stay.

mowi12 (changes requested):
Perfect. This looks better now.
Only the empty line left, then we are good to go!

mowi12 (approved):
Perfect. I think we are done now!

I will approve your changes, but before you merge, could you please rename the first commit back to 'feat: ...' not ' chore: ... applied review' this is not completely correct.

PR #372, opened 2025-05-13
Implement Sign Out

### LINKED ISSUE #5: [Use Case]: Sign out

User

### Describe the use case

As the user, I want to sign out when I'm currently signed in.

### Acceptance criteria

- the user should be able to sign out from everywhere in the application as long as they're signed in
- after the user is signed out, the sign-in screen is shown

### Anything else?

Related to #1

### LINKED ISSUE #371: [Feature]: Sign Out

#5

Users should be able to log out.

The `Sign Out` item in the profile dropdown should sign me out.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #5, #371
Requires: #220 for E2E Tests

## What I have made

- Added a new `/signout` route.
- This route calls the backend's `logout` method and redirects to `/signin`
- The `Sign Out` item in the user menu dropdown now redirects to `/signout`.

## Checklist

- [ ] ~~I have updated the documentation accordingly and commented my code~~ No need
  - [ ] unit tests
  - [ ] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/navigation-bar/UserMenu.svelte
+++ b/src/lib/components/composites/navigation-bar/UserMenu.svelte
@@ -70,11 +70,13 @@
                 {/each}
             </DropdownMenu.Group>
             <DropdownMenu.Separator />
-            <DropdownMenu.Item>
-                <LogOut class="mr-2 size-4" />
-                <span>Sign out</span>
-                <DropdownMenu.Shortcut>⇧⌘Q</DropdownMenu.Shortcut>
-            </DropdownMenu.Item>
+            <a href="/signout">
+                <DropdownMenu.Item>
+                    <LogOut class="mr-2 size-4" />
+                    <span>Sign Out</span>
+                    <DropdownMenu.Shortcut>⌘⇧Q</DropdownMenu.Shortcut>
+                </DropdownMenu.Item>
+            </a>
         </DropdownMenu.Group>
     </DropdownMenu.Content>
 </DropdownMenu.Root>
--- a/src/routes/(auth)/signout/+page.svelte
+++ b/src/routes/(auth)/signout/+page.svelte
@@ -1,0 +1,16 @@
+<script lang="ts">
+    import { backendService } from "$lib/grpc-api";
+    import { goto } from "$app/navigation";
+    import { onMount } from "svelte";
+    import { Nothing } from "$lib/model/api/base";
+
+    onMount(async () => {
+        try {
+            await backendService.logout(Nothing);
+        } catch (e) {
+            console.error("There was an error logging out: ", e);
+        } finally {
+            await goto("/signin");
+        }
+    });
+</script>
--- a/tests/e2e/pom/home-page-model.ts
+++ b/tests/e2e/pom/home-page-model.ts
@@ -1,6 +1,6 @@
 import { expect, type Locator, type Page } from "@playwright/test";
 
-type LinkName = "Reading List" | "Archived Projects" | "Invitations" | "Settings";
+type LinkName = "Reading List" | "Archived Projects" | "Invitations" | "Settings" | "Sign Out";
 type SettingName = "Account" | "Project Setup" | "Shortcuts" | "Review";
 
 export class DevHomePage {
@@ -27,10 +27,12 @@ export class DevHomePage {
      *
      * @param linkName - The name of the link to be clicked in the user menu dialog.
      */
-    async openLinkInUserMenuDialog(linkName: LinkName) {
+    async openLinkInUserMenuDialog(linkName: LinkName, checkDestination: boolean = true) {
         await this.page.getByRole("button", { name: /^[A-Z]{2}$/ }).click();
         await this.page.getByRole("link", { name: linkName }).click();
-        await expect(this.page.getByRole("heading", { name: linkName })).toBeVisible();
+        if (checkDestination) {
+            await expect(this.page.getByRole("heading", { name: linkName })).toBeVisible();
+        }
     }
 
     /**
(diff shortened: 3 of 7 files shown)


src/lib/components/composites/navigation-bar/UserMenu.svelte
```diff
@@ -70,11 +70,13 @@
...
-            <DropdownMenu.Item>
-                <LogOut class="mr-2 size-4" />
-                <span>Sign out</span>
-                <DropdownMenu.Shortcut>⇧⌘Q</DropdownMenu.Shortcut>
-            </DropdownMenu.Item>
+            <a href="/signout">
+                <DropdownMenu.Item>
+                    <LogOut class="mr-2 size-4" />
+                    <span>Sign out</span>
+                    <DropdownMenu.Shortcut>⇧⌘Q</DropdownMenu.Shortcut>
```
mowi12: As all other shortcuts display the command symbol before the shift symbol, please switch the symbols so they match.

src/lib/components/composites/navigation-bar/UserMenu.svelte
```diff
@@ -70,11 +70,13 @@
...
             <DropdownMenu.Separator />
-            <DropdownMenu.Item>
-                <LogOut class="mr-2 size-4" />
-                <span>Sign out</span>
-                <DropdownMenu.Shortcut>⇧⌘Q</DropdownMenu.Shortcut>
-            </DropdownMenu.Item>
+            <a href="/signout">
+                <DropdownMenu.Item>
+                    <LogOut class="mr-2 size-4" />
+                    <span>Sign out</span>
```
mowi12:
```suggestion
                    <span>Sign Out</span>
```

Reading list and archived projects are also written with capital letters.

src/routes/(auth)/signout/+page.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+    import { backendService } from "$lib/grpc-api";
+    import { goto } from "$app/navigation";
+    import { onMount } from "svelte";
+
+    onMount(async () => {
+        try {
+            await backendService.logout({});
```
mowi12: Please use the `Nothing` object to keep everything consistent.

src/routes/(auth)/signout/+page.svelte
```diff
@@ -0,0 +1,14 @@
...
+        } catch (e) {
+            console.error("There was an error logging out: ", e);
+        }
+        goto("/signin");
```
mowi12:
```suggestion
        } finally {
            goto("/signin", { replaceState: true });
        }
```

Maybe add this option to goto. If I understand this correctly, with this option, this page will not be added to the browser history or at least it will not be visited if you click on the back and forward buttons in the top left. I hope you know what I mean.

As well as moving this into a finally block, since it is literally for this use case.

tests/e2e/pom/home-page-model.ts
```diff
@@ -27,10 +27,12 @@ export class DevHomePage {
      *
      * @param linkName - The name of the link to be clicked in the user menu dialog.
      */
-    async openLinkInUserMenuDialog(linkName: LinkName) {
+    async openLinkInUserMenuDialog(linkName: LinkName, checkDestinaiton: boolean = true) {
```
mowi12:
```suggestion
    async openLinkInUserMenuDialog(linkName: LinkName, checkDestination: boolean = true) {
```

Just a small typo.

tests/e2e/signout.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import { DevHomePage } from "./pom/home-page-model";
+
+test.describe("Sign Out Functionality", () => {
+    test("When the user clicks logout, then they are logged out and will thus be redirected to /signin", async ({
+        page,
+    }) => {
+        await page.goto("/");
+        await expect(page.getByRole("heading", { name: "SnowballR" })).toBeVisible();
+        await expect(page.getByText("Create Project", { exact: true })).toBeVisible();
+        const homePage = new DevHomePage(page);
```
mowi12: Can't we move this part into the fixture just like in any other fixture (e.g. reading list, project settings, ...)

mowi12 (changes requested):
Nice work. I found no errors or problems concerning the logic.
There are only some small details, some typos and stuff you need to fix. Then we are good to go.

I also looked into the broken link error, but I found no solution or anyone else with the same problem. In addition, the page seems to be working, so ... no idea what to do. Maybe it's working again after the next commit.

src/routes/(auth)/signout/+page.svelte
luca-schlecker: Great suggestion :+1:

mowi12 (approved):
Nicely done. I think this was the first time, that everything is resolved after the first review and there is nothing else to change.

mowi12 (changes requested):
Nvm, I found this issue:

Screenshot 2025-05-20 at 15 40 53

This is the output of the console, after opening the /signout path manually after I am already signed out.

I don't know, but I don't think this can be ignored, or can it?

mowi12 (changes requested):
Everything looks good so far. You only need to answer the question i asked you in the last review.

luca-schlecker:
> Nvm, I found this issue:
> 
> Screenshot 2025-05-20 at 15 40 53
> This is the output of the console, after opening the /signout path manually after I am already signed out.
> 
> I don't know, but I don't think this can be ignored, or can it?

I fixed it. It was due to the `throw`s in `+layout.ts`.

mowi12 (approved):
This definitely better than before, nice work!

PR #373, opened 2025-05-13
Implement Sign In Page

### LINKED ISSUE #1: [Use Case]: Sign In

luca-schlecker:
Closes #1 
Supersedes #124 
Requires #220 for e2e tests

## What I have made

- Allow disabling validation in `PasswordInput`
- Allow specifying `link` for `PasswordInput` just like the underlying `Input` component does.
- Added signin page aligned with the style of the signup page
- Implement signin logic (sending backend request, redirecting)

image

## Remarks

PR is in my opinion best evaluated with #220 as a base, because then the signin process can be tried out.

## Checklist

  - [ ] unit tests
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/routes/(auth)/signin/+page.svelte
+++ b/src/routes/(auth)/signin/+page.svelte
@@ -1,4 +1,116 @@
+<script lang="ts">
+    import Input from "$lib/components/composites/input/Input.svelte";
+    import PasswordInput from "$lib/components/composites/input/PasswordInput.svelte";
+    import { Button } from "$lib/components/primitives/button/index.js";
+    import * as Card from "$lib/components/primitives/card/index.js";
+    import { backendService } from "$lib/grpc-api";
+    import { Schema } from "$lib/schemas";
+    import type { ApiError } from "$lib/model/general";
+    import ErrorAlert from "$lib/components/composites/utils/ErrorAlert.svelte";
+    import { StatusCodes } from "$lib/model/error-codes";
+    import { goto } from "$app/navigation";
+    import { onMount } from "svelte";
+    import { Nothing } from "$lib/model/api/base.js";
+    import { AuthenticationStatus } from "$lib/model/api/authentication.js";
+    import { cn } from "$lib/utils/shadcn-helper";
+
+    let emailInput: Input;
+    let passwordInput: PasswordInput;
+
+    let signinError: ApiError | undefined = $state(undefined);
+
+    let isLoading = $state(true);
+
+    onMount(async () => {
+        try {
+            const authStatus = (await backendService.getAuthenticationStatus(Nothing).response)
+                .authenticationStatus;
+            if (authStatus === AuthenticationStatus.AUTHENTICATED) {
+                await goto("/");
+            }
+        } catch {
+            console.error("There was an error acquiring the authentication status.");
+        }
+
+        isLoading = false;
+    });
+
+    async function handleSubmit(event: Event) {
+        event.preventDefault();
+        signinError = undefined;
+
+        const isEmailValid = emailInput.validate();
+        if (!isEmailValid) {
+            return;
+        }
+
+        const userData = {
+            email: emailInput.getValue(),
+            password: passwordInput.getValue(),
+        };
+
+        backendService
+            .login(userData)
+            .then(async () => await goto("/"))
+            .catch((error) => {
+                if (error.code === StatusCodes.UNAUTHENTICATED) {
+                    signinError = {
+                        errorTitle: "The provided credentials are not correct.",
+                        errorDetails:
+                            "Check if the email and password are correct or try resetting your password.",
+                    };
+                } else {
+                    signinError = { errorTitle: "Something unknown went wrong." };
+                }
+                console.error(error);
+            });
+    }
+</script>
+
 <svelte:head>
     <title>Sign In</title>
 </svelte:head>
-<h2>Sign In</h2>
+
+<Card.Root
+    class={cn(
+        "flex w-full max-w-xl flex-col border-slate-500 shadow-lg",
+        isLoading ? "opacity-0" : "",
+    )}
+>
+    <Card.Header class="flex w-full flex-col">
+        <Card.Title class="text-3xl">Sign In</Card.Title>
+        <Card.Description>Enter your information to sign in to your account</Card.Description>
+    </Card.Header>
+    <Card.Content class="flex w-full flex-col">
+        <form class="flex flex-col gap-5" onsubmit={handleSubmit}>
+            <Input
+                bind:this={emailInput}
+                class="w-full"
+                errorMessagePrefix="Email must have"
+                inputId="email-input"
+                label="Email"
+                placeholder="john.doe@example.com"
+                required
+                schema={Schema.email}
+                type="email"
+            />
+            <PasswordInput
+                bind:this={passwordInput}
+                class="w-full"
+                link={{ href: "/resetpassword", text: "Forgot Password?" }}
+                validate={false}
+            />
+            <Button class="w-full" type="submit">Sign In</Button>
+            {#if signinError}
+                <ErrorAlert
+                    errorDetails={signinError.errorDetails}
+                    errorTitle={signinError.errorTitle}
+                />
+            {/if}
+        </form>
+        <div class="mt-4 text-center text-sm">
+            You don't have an account yet?
+            <a class="underline" href="/signup"> Sign Up </a>
+        </div>
+    </Card.Content>
+</Card.Root>
--- a/tests/e2e/signin.test.ts
+++ b/tests/e2e/signin.test.ts
@@ -1,0 +1,54 @@
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/isolated-fixture";
+import { alice } from "./users";
+
+test.use({ user: null });
+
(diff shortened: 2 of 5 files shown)


tests/integration/input/password-input.test.ts
```diff
@@ -30,6 +30,42 @@ describe("PasswordInput", () => {
         expect(validationCriteria.length).toBeGreaterThan(0);
     });
 
+    test("When validation is disabled, then no validation occurs", () => {
+        const invalidPassword = "a";
```
mowi12: Maybe create this constant at the same level as the validPassword istead of creating a local variable as there is at least 1 other test which uses an invalid password.

src/routes/(auth)/signin/+page.svelte
```diff
@@ -1,4 +1,92 @@
...
+        const userData = {
+            email: emailInput.getValue(),
+            password: passwordInput.getValue(),
+        };
+
+        backendService
+            .login(userData)
+            .then(() => {
+                console.log("User successfully logged in.");
+                goto("/");
```
mowi12: Same as in the signout. Maybe remove the signin page from the history?

```suggestion
                goto("/", { replaceState: true });
```

src/routes/(auth)/signin/+page.svelte
```diff
@@ -1,4 +1,92 @@
...
+    import { StatusCodes } from "$lib/model/error-codes";
+    import { goto } from "$app/navigation";
+
+    let emailInput: Input;
+    let passwordInput: PasswordInput;
+
+    let registrationError: ApiError | undefined = $state(undefined);
+
+    async function handleSubmit(event: Event) {
+        event.preventDefault();
```
mowi12, line 39: You should clear the errors at the start of the submission, because if you get an error the first time, but the second time enters correct input, the last error will still be displayed until the call completes.

src/routes/(auth)/signin/+page.svelte
```diff
@@ -1,4 +1,92 @@
...
+
+        const userData = {
+            email: emailInput.getValue(),
+            password: passwordInput.getValue(),
+        };
+
+        backendService
+            .login(userData)
+            .then(() => {
+                console.log("User successfully logged in.");
```
mowi12: Maybe remove this console statement as it is completely clear if the user was succesfully logged in -> redirect

src/routes/(auth)/signin/+page.svelte
mowi12, line 1: Imo you should disable the login button after you clicked it once and only enable it after the call was successful/error. So no multiple request are being send. Use the loading variant with the spinner?

tests/e2e/signin.test.ts
mowi12, line 1: Why are you using the isolated fixture for the signout process and here the shared?

tests/e2e/signin.test.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from "@playwright/test";
+import { test } from "./fixtures/shared-fixture";
+import { alice } from "./users";
+
+test.use({ user: null });
```
mowi12, line 5: Why is this needed?
So there is no initial user (represents a logged out state)?

tests/e2e/signin.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+test.describe("Sign In Functionality", () => {
+    test.beforeAll(async ({ apiClient }) => {
```
mowi12: Maybe delete this user afterAll? Or do you thing this is not necessary?

tests/e2e/signin.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    test("When the user is unauthenticated, then they will be redirected to /signin", async ({
+        page,
+    }) => {
+        await page.goto("/");
+        await page.waitForURL("/signin");
+        await expect(page.getByRole("heading", { name: "Sign In" })).toBeVisible();
+        await expect(page.getByLabel("Email")).toBeVisible();
+        await expect(page.getByLabel("Password", { exact: true })).toBeVisible();
```
mowi12, line 20: Why use exact here but not for the email?

tests/e2e/signin.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+
+    test("When the user enters invalid credentials, then an error is displayed", async ({
+        page,
+    }) => {
+        await page.getByLabel("Email").fill(alice.email);
+        await page.getByLabel("Password", { exact: true }).fill("ThisIsNotTheRightPassword");
```
mowi12: Same as above.

tests/e2e/signin.test.ts
```diff
@@ -0,0 +1,57 @@
...
+        await page.getByRole("button", { name: "Sign In" }).click();
+        await expect(page.getByRole("alert")).toBeVisible();
+    });
+
+    test("When the user enters valid credentials, then they are logged in", async ({ page }) => {
+        await page.getByLabel("Email").fill(alice.email);
+        await page.getByLabel("Password", { exact: true }).fill(alice.password);
```
mowi12, line 34: Same as above.

mowi12 (changes requested):
First of all, when you're already logged in and you navigate (whatever the reason) to the signin page, then the user should directly be redirected to the homepage.
Similar to [github.com/login](github.com/login).

But everything else is working fine!

There are also only some minor adjustments to finish this issue!

src/routes/(auth)/signin/+page.svelte
luca-schlecker: Won't be implemented immediately, like we discussed.

tests/e2e/signin.test.ts
luca-schlecker: Good catch, it should be isolated.

tests/e2e/signin.test.ts
luca-schlecker: Correct, the default fixtures come logged-in by default. This line prevents the automatic login and just provides a working mock backend.

tests/e2e/signin.test.ts
luca-schlecker: Not necessary anymore as I switched it to an isolated fixture.

tests/e2e/signin.test.ts
luca-schlecker: I just took the code from the old `global.setup.ts`. :sweat_smile: 
https://github.com/SE-UUlm/snowballr-frontend/blob/146af091355c09b697d808a5b58e0162fc8bb1ce/tests/e2e/global.setup.ts

tests/e2e/signin.test.ts
luca-schlecker: Same reason as above

tests/e2e/signin.test.ts
luca-schlecker: Same reason as above.

tests/e2e/signin.test.ts
mowi12: Okay, thanks.

tests/e2e/signin.test.ts
mowi12: Well, interesting. But this is no reason to do it as well. Could you please test if the `exact` option can be removed, and if yes, remove it? This makes the code easier to understand as there won't be the question why this `exact` is there.

mowi12 (changes requested):
There is one conversation you missed. Everything else is working fine!
On my laptop, if I click the back button, I still get directed to the /signin path. I thought this was disabled. Maybe I understood this option incorrectly.

On the other hand, I am redirected back to /, this is working well, although it is a bit slow imo.

luca-schlecker:
The `exact` is necessary as there would otherwise be multiple elements being resolved, leading to an error.

mowi12 (changes requested):
Sorry, but there is one last thing that really annoys me. If I'm logged in and navigate to `/signup` or `/signin` there is still a timeframe of about 1 second.
@domienderle and I talked about this and agreed that this is not optimal.

I propose the following:
Save the authentication and wrap the whole signin/signup dialog into an if block and only display this if the user is not authenticated. This way there is only a 'loading' screen (white screen) before the user is redirected. I really hope you see this the way we do.

Please update this

luca-schlecker:
> Sorry, but there is one last thing that really annoys me. If I'm logged in and navigate to `/signup` or `/signin` there is still a timeframe of about 1 second. @domienderle and I talked about this and agreed that this is not optimal.
> 
> I propose the following: Save the authentication and wrap the whole signin/signup dialog into an if block and only display this if the user is not authenticated. This way there is only a 'loading' screen (white screen) before the user is redirected. I really hope you see this the way we do.
> 
> Please update this

I like your solution of not showing the form, but I opted for opacity instead of `{#if}...{/if}` because of the bindings in the top level.

mowi12 (approved):
I finally could test all features combined (signout, signup, and signin) and everything worked really smooth. The opacity works great, and the redirects are fast. Great work!

PR #378, opened 2025-05-19
Add option to filter papers

### LINKED ISSUE #38: [Use Case]: Filter Papers

### LINKED ISSUE #386: [Bug]: Floating Nav Bar is Not Really Floating

_No response_

### What happened?

No content is allowed to be behind the _floating_ navigation menu, in my opinion defeating the point of it being floating. The content is cut off instead of rendering behind the nav bar.

Image

### Steps to reproduce

1. Go to a project.
2. Go to the `Papers` tab.
3. Scroll down (optionally with a paper preview currently open)

### What did you expect?

The nav bar should be really floating and thus content should be able to be rendered behind it.

Image

### Anything else?

_No response_

### LINKED ISSUE #387: [Bug]: Paper Preview Should Follow Scrolling

_No response_

### What happened?

In the `Papers` tab of a project, the paper preview pane should in my opinion not scroll out of view.

Image

### Steps to reproduce

1. Go to a project.
2. Select the `Papers` tab.
3. Click a Paper
4. Scroll down

### What did you expect?

The preview pane should stay in view.

Image

### Anything else?

_No response_

Merseleo:
Closes #38

## What I have made

- I updated the general `Select` component to 
  - have the option to select / unselect all possible options
  - display the selected values in the label / header (truncated at 50 characters at the moment)
- I slightly redesigned the search and filter bar on the papers view in a project: the filter and reset button are now at the right side and not left; furthermore the paper details card and the search / filter bar is sticky and does not scroll (closes #387 and closes #386)
- I fixed the problem, that the application dies if the mock backend is not running and no stages could be loaded
- I added the feature, that the selected filters and the current search text are stored in the URL as query parameters

- I adapted the markdown linting job to do not check the stack overflow link as it seems to be "forbidden" to automatically check this site

## Checklist

  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests


Merseleo:
At the moment I implemented the "validation" of query parameters in the URL by filtering all values that does not exist in the possible options. So if the user manually manipulates a query parameter to a value, that is not allowed, this query parameter is deleted. Do you agree with this implementation or do you have any better idea?

src/lib/components/composites/select/Select.svelte
```diff
@@ -24,8 +41,40 @@
...
+     * Select or unselect all options at once.
+     */
+    function toggleAllOptions() {
+        if (doSelectAllOptions) {
+            selectedValues = [];
+        } else {
+            selectedValues = [...options.map((option) => option.value), "all-options"];
+        }
+        doSelectAllOptions = !doSelectAllOptions;
+    }
```
Slartibartfass2: Currently, if we manually select or unselect all options, the "Select all"/"Unselect all" button doesn't change. We can simply add this feature by using an effect:
```svelte
    $effect(() => {
        if (selectedValues.length === 0) {
            doSelectAllOptions = false;
        } else if (selectedValues.length === options.length) {
            doSelectAllOptions = true;
        }
    })
```

src/lib/components/composites/select/Select.svelte
```diff
@@ -24,8 +41,40 @@
...
+
+    // Automatically drop any selected items that not exist in options, if the options are already
+    // loaded.
+    $effect(() => {
+        const invalid = selectedValues.filter((v) => !options.some((option) => option.value === v));
+        if (invalid.length > 0 && options.length > 0) {
+            // re-filter away the invalid ones and reassign
+            selectedValues = selectedValues.filter((v) =>
+                options.some((option) => option.value === v),
+            );
```
Slartibartfass2: Is there a reason, why don't we directly filter these?

I think I see it now.
What about storing the filtered values in a local variable and if it is different, then update the state?

src/lib/components/composites/select/Select.svelte
```diff
@@ -44,16 +93,24 @@ Usage:
     />
 ```
 -->
+
 <Select.Root type="multiple" bind:value={selectedValues}>
```
Slartibartfass2:
```suggestion
-->
<Select.Root type="multiple" bind:value={selectedValues}>
```

src/lib/components/composites/select/Select.svelte
```diff
@@ -44,16 +93,24 @@ Usage:
...
+        <Select.Content>
+            <Select.Item onclick={() => toggleAllOptions()} value="all-options">
+                {doSelectAllOptions ? "Unselect all" : "Select all"}
+            </Select.Item>
+            <Separator />
+            {#each options as option (option.value)}
+                <Select.Item value={option.value}>{option.label}</Select.Item>
+            {/each}
+        </Select.Content>
+    {/if}
```
Slartibartfass2: `Select.Content` can be outside the if condition.

What are your thoughts on adding the `Select.ScrollUpButton` and `Select.ScrollDownButton` components?

src/lib/components/composites/select/Select.svelte
```diff
@@ -24,8 +41,40 @@
...
 
-        return `${categoryLabel}: ${selectedValues.length} selected`;
+        let selectedValueLabel = selectedValues
+            .map((value) => options.find((option) => option.value === value)?.label)
+            .join(", ");
+        if (selectedValueLabel.length > MAX_SELECTED_VALUES_LABEL_LENGTH) {
+            selectedValueLabel = `${selectedValueLabel.substring(0, MAX_SELECTED_VALUES_LABEL_LENGTH - 3)}...`;
+        }
+
+        return `${categoryLabel}: ${selectedValueLabel} (${selectedValues.length})`;
```
Slartibartfass2, line 51: Update the component's documentation to match this behavior.

src/lib/utils/filters.ts
```diff
@@ -29,6 +30,59 @@ function filter<T>(
...
+ * This means for the stage, publisher, year or decision that the paper's stage, ... must be
+ * in the list of selected values of the filters.
+ * Regarding the reviewers or selected criteria, at least one review or selected criterion must be
+ * in the list of selected values of the filters.
+ *
+ * @param paper - The paper to evaluate against the filters.
+ * @param filters - The filters to apply.
+ * @returns True if the paper matches all filters, false otherwise.
+ */
+function matchesFilters(paper: Project_Paper, filters: ProjectPaperFilter) {
```
Slartibartfass2: As this file is not only about the paper filter functionality, what about naming this `matchesPaperFilters`?

src/lib/utils/filters.ts
```diff
@@ -29,6 +30,59 @@ function filter<T>(
...
+/**
+ * Checks if a given paper matches all active project paper filters.
+ *
+ * @remarks
+ * Only filter that have one or more selected values are applied, the remaining filters are ignored.
+ * A project paper must match all *active* filters.
+ * This means for the stage, publisher, year or decision that the paper's stage, ... must be
+ * in the list of selected values of the filters.
+ * Regarding the reviewers or selected criteria, at least one review or selected criterion must be
+ * in the list of selected values of the filters.
```
Slartibartfass2: I have a hard time understanding this. Could you please explain what you mean by that?

src/lib/utils/filters.ts
```diff
@@ -29,6 +30,59 @@ function filter<T>(
...
+        return false;
+    }
+
+    if (years.length > 0 && !years.includes(String(paper.paper?.year))) {
+        return false;
+    }
+
+    if (decisions.length > 0 && !decisions.includes(String(paper.decision))) {
+        return false;
+    }
```
Slartibartfass2: What are your thoughts on this refactoring?
```suggestion
    const matchesFilter = (filter: string[], value: string | bigint | number | undefined) =>
        filter.length === 0 || (value !== undefined && filter.includes(String(value)));

    if (!matchesFilter(stages, paper.stage)) {
        return false;
    }

    if (
        reviewers.length > 0 &&
        !paper.reviews.some((review) => reviewers.includes(review.userId))
    ) {
        return false;
    }

    if (!matchesFilter(publishers, paper.paper?.publisher)) {
        return false;
    }

    if (!matchesFilter(years, paper.paper?.year)) {
        return false;
    }

    if (!matchesFilter(decisions, paper.decision)) {
        return false;
    }
```

src/lib/utils/filters.ts
```diff
@@ -50,7 +104,7 @@ function filterPapers(allPapers: Paper[], searchText: string) {
 }
 
 /**
- * Filters project papers based on search text and sorts them by best match.
+ * Filters project papers based on filter and / or a search text and sorts them by best match.
```
Slartibartfass2:
```suggestion
 * Filters project papers based on filters and / or a search text and sorts them by best match.
```

src/lib/utils/filters.ts
```diff
@@ -59,11 +113,30 @@ function filterPapers(allPapers: Paper[], searchText: string) {
...
+    allProjectPapers: Project_Paper[],
+    filters?: ProjectPaperFilter,
+    searchText?: string,
+) {
+    // filter by project paper filter
+    if (filters) {
+        allProjectPapers = allProjectPapers.filter((paper) => matchesFilters(paper, filters));
+    }
+
+    // if no search text is given, only the filter are applied
```
Slartibartfass2:
```suggestion
    // if no search text is given, only the filters are applied
```

src/lib/utils/search-parameters.ts
```diff
@@ -0,0 +1,82 @@
...
+ * If the search text is empty, the query parameter is deleted.
+ *
+ * @remarks
+ * This function does not modify the browser's URL directly.
+ *
+ * @param searchText - The search text to set in the URL
+ * @param searchParams - The current URL search parameters
+ * @returns The updated search parameters with the added, deleted or updated 'searchText' value.
+ */
+export function updateSearchTextInURL(
```
Slartibartfass2: If the function doesn't update the URL, what about `updateSearchTextParam` instead?

src/lib/utils/search-parameters.ts
```diff
@@ -0,0 +1,82 @@
...
+ * If a filter array is empty, the corresponding query parameter is deleted.
+ *
+ * @remarks
+ * This function does not modify the browser's URL directly.
+ *
+ * @param filters - The object containing currently applied filter values
+ * @param searchParams - The current URL search parameters
+ * @returns The updated search parameters with the added, deleted or updated filter values.
+ */
+export function updateFiltersInURL(
```
Slartibartfass2: Same here with the function name.

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -39,36 +49,47 @@
...
 
-    interface PapersFilters {
-        stages: string[];
-        reviewers: string[];
-        publishers: string[];
-        years: string[];
-        decisions: string[];
-        criteria: string[];
-    }
+    let showFilters = $state(false);
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/papers/+page.svelte:58]: Can we show the filters when there are filters applied? This would seem to me like a logical use case. Otherwise, the use might wonder why not all papers are shown.

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -85,30 +106,30 @@
...
+                        {#if showFilters}
+                            <ChevronUp class="size-4" />
+                        {:else}
+                            <ChevronDown class="size-4" />
+                        {/if}
+                    </Button>
+                    <Button onclick={() => (papersFilters = emptyFilters)}>
+                        <Trash />
+                        Reset
+                    </Button>
```
Comment (Slartibartfass2) [inline src/routes/project/[projectId]/papers/+page.svelte:136]: I thought the idea was to also clear the search text. What is the current idea?

tests/e2e/papers-overview.test.ts
```diff
@@ -293,4 +291,92 @@ test.describe("View all papers of a project", () => {
...
+        await projectPapersPage.applyFilter(0, 1990);
+
+        await projectPapersPage.expectStageCounts(0, `(1 / ${NUM_PAPERS_PER_STAGE} papers)`);
+        await projectPapersPage.expectStageCounts(1, `(0 / ${NUM_PAPERS_PER_STAGE} papers)`);
+
+        await expect(projectPapersPage.getPaperByTitle(matchingPaper)).toBeVisible();
+        await expect(projectPapersPage.getNoSearchResultsText().last()).toBeVisible();
+    });
+
+    test("When the user selects filter that no paper matches, then the 'no results' message is shown.", async ({
```
Slartibartfass2:
```suggestion
    test("When the user selects a filter that no paper matches, then the 'no results' message is shown.", async ({
```

tests/e2e/papers-overview.test.ts
```diff
@@ -293,4 +291,92 @@ test.describe("View all papers of a project", () => {
...
+        // Check counts
+        await projectPapersPage.expectStageCounts(0, `(0 / ${NUM_PAPERS_PER_STAGE} papers)`);
+        await projectPapersPage.expectStageCounts(1, `(0 / ${NUM_PAPERS_PER_STAGE} papers)`);
+
+        // Check messages
+        await expect(projectPapersPage.getNoSearchResultsText().first()).toBeVisible();
+        await expect(projectPapersPage.getNoSearchResultsText().last()).toBeVisible();
+    });
+
+    test("When the user clears the filter using the reset button, then all papers are shown again.", async ({
```
Slartibartfass2:
```suggestion
    test("When the user clears the filters using the reset button, then all papers are shown again.", async ({
```

tests/e2e/papers-overview.test.ts
```diff
@@ -293,4 +291,92 @@ test.describe("View all papers of a project", () => {
...
+        await expect(projectPapersPage.getPaperByTitle(nonMatchingPaper)).toBeHidden();
+
+        await projectPapersPage.clearFiltersButton.click();
+
+        // Verify reset state
+        await projectPapersPage.expectStageCounts(0, `(${NUM_PAPERS_PER_STAGE} papers)`);
+        await projectPapersPage.expectStageCounts(1, `(${NUM_PAPERS_PER_STAGE} papers)`);
+    });
+
+    test("When the user reloads the page, the search text is restored from the URL.", async ({
```
Slartibartfass2:
```suggestion
    test("When the user reloads the page, then the search text is restored from the URL.", async ({
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
...
+            "then the same list of papers is returned ",
+        async () => {
+            const papers = [Papers.demoPaper1, Papers.demoPaper2, Papers.demoPaper3];
+
+            expect(filterPapers(papers, "")).toStrictEqual(papers);
+        },
+    );
+
+    test(
+        "When papers are filtered, only papers where the id, title or authors match " +
```
Slartibartfass2:
```suggestion
        "When papers are filtered, then only papers where the id, title or authors match " +
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
...
+                ProjectPapers.demoProjectPaper2,
+                ProjectPapers.demoProjectPaper3,
+            ];
+
+            expect(filterProjectPapers(projectPapers)).toStrictEqual(projectPapers);
+        },
+    );
+
+    test(
+        "When project papers are filtered by search text, only papers where the id, title or " +
```
Slartibartfass2:
```suggestion
        "When project papers are filtered by search text, then only papers where the id, title or " +
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
...
+
+            // test separate search starting with '#' to search directly by paper id
+            expect(filterProjectPapers(projectPapers, undefined, "#0")).toStrictEqual([
+                ProjectPapers.demoProjectPaper1,
+            ]);
+        },
+    );
+
+    test(
+        "When project papers are filtered by specific filter arrays, only papers conforming all" +
```
Slartibartfass2:
```suggestion
        "When project papers are filtered by specific filter arrays, then only papers conforming all" +
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
...
+                decision: PaperDecision.UNREVIEWED,
+                paper: createPaper({ publisher: "IEEE Explore", year: 2013 }),
+            });
+            expect(filterProjectPapers([{ ...projectPaper, ...property }], filters)).toStrictEqual(
+                [],
+            );
+        },
+    );
+
+    test("When a project paper match all filters, then it is not filtered out", () => {
```
Slartibartfass2:
```suggestion
    test("When a project paper matches all filters, then it is not filtered out", () => {
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
...
+            "then the same list of users is returned ",
+        async () => {
+            const users = [Users.johnDoe, Users.janeDoe, Users.aliceSmith, Users.bobJohnson];
+
+            expect(filterUsers(users, "")).toStrictEqual(users);
+        },
+    );
+
+    test(
+        "When users are filtered, only users matching the search string - " +
```
Slartibartfass2:
```suggestion
        "When users are filtered, then only users matching the search string - " +
```

tests/unit/utils/filters.test.ts
```diff
@@ -0,0 +1,161 @@
+import { describe, expect, test } from "vitest";
+import { Papers, ProjectPapers, Users } from "$tests/example-data";
+import { filterPapers, filterProjectPapers, filterUsers } from "$lib/utils/filters";
+import { PaperDecision, type Project_Paper } from "$lib/model/api/project";
+import { createPaper, createProjectPaper, createReview } from "$tests/model-builder";
+
+describe("Custom filter", () => {
```
Slartibartfass2: What about using a `describe` block for each filtering method, similar to `common-helper.test.ts`?

Slartibartfass2 (changes requested):
Great work! I see that you put great effort into this. The functionality feels good, and I like that we now can filter papers. This is another step to a good-looking and working frontend.

While the number of comments I left tell otherwise, there are only minor things of which, I think, need to be changed. Most of them are typos/other names.

Next to my comments, I have these three questions/comments left:
1. Do you think we should add a filter for authors? Is there a use case? If so, we can create an issue for it.
2. Since the new way of how the IDs work, the ID sorting doesn't work anymore. There's an easy fix for it: Go to the `+page.ts` file and, on line 75, sort according to the **integer value** of the `localId` property. `comparePaperId` has to be changed accordingly.
3. It seems like not every reviewer is shown as a filter. Do you know if this is just how the mock backend works, or is this a bug right there?

That's all. Again, great job 💪🏼

tests/unit/utils/filters.test.ts
Merseleo: Good idea.

src/lib/utils/search-parameters.ts
Merseleo: Yeah, probably a good idea.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte:136]: This is right that this was the idea, however, I tested it and imo it would be nice to have the opportunity to only reset all filters and keep the search text. Or do you want me to reset everything?

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte:58]: Great idea, I noticed too that it is weird that not all papers are shown but did not consider to show the filters when the filters are applied.

src/lib/utils/filters.ts
Merseleo: Good idea.

src/lib/utils/filters.ts
Merseleo: I agree, your code looks a bit cleaner. I will refactor it.

src/lib/utils/filters.ts
Merseleo: I rewrote it a bit, is it now clearer? Because I do not quite understand what is not understandable in this comment.

src/lib/components/composites/select/Select.svelte
Merseleo: You are right. I missed to update the components comment.

src/lib/components/composites/select/Select.svelte
Merseleo: I already implemented them and played around with these buttons and imo they somehow feel a bit strange as no scrollbar is shown and they have a high sensitivity that can not be changed afaik, so I decided to remove them again. Do you think we should add them?

src/lib/components/composites/select/Select.svelte
Merseleo: Yeah there is and the reason is described in the comment below the effect. 
What do you mean with storing the filtered values in a local variable? The filtered values are the `selectedValues` and automatically binded to the `Select.Item` component? Or do you speak of the `invalid` array? If so, I do not know why this could be useful. 
Can you explain a bit more?

src/lib/components/composites/select/Select.svelte
Merseleo: Good catch.

Merseleo (commented):
Thank you for the review and your ideas.
Regarding your questions: 
> 1. Do you think we should add a filter for authors? Is there a use case? If so, we can create an issue for it.

No imo I do not see any use case for filtering by authors at the moment or in the future. Considering a real project, I think there could be hundreds of authors and the select list would be very long. Furthermore, I would search by the authors name if I want to find something from a specific author as one author doesn’t publish that much papers regarding a specific topic. So I would suggest to wait for user feedback and if some users request this, we could still easily add another filter. 

> 2. Since the new way of how the IDs work, the ID sorting doesn't work anymore. There's an easy fix for it: Go to the `+page.ts` file and, on line 75, sort according to the **integer value** of the `localId` property. `comparePaperId` has to be changed accordingly.

You are right, thank you for that hint. 

> 3. It seems like not every reviewer is shown as a filter. Do you know if this is just how the mock backend works, or is this a bug right there?

Yep this is due to the behavior of the mock backend. The reviewers in the filter are all project members so theoretically every person that could reviewed a paper in this project. However, the mock backend does not check, whether a person is in the project in which the paper that was „reviewed“ is in, but the reviewing persons are random persons afaik.

PS: Do you an idea, why the `Testing.md` file now has a line that is too long? I did not change anything regarding the Wiki?!

Slartibartfass2:
> PS: Do you have an idea, why the `Testing.md` file now has a line that is too long? I did not change anything regarding the Wiki?!

Ah yes, I do. In the last PR that was merged to develop, I adjusted the wiki github action so that the GitHub source path URLs are changed from develop to the current branch. This makes it possible to detect issues in the PR and not after the PR is merged to develop. While this works great, replacing "develop" with the current branch name in GitHub URLs, makes these URLs longer given the branch name is longer.

So this is what raises that error. You can fix this by moving the words at the end of the line into the next line.
I'm not sure if there's another good way to solve this. Probably, we should lint the wiki before making the modification:
1. lint markdown files
2. modify URLs
3. check for broken links

Edit: You can, of course, just move the order in the workflow file.

Merseleo:
> > PS: Do you have an idea, why the `Testing.md` file now has a line that is too long? I did not change anything regarding the Wiki?!
> 
> Ah yes, I do. In the last PR that was merged to develop, I adjusted the wiki github action so that the GitHub source path URLs are changed from develop to the current branch. This makes it possible to detect issues in the PR and not after the PR is merged to develop. While this works great, replacing "develop" with the current branch name in GitHub URLs, makes these URLs longer given the branch name is longer.
> 
> So this is what raises that error. You can fix this by moving the words at the end of the line into the next line. I'm not sure if there's another good way to solve this. Probably, we should lint the wiki before making the modification:
> 
> 1. lint markdown files
> 2. modify URLs
> 3. check for broken links
> 
> Edit: You can, of course, just move the order in the workflow file.

I just figured out the solution myself ^^. I reordered the steps as you described and it worked. And imo this is the better solution.

Slartibartfass2:
> > > PS: Do you have an idea, why the `Testing.md` file now has a line that is too long? I did not change anything regarding the Wiki?!
> > 
> > 
> > Ah yes, I do. In the last PR that was merged to develop, I adjusted the wiki github action so that the GitHub source path URLs are changed from develop to the current branch. This makes it possible to detect issues in the PR and not after the PR is merged to develop. While this works great, replacing "develop" with the current branch name in GitHub URLs, makes these URLs longer given the branch name is longer.
> > So this is what raises that error. You can fix this by moving the words at the end of the line into the next line. I'm not sure if there's another good way to solve this. Probably, we should lint the wiki before making the modification:
> > 
> > 1. lint markdown files
> > 2. modify URLs
> > 3. check for broken links
> > 
> > Edit: You can, of course, just move the order in the workflow file.
> 
> I just figured out the solution myself ^^. I reordered the steps as you described and it worked. And imo this is the better solution.

Yes, I didn't think of that when I was making the modifications to the CI.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte:136]: I think this is fine for now. Maybe get some user feedback and then learn what feels best.

src/lib/components/composites/select/Select.svelte
Slartibartfass2: This works great now.

src/lib/components/composites/select/Select.svelte
Slartibartfass2: I just saw that they are there by default. I think it's good to cap the height at some point. I am not sure if we could make this any better.

src/lib/utils/filters.ts
Slartibartfass2: I think it's better now. Only change "Only filter that ..." to "Only filters that ...".

src/lib/components/composites/select/Select.svelte
Slartibartfass2: Not sure what I meant by that. But I know it wasn't that important.

Slartibartfass2 (changes requested):
We reached a superb stage of this functionality. Great job!

Only these minor things left.

src/lib/utils/filters.ts
Merseleo: I changed it and rebased it into the commit where I rewrote it.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Merseleo) [reply src/routes/project/[projectId]/papers/+page.svelte:136]: So do you want me to test it now or do we move this into the future when we lets users test everything?

src/lib/components/composites/select/Select.svelte
Merseleo: You are right. I limited the height to 300px, so the select component does not get too big and the scroll behavior is implemented by default.

Merseleo (commented):
I hope I did not miss anything. If not so, I think we are done?

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte:136]: No, there should be a "we" after the "Maybe". Let's keep this as is.

Slartibartfass2 (approved):

PR #384, opened 2025-05-20
Make user object global context

### LINKED ISSUE #340: [Feature]: Make user object global context

_No response_

Currently, the user object must be passed manually through multiple component layers, leading to deep prop drilling and tightly coupled component structures.

I propose introducing a global context for the user object. This would make the user data available throughout the application without the need for explicitly passing it as a prop.

**Benefits**:

- Eliminates the need for deeply nested prop structures.
- Simplifies component interfaces and improves readability.
- Encourages cleaner and more modular component design.
- Improves maintainability and scalability of the application.

_No response_

### Anything else?

_No response_

mowi12:
Closes #340 

## What I have made

- I added a context which stores the user in the case, the user is logged in
- Removed user propagation
- Updated all impacted components
- I updated the integration, and e2e tests

## Checklist

  - [ ] ~~unit tests~~
  - [x] integration tests
  - [ ] ~~end-to-end tests~~

### DIFF
--- a/src/lib/components/composites/navigation-bar/NavigationBar.svelte
+++ b/src/lib/components/composites/navigation-bar/NavigationBar.svelte
@@ -4,11 +4,9 @@
     import * as Tabs from "$lib/components/primitives/tabs/index.js";
     import type { Snippet } from "svelte";
     import UserMenu from "./UserMenu.svelte";
-    import type { User } from "$lib/model/api/user";
     import type { LinkTab } from "$lib/model/tabs";
 
     interface Props {
-        user: User;
         backRef?: string | undefined;
         tabs?: LinkTab[] | undefined;
         defaultTabValue?: (typeof tabs)[number]["value"] | undefined;
@@ -16,7 +14,6 @@
     }
 
     const {
-        user,
         backRef = undefined,
         tabs = [],
         defaultTabValue = "",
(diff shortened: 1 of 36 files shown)


domienderle (changes requested):
Since you mostly just removed or replaced the user, this looks good. But there is one major question that came up for me. Is there a reason, why you use the `getContext` and `setContext`. As I understood, the user can easily be propagated down the components, but gets lost on a page reload. The alternative would be `localStorage`/`sessionStorage`. I am not sure how this would behave in context of reactivity, but this would ensure, that the page can be reloaded or even reopened after closing and still show the same user, which I would prefer in context of usability. Additionally, this is how we store for example `review-mode-state` oand others as well. But maybe you already tried this or it isn't possible due to the reactivity. Let me hear your thoughts about this.

mowi12:
Merseleo and I talked a lot about this topic. It is far too complex to explain it via text, but here are two articles which explain this:

- https://joyofcode.xyz/master-the-svelte-context-api#global-state-versus-context
- https://joyofcode.xyz/avoid-sharing-server-and-client-state-in-sveltekit

But in short, if you use global state or session storage, then this can be exploited. Context is not 'visible' from the outside and is much safer for this use case. This use case being the 'storage' of the user object itself.

You can test this yourself by trying to find anything related to the user data in you storage/memory tabs in the inspector, you will only find the global states, but nothing about the user.

domienderle (dismissed):

domienderle (commented):
Please rebase, after that we are good to go.

domienderle (approved):

PR #385, opened 2025-05-20
An effort to fix the flakiness of Pipeline E2E tests

luca-schlecker:
## What I have made

- Increase timeouts 
- Pull docker image of mock backend beforehand
- Implement possible fix for the obscure ERR_NETWORK_CHANGED error found here: https://github.com/plan-player-analytics/Plan/issues/3535

## Checklist

  - [ ] unit tests
  - [ ] integration tests
  - [x] end-to-end tests

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -12,7 +12,7 @@ export default defineConfig({
         ["html", { outputFolder: "e2e-report" }],
     ],
     expect: {
-        timeout: 2_500,
+        timeout: 7_500,
     },
     timeout: 15_000,
     use: {
--- a/tests/e2e/mock-backend-version.js
+++ b/tests/e2e/mock-backend-version.js
@@ -1,0 +1,1 @@
+export const MOCK_BACKEND_IMAGE = "ghcr.io/se-uulm/snowballr-mock-backend:main";
--- a/tests/e2e/mock-backend.ts
+++ b/tests/e2e/mock-backend.ts
@@ -7,6 +7,7 @@ import crossFetch from "cross-fetch";
 import cookieFetch from "fetch-cookie";
 import { StatusCodes } from "$lib/model/error-codes";
 import { Nothing } from "$lib/model/api/base";
+import { MOCK_BACKEND_IMAGE } from "./mock-backend-version";
 
 async function isDockerInstalled(): Promise<boolean> {
     return new Promise((resolve) => {
@@ -100,8 +101,6 @@ export class AuthSnowballRClient extends SnowballRClient {
     }
 }
 
-const MOCK_BACKEND_IMAGE = "ghcr.io/se-uulm/snowballr-mock-backend:main";
-
 /**
  * Manages a docker-backed mock backend.
  * Use `DockerMockBackend.create()` instead of trying to construct a new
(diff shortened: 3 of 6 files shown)


playwright.config.ts
```diff
@@ -12,9 +12,9 @@ export default defineConfig({
         ["html", { outputFolder: "e2e-report" }],
     ],
     expect: {
-        timeout: 2_500,
+        timeout: 5_000,
     },
-    timeout: 15_000,
+    timeout: 30_000,
```
mowi12: Is the increase of the general timeout to 30 seconds really necessary? Slartibartfass2 reduced this to 15 seconds not long ago to reduce the time the e2e take.

I understand why you want to increase the timeout, but if a test only succeeds after 15 seconds, how valuable can this test be?

The frontend is already built before the test, so only the test itself and the interaction should take this long. Imo 15 is already long enough. If a test fails due to this, then the pipeline should fail so we can fix the potential reason, why this test is actually slow, e.g. the implementation in the frontend is just bad.

tests/e2e/mock-backend.ts
mowi12, line 1: As we discussed in our Sprint meeting today, maybe create a directory for all the mock-backend files? This would keep the e2e tests more easier to navigate.

tests/e2e/mock-backend-version.js
mowi12, line 1: What exactly are the `mock-backend-version.js` and `mock-backend-version.d.ts` needed? Why don't you do i as before, and just declare this string in the `mock-backend.ts` file?

mowi12 (changes requested):
Nicely done! I cannot really test your changes, but you documented everything important with comments. And as I understand, you tested this in the pipeline for a few tries and everything worked.

I only have some questions and one minor change request, but then we are good to go!

playwright.config.ts
luca-schlecker: I'll drop this modification, and we'll see how it turns out. You're probably right that 15 seconds should be enough, although 2.5s for `expect` could be a little low.

tests/e2e/mock-backend.ts
luca-schlecker: These are three files which aren't even tests. I'd prefer just keeping them there and wait for the broader restructuring of the e2e test folder structure.

tests/e2e/mock-backend-version.js
luca-schlecker: As the docker image is pulled outside the tests, the specific image has to be accessible to `scripts/run-playwright.js`. I'd prefer to maintain a single source of truth for this information, and am thus required to put it inside a `.js` file, as `.js` files cannot import `.ts` files but `.ts` files can import `.js` files using a `.d.ts`-helper.

playwright.config.ts
mowi12: I actually like the 5 seconds you chose for the `expect` timeout. I myself had issues with the 2.5 seconds timeout and since the `reload()` function took about 4.5 seconds to load and therefore the `expect()` did not complete.

I suggest adding this timeout back : D

tests/e2e/mock-backend-version.js
mowi12: Okay, thanks!

mowi12 (changes requested):
You now removed a timeout I didn't even want you to remove. Could you please add the timeout back? Then this PR is done!

playwright.config.ts
luca-schlecker: If you measured reload to be 4.5 seconds, I'd like to set this timeout a bit higher to have some headroom.

mowi12: Okay.

mowi12 (approved):
Lets ship this!

PR #395, opened 2025-05-23
Make criteria lists scrollable when there's not enough space

### LINKED ISSUE #347: [Bug]: Overflow of criteria list not always hidden

_No response_

### What happened?

When I reduced the size of my browser window, especially in height, the overflow of the criteria list was not hidden, but went beyond the border of the paper card:

Image

### Steps to reproduce

1. Go to the paper view of any paper
2. Decrease the height of the browser window
3. Scroll down to see the end of the criteria list
4. Decrease the height of the browser window further

### What did you expect?

I would expect that the remaining part of criteria list that could not be displayed because of missing vertical space is hidden and does not overflow. 

### Anything else?

_No response_

Slartibartfass2:
Closes #347 

## What I have made
Here, we simply use the NamedList component as this already implements this behavior. We only need to add the overflow classes so that everything works together.

## Checklist

### DIFF
--- a/src/lib/components/composites/criteria/CriteriaList.svelte
+++ b/src/lib/components/composites/criteria/CriteriaList.svelte
@@ -3,7 +3,7 @@
     import type { CriterionWithReviews } from "$lib/model/general";
     import type { User } from "$lib/model/api/user";
     import CriterionListEntrySkeleton from "./CriterionListEntrySkeleton.svelte";
-    import ErrorIndicator from "../utils/ErrorIndicator.svelte";
+    import NamedList from "../list/NamedList.svelte";
 
     interface Props {
         listTitle: "Hard Exclusion" | "Soft Exclusion" | "Inclusion";
@@ -20,6 +20,10 @@
         numberOfSkeletons = 2,
         emptyHint,
     }: Props = $props();
+
+    const items = Promise.all([loadingReviewers, loadingCriteria]).then(([reviewers, criteria]) =>
+        criteria.map((criterion) => ({ criterion, reviewers })),
+    );
 </script>
 
 <!--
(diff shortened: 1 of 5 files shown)


mowi12 (changes requested):
I tested your implementation and think that there are adjustments you need to make.

Screenshot 2025-05-24 at 09 51 56

If you take a look at this picture, you can see that, yes, the inclusion criteria are inside a scrollable window, but the other ones are not. They are not visible and also not scrollable.

And the other point is visible on the left side. I am aware that this issue only regards the criteria list, but since the left side has the same problem, maybe we can extend the issue. If not, please create a new one.
As you can see, the General Information/Abstract are also overflowing.

Slartibartfass2:
> I tested your implementation and think that there are adjustments you need to make.
> 
> Screenshot 2025-05-24 at 09 51 56
> 
> If you take a look at this picture, you can see that, yes, the inclusion criteria are inside a scrollable window, but the other ones are not. They are not visible and also not scrollable.
> 
> And the other point is visible on the left side. I am aware that this issue only regards the criteria list, but since the left side has the same problem, maybe we can extend the issue. If not, please create a new one.
> As you can see, the General Information/Abstract are also overflowing.

Hi there, in general I totally agree with you. In this case, the height is very small and I'm wondering how to deal with this. The newly introduced fix, fixes the height problem for a lot of space until the first line of the abstract is reached. There, we don't have another strategy.
In my opinion, this already solves a lot of problems and I don't think we don't need to support even smaller screens as we didn't plan on doing this.
What are your thoughts about this? Where do you see room for improvement?

mowi12 (approved):
You're right. It makes no sense to optimize for heights/widths which are never used, or at least not the focus of this application.
Since the abstract field (textarea) itself is already scrolling, I don't think we'll ever encounter this special case of overflow.

PR #399, opened 2025-05-23
Go to next and previous paper

### LINKED ISSUE #46: [Use Case]: Go To Previous Paper

Project Member

### Describe the use case

As project member, I want to go to the previous paper of a specific paper in the paper view.

### Acceptance criteria

- provide button to go to the previous paper
- show paper view of previous paper after clicking on button
- if there's no previous paper disable button

### Anything else?

Related to #34 and #47

### LINKED ISSUE #47: [Use Case]: Go To Next Paper

Project Member

### Describe the use case

As project member, I want to go to the next paper of a specific paper in the paper view.

The next paper should be either:
 - the paper with the successor local paper id (when not reviewing)
 - the paper that should be reviewed next (when reviewing)

The paper that should be reviewed next, is
- the paper with the successor local paper id that is unreviewed and in the same stage
- if no such paper exists, the paper with the successor local paper id that is unreviewed and in the next stage ...
- if no unreviewed paper exists in the current project, I want to be redirected back to the homepage / dashboard and see the hint that no open review exists for this project anymore

### Acceptance criteria

- provide button to go to the next paper
- show paper view of next paper after clicking on button
- if there's no next paper disable button

### Anything else?

Related to #34 and #46

domienderle:
Closes #46 and closes #47 

## What I have made

I have added the functionality to the buttons to go to the next/previous `ProjectPaper` (this buttons do not exist in the `PaperView` for normal papers, thus the functionality is only necessary for `ProjectPaper`). 
 - Therefore I added an `$effect` in the `PaperView` to make sure, that every ressource is loaded before one of the buttons can be clicked. While loading the `PaperNavigationButton` simultaneously the next and previous `ProjectPaper` is pre-loaded to examine if the according button is enabled and disabled and to speed up the navigation if one of the buttons is clicked. 
 - Additionally I added the navigation functionality to the `PaperDecisionButtons` to directly navigate to the next not review paper if one of the `PaperDecisionButtons` is clicked (if a next paper to review exists).

## Checklist

  - ~~[ ] unit tests~~ logic in the backend
  - [x] integration tests
  - [x] end-to-end tests


src/routes/project/[projectId]/paper/[paperId]/+page.svelte
```diff
@@ -10,7 +10,7 @@
         forwardReferencedPapers,
         reviewers,
         criteriaWithReviews,
-    } = data;
+    } = $derived(data);
```
Comment (mowi12) [inline src/routes/project/[projectId]/paper/[paperId]/+page.svelte:13]: Why does this need to be a state?

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -185,42 +217,71 @@ Usage:
     {#if showButtonBar}
         <div class="flex h-fit w-full flex-row justify-between gap-4" data-testid="button-bar">
             <!-- TODO: Implementation of navigation buttons will be done in #46 and #47 -->
```
mowi12: Please remove the comment with the todo note.

src/lib/utils/paper-navigation.ts
```diff
@@ -0,0 +1,27 @@
+import { goto } from "$app/navigation";
+import type { Project, Project_Paper } from "$lib/model/api/project";
+
+/**
+ * Handles the navigation of the button with the direction "right". Therefore, it is checked whether the
+ * review mode is activated or not.
+ */
```
mowi12: Please review this comment. I don't think it is quite right.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -108,11 +136,14 @@
...
-    const loadingUserReview = getUserReviewIfAlreadySubmitted();
-    loadingUserReview.then((review) => {
-        const selectedCriteria = review?.selectedCriteriaIds ?? [];
-        selectedReviewCriteria.criteria = selectedCriteria;
-        wasAlreadyReviewedState.wasReviewed = review !== undefined;
+
+    const loadingUserReview = $derived.by(() => getUserReviewIfAlreadySubmitted());
+    $effect(() => {
+        (async () => {
+            const review = await loadingUserReview;
```
mowi12: Maybe put a try catch around this block or use await ..then ..catch syntax

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -95,6 +98,31 @@
...
     });
 
+    let loading = $state(true);
+    let paperQueue = $state([]);
+    let nextProjectPaper: Project_Paper | undefined = $state(undefined);
+
+    /**
+     * Makes sure, that all necessary ressources are loaded.
+     */
+    $effect(() => {
```
mowi12, line 112: Why do we need an effect here?
If we click on any next or previous button or choose a review, we are getting redirected. Isn't the component reloaded, or mounted again?
If not, then it would make sense, but I'm not quite sure.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
+The loadingProjectPaper has to be the Promise of the project paper or undefined if no
+such paper could be found.
 
 Usage:
 ```svelte
     <PaperNavigationButton
         direction="left"
-        href="/papers/1"
-        onClick={() => console.log("button was clicked")}
+        {loadingProjectPaper}
```
mowi12: I think you need to at some variables.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
+    async function getNextProjectPaperToReview(paper: Project_Paper) {
+        await backendService
+            .getNextPaperToReview({ id: paper.id })
+            .response.then((nextPaper) => {
+                nextProjectPaper = nextPaper;
+            })
+            .catch((error) => {
+                nextProjectPaper = undefined;
+                if (error.message !== "No%20next%20paper%20available.") {
+                    toast("Error, while loading the next paper:" + error);
```
mowi12:
```suggestion
                    console.error("Error while loading the next paper:" + error);
                    toast("Error while loading the next paper);
```

Same for the ones below.
I thought we said we still include the console.error part.
And imo, the error itself should not be printed into the toast, as this is only for the user.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
+            });
+    }
+
+    /**
+     * Pre-fetches the next and the previous paper to find out if the according button is enabled or
+     * disabled and the according paper does not have to be loaded on clicking of the button, but
+     * before.
+     */
+    $effect(() => {
+        // Loads the next or previous paper only if all necessary information is available.
```
mowi12: Why not move this comment into the description above the method?

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
+                buttonLeftDisabled = previousProjectPaper === undefined;
+            }
+            showLoadingSpinner = false;
+        })();
+    });
+
+    /**
+     * Handles the navigation of the button with the direction "right". Therefore, it is checked whether the
+     * review mode is activated or not.
+     */
```
mowi12: Again, this comment does not look quite right.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
+        loading = showLoadingSpinner = true;
+        await navigatePaper(
+            direction,
+            loadingProjectPaper,
+            paperQueue,
+            loadingProject,
+            nextProjectPaper,
+            previousProjectPaper,
+        );
+    };
```
mowi12: Why not:

```suggestion
    async function navigate() {
        loading = showLoadingSpinner = true;
        await navigatePaper(
            direction,
            loadingProjectPaper,
            paperQueue,
            loadingProject,
            nextProjectPaper,
            previousProjectPaper,
        );
    };
```

And then below:

```
onclick={navigate}
```

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,157 @@
...
 <Tooltip
     class="text-primary max-w-xs min-w-32 grow bg-slate-200 shadow-lg hover:bg-slate-400"
     aria-label={tooltipText}
     data-testid="navigation-button"
-    onclick={() => {
-        if (onClick) onClick();
-        goto(href);
-    }}
+    disabled={loading || (direction === "right" ? buttonRightDisabled : buttonLeftDisabled)}
+    onclick={() => navigate()}
```
mowi12: As mentioned above `onclick={navigate}`

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -110,7 +122,15 @@
             await backendService.createReview(review);
 
             toast.success("Successfully submitted a review.");
-            // TODO: navigate automatically to next paper that will be implemented in #47
+            loading = true;
```
mowi12: Why do we set loading here and not in the beginning of the function?

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -110,7 +122,15 @@
...
+            await navigatePaper(
+                "right",
+                loadingProjectPaper,
+                paperQueue,
+                loadingProject,
+                nextProjectPaper,
+                undefined,
+            );
         } catch (err) {
```
mowi12, line 143: Don't we need a `finally` block to set `loading` to `false` ?

tests/integration/paper-components/paper-view/paper-navigation-button.test.ts
```diff
@@ -2,70 +2,97 @@ import { render, screen, waitFor } from "@testing-library/svelte";
...
-
         await userEvent.hover(button);
-
         await waitFor(() => {
             expect(button).toHaveAttribute("data-state", "delayed-open");
         });
-
         const tooltip = screen.getByText("Previous Paper");
         expect(tooltip).toBeInTheDocument();
     });
```
mowi12: Why are you removing all the space? Imo it makes sense to structure your code into steps.
I thought the same about some of your source code, but there it wasn't that important.
But since I now see, that you are atively removing this space/structuring, I thought I mentioned it.

If you want you can look at your own code again and maybe at a line to separate different parts :)

tests/integration/paper-components/paper-view/paper-navigation-button.test.ts
```diff
@@ -2,70 +2,97 @@ import { render, screen, waitFor } from "@testing-library/svelte";
...
         const button = screen.getByRole("button");
         expect(button).toHaveAttribute("aria-label", "Next Paper");
-
         await userEvent.hover(button);
-
         await waitFor(() => {
             expect(button).toHaveAttribute("data-state", "delayed-open");
         });
-
         const tooltip = screen.getByText("Next Paper");
```
mowi12: Same as above.

tests/e2e/pom/paper-view-model.ts
```diff
@@ -22,6 +24,8 @@ export class DevPaperViewPage {
         this.referenceListEntry0 = this.getFirstListEntry("link")
             ? this.getFirstListEntry("link")
             : this.getFirstListEntry("button");
+        this.nextPaperButton = page.locator("button[aria-label='Next Paper']");
+        this.previousPaperButton = page.locator("button[aria-label='Previous Paper']");
```
mowi12: Maybe try to use `getByRole()` or other more descriptive methods, same as above.

tests/e2e/pom/paper-view-model.ts
```diff
@@ -42,7 +46,7 @@ export class DevPaperViewPage {
      */
     async openProjectPaperView(projectId: string, paperId: string) {
         await this.page.goto(`/project/${projectId}/paper/${paperId}`);
-        await expect(this.acceptButton).toBeVisible();
+        await expect(this.nextPaperButton).toBeVisible();
```
mowi12: Are you changing this becuase the review mode is not always on?

tests/e2e/fixtures/paper-view-fixture.ts
```diff
@@ -1,6 +1,5 @@
 import { test as base } from "./shared-fixture";
 import { DevPaperViewPage } from "$tests/e2e/pom/paper-view-model";
-
```
mowi12: Imo, this line should stay.

tests/e2e/fixtures/paper-view-fixture.ts
```diff
@@ -11,6 +10,7 @@ type PaperViewFixture = {
  */
 export const test = base.extend<PaperViewFixture>({
     paperViewPage: async ({ page }, use) => {
+        await page.goto("/");
```
mowi12: If this line is important, although I do not quite understand why you want to navigate to the homepage for the `PaperViewFixture`, then await the SnowballR label (see in homepage fixture).

tests/e2e/decide-on-paper.test.ts
mowi12, line 1: Maybe test the case if the user has decided the last paper and nothing else needs to be decided anymore.

tests/e2e/paper-navigation-button.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        apiClient.softDeleteProject({ id: projectId });
+    });
+
+    test(
+        "When the user is not in review mode and clicks the next paper button, " +
+            "then the next paper in the same project with the succeeding localId is shown.",
+        async ({ page, paperViewPage }) => {
+            await page.evaluate(() => {
+                localStorage.setItem("reviewMode", JSON.stringify(false));
+            });
```
mowi12: Take a look at the `reading-list-page-fixture.ts` file. It's using more than one page in the `use` part. This allows to use different pages inside a test.
You could use this to modify the review mode with a real click action instead of only setting the local storage.
What do you think?

mowi12 (changes requested):
Great work!
There are some parts that need a bit of work, but in general the logic checks out!

I have mentioned it in some of my conversations, but please look at all the code and add some spaces so the code looks more structured. You can even let AI do that for you, I also do that sometimes.

src/routes/project/[projectId]/paper/[paperId]/+page.svelte
Comment (domienderle) [reply src/routes/project/[projectId]/paper/[paperId]/+page.svelte:13]: Because the changes of data have to be translated through the pages. If this is no change, the `PaperView` component doesn't recognize any data changes if we only use `goto`

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
domienderle: Good catch.

src/lib/utils/paper-navigation.ts
domienderle: You are right. This should have been updated.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
domienderle: You are right.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
domienderle: If we wouldn't use the effect here, the page wouldn't update. Because we use `goto`, the page is not reloaded. Thus we have to use effect, to react on the changed data. I know it looks ugly, but I found no better way to do that.

tests/e2e/pom/paper-view-model.ts
domienderle: Yes, right. That is the reason.

tests/e2e/fixtures/paper-view-fixture.ts
domienderle: Yes we can do this. I don't know why I have removed it. It wasn't really on purpose.

tests/e2e/paper-navigation-button.test.ts
domienderle: I get your point, but the problem is, that I would have to navigate in between the different tests to this page. I cannot include this in the fixture, because the `reviewMode` shouldn't be changed before each test, but only before some of the tests.

tests/e2e/fixtures/paper-view-fixture.ts
domienderle: This line is important as long as I change the `reviewMode` the way I do. The reason for this is, that the page has to be loaded first to change the context of the page (I guess I wouldn't matter to which page we would navigate, so I chose `goto("/")`).

tests/e2e/decide-on-paper.test.ts
domienderle: I didn't add a sperate test case for it. But since I removed the page reload in the last test (which accepts a paper, after which no other paper to review is available), exactly this case that you mentioned is tested implicitly.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: True.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: You are right. This was to get the error messages easier during the e2e tests, but is now unnecessary.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: Good suggestions. I will add this.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: True.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: Good call. Thanks.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
domienderle: My implementation doesn't make sense. It is better to move this up.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
domienderle: Good call.

tests/integration/paper-components/paper-view/paper-navigation-button.test.ts
domienderle: I don't know. I didn't really think about it intentionally, but you are right. It is better structured with spaces.

tests/e2e/pom/paper-view-model.ts
domienderle: I also thought about that. The problem is, that you cannot get button then, if it is disabled, because the button text doesn't show `Previous Paper` or `Next Paper` but only the text on-hover. This text is not shown, when the button is disabled. This is problemtatic for the tests, so I have to. In some cases we have to do this. Felix also did this once.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
mowi12: Did you change something?

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
mowi12: You now only need to remove the `:` at the end of the toast message.

src/routes/project/[projectId]/paper/[paperId]/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/paper/[paperId]/+page.svelte:13]: Makes sense.

tests/e2e/pom/paper-view-model.ts
mowi12: Ahh, makes sense. Then it's okay.

tests/e2e/fixtures/paper-view-fixture.ts
mowi12: But then please wait for the SnowballR text to be visible (see homepage fixture).

tests/e2e/decide-on-paper.test.ts
mowi12: I see. As we discussed there is a new thing that came up.
Since we do not get redirected, the user needs to know why. A toast (no success, default should be best here), to notify, that there are no more papers to review.

Then we need a new test case to test this, or not?

tests/e2e/paper-navigation-button.test.ts
mowi12: As we discussed the usage of the reviewmode this way is not optimal, but in order to use this, you would need to create the whole test suite of the review mode, which is a lot.
On the other hand, the review mode component is only a switch and there is only 1 test case, two at the most.

If you decide to not do this in this issue please tell me, or add this to the other issue, where we update all e2e tests.

src/lib/utils/paper-navigation.ts
```diff
@@ -0,0 +1,32 @@
+import { goto } from "$app/navigation";
+import type { Project, Project_Paper } from "$lib/model/api/project";
+
+/**
+ * Handles the navigation of the navigation button depending on the direction.
+ * @param direction - the direction to navigate in ("left" or "right")
```
mowi12, line 7: Please add a new line between the description and the params.

mowi12 (changes requested):
We already discussed a lot, but there are still one or two things.
But all in all, well done!

Things we disussed:

- review mode tests in this issue, or new one
- toast at the end
- Review the last paper is not updated

tests/e2e/decide-on-paper.test.ts
domienderle: I think an info toast could be useful here. What do you think?

I don't think we need an extra test case here. I the check in one of the existing test cases, which should be enough imo.

tests/e2e/paper-navigation-button.test.ts
domienderle: I would propose to implement this in the extra test issue. I have added it to the description of the issue.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
mowi12: In my code, the `:` is still there. Also in the next functions.

tests/e2e/fixtures/paper-view-fixture.ts
mowi12: The code looks exactly like before.

tests/e2e/decide-on-paper.test.ts
mowi12: Yes, a info toast would be great! But there is no tests case for this. Which test includes this case in your oppinion?

mowi12 (changes requested):
I don't know what happend, but there are still some conversation where you reacted with a thumbs up, but no changes happend. Did you push all your changes?

I will take another long once you applied all the current requests.

Also rebase if you want, there have has been one small fix regarding the e2e tests.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: I don't know what happened. I was sure I have added them. But now it should be.

tests/e2e/fixtures/paper-view-fixture.ts
domienderle: Same here. Maybe I accidentally dropped some changes without recognizing it. Should be changed now.

tests/e2e/project/paper/decide-on-paper.test.ts
```diff
@@ -73,38 +73,47 @@ test.describe("Decide on paper", () => {
...
+            "then a review is submitted and the decision is shown (because there is no" +
+            "other paper to review).",
         async ({ page, paperViewPage }) => {
             await paperViewPage.openProjectPaperView(projectId, localProjectPaperIds[2]);
 
             await paperViewPage.decideOnPaper(ReviewDecision.ACCEPTED);
             await expect(page.getByText("Successfully submitted a review.")).toBeVisible();
-
-            await page.reload();
+            await expect(page.getByText("No mor papers to review.")).toBeVisible();
```
domienderle: I have added the test for the toast here. Imo opinion this should be enough.

domienderle:
I have changed the open problem. I am not sure why the changes weren't there, but should be now.

tests/e2e/project/paper/decide-on-paper.test.ts
mowi12: Okay. I didn't see that. You're right. This should suffice.

tests/e2e/project/paper/decide-on-paper.test.ts
```diff
@@ -73,38 +73,47 @@ test.describe("Decide on paper", () => {
...
+                }),
+            ).toBeVisible();
         },
     );
 
     test(
         "When the user selects certain review criteria and clicks a decision button, " +
-            "then a review is submitted and the decision is shown after a reload.",
+            "then a review is submitted and the decision is shown (because there is no" +
+            "other paper to review).",
```
mowi12, line 110: What has this test to do with review criteria? No review criteria is selected in this test. You check if some are selected or not, but why?
If you want to check if they are kept after a review is done, please add a comment above those checks so this is clear.
Maybe also rephrase the test description, since they are not selected. We want to see if they persist, if you know what I mean.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,164 @@
...
-        href="/papers/1"
-        onClick={() => console.log("button was clicked")}
-    />
+<PaperNavigationButton
+                direction="right"
+                {loadingProject}
+                {loadingProjectPaper}
+                bind:loading
+                bind:paperQueue
+                bind:nextProjectPaper
```
mowi12, line 148: Is `bind:previousProjectPaper` missing, or just not needed for an example?

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,47 +2,164 @@
...
+    /**
+     * Pre-fetches the next and the previous paper to find out if the according button is enabled or
+     * disabled and the according paper does not have to be loaded on clicking of the button, but
+     * before. Loads the next or previous paper only if all necessary information is available.
+     */
+    $effect(() => {
+        if (loading) return;
+        (async () => {
+            const paper = await loadingProjectPaper;
+            if (!paper) return;
```
mowi12, line 95: I know, I probably really annoy you, but please add a new line after this.
This makes it more clear that a new logical section is starting.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
```diff
@@ -108,11 +144,15 @@
...
+
+    $effect(() => {
+        (async () => {
+            const review = await getUserReviewIfAlreadySubmitted();
+            loadingUserReview = Promise.resolve(review);
+            selectedReviewCriteria.criteria = review?.selectedCriteriaIds ?? [];
+            wasAlreadyReviewedState.wasReviewed = review !== undefined;
+            return review;
+        })();
     });
```
mowi12: I don't quite when and why this `effect()` call is updating and on which vars it is depending on.

mowi12 (changes requested):
Thanks for reacting to all my conversations. Now they are done!
There are some small things left you need to take a look at.
I also noticed some other things we need to talk about.

Take a look at this video. When I click, the button is changing its size. We do not want that. It's probably from the resizing as soon as the loading circle appears, but this shouldn't resize the button:

https://github.com/user-attachments/assets/8d688982-8311-492b-88c6-792d86d1fc30

Some other not as important things:

While testing, I noticed that there is maybe something wrong with the implementation in the mock backend. If we have stage 0 and 1 and i review a paper in stage 0 which is not the last in this stage, i will get directed to the first unreviewed paper in stage 1 instead of the next in stage 0.
I don't think we need to fix this in the mock backend, but just to let you know.

But I also see, that you fixed the bug, where I was able to resubmit my review on the last paper. This seems to be working correctly now!

Another thing I think is wrong with the mock backend is the next button. paper with id 13 and 14 in the same stage are both unreviewed. Shouldn't I be able to click the next button and go the the paper with id 14?

tests/e2e/project/paper/decide-on-paper.test.ts
domienderle: I am not sure if get what you mean. By the way, this test was written by Merseleo, I only adjusted it. But I will try to answer your question. I you have a look into the `decideOnPaper` method, there gets an `exampleIncllusionCriteria` selected once we click the `ACCEPT` button. Thus, we do select the criteria. I hope this answers your question.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: It is not needed for the example.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: No worries you don't annoy me. I get, that my code is hardly structured sometimes. I am not sure if I got the right spot you meant.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
domienderle: You are right. This was a weird `effect` and just a work-around. But since this was one of the reasons that caused the resizing of the buttons, I had to refactor it and now it should be better and more intuitive. It is for updating the all the variables that are affected by a new `userReview`.

domienderle:
> Thanks for reacting to all my conversations. Now they are done! There are some small things left you need to take a look at. I also noticed some other things we need to talk about.
> 
> Take a look at this video. When I click, the button is changing its size. We do not want that. It's probably from the resizing as soon as the loading circle appears, but this shouldn't resize the button:
> 
>  Screen.Recording.2025-06-01.at.20.01.43.mov 
> Some other not as important things:
> 
> While testing, I noticed that there is maybe something wrong with the implementation in the mock backend. If we have stage 0 and 1 and i review a paper in stage 0 which is not the last in this stage, i will get directed to the first unreviewed paper in stage 1 instead of the next in stage 0. I don't think we need to fix this in the mock backend, but just to let you know.
> 
> But I also see, that you fixed the bug, where I was able to resubmit my review on the last paper. This seems to be working correctly now!
> 
> Another thing I think is wrong with the mock backend is the next button. paper with id 13 and 14 in the same stage are both unreviewed. Shouldn't I be able to click the next button and go the the paper with id 14?

Really good catch. I didn't see that. Should be fixed now. It wasn't just the resizing of the buttons, it was also, that the submit buttons disappeared onClick and appeared after the `loadingUserReview` was resolved. Should be done now.

tests/e2e/project/paper/decide-on-paper.test.ts
mowi12: I misunderstood the test. I talked with Merseleo about this and I agree, the test is correct.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
mowi12: This was the correct spot.

src/lib/components/composites/paper-components/paper-view/PaperView.svelte
mowi12: Perfect!

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -2,50 +2,170 @@
...
+        loadingProject,
+        loadingProjectPaper,
+        loading = $bindable(false),
+        paperQueue = $bindable([]),
+        nextProjectPaper = $bindable(undefined),
+        previousProjectPaper = $bindable(undefined),
+    }: Props = $props();
+    let buttonLeftDisabled: boolean = $state(true);
+    let buttonRightDisabled: boolean = $state(true);
+    const showLoadingSpinner = $state({ value: false });
```
mowi12, line 33: Why do you use an object with a boolean value? Why not simply a boolean directly, like the lines above.

mowi12 (changes requested):
One last question!

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: To be honest I am not sure. This is just what I have rebased from Felix issue with the loadingSpinner. Since this was approved like that, I didn't really question this.

mowi12: Ok, then we will keep it. Maybe it will be refactored in another refactoring round : )

mowi12 (approved):

PR #400, opened 2025-05-23
Remove whitespace in change name functionality

### LINKED ISSUE #391: [Bug]: Remove whitespace in change name functionality

#58

### What happened?

Currently the user name can start with a space. This should not happen.

### Steps to reproduce

Go to user settings -> change name
Enter a space with other characters
click rename
top left only shows 1 or no initials

### What did you expect?

Two initials

### Anything else?

_No response_

mowi12:
Closes #391 

## What I have made

- Fixed schema validation
- Updated all impacted components
- Updated all tests

## Checklist

  - [x] unit tests
  - [ ] ~~integration tests~~
  - [x] end-to-end tests

### DIFF
--- a/tests/unit/schemas.test.ts
+++ b/tests/unit/schemas.test.ts
@@ -1,6 +1,6 @@
 import { Schema, ZodIssueSubCode } from "$lib/schemas";
 import { assert, describe, expect, test } from "vitest";
-import { z } from "zod";
+import { z, ZodIssueCode } from "zod";
 
 /**
  * Wrapper for `describe` that provides a `testValid` and `testInvalid` function.
@@ -112,8 +112,12 @@ function customCode(subCode: ZodIssueSubCode): string {
 }
 
 schemaTest("First Name Schema", Schema.firstName, (testValid, testInvalid) => {
-    test.each(["", " ", "  "])("When first name is blank, then validation fails", (input) => {
-        testInvalid(input, { codes: [z.ZodIssueCode.too_small] });
+    test("When first name is empty, then validation fails", () => {
+        testInvalid("", { codes: [z.ZodIssueCode.too_small] });
+    });
+
+    test.each([" ", "  "])("When first name is blank, then validation fails", (input) => {
+        testInvalid(input, { codes: [customCode(ZodIssueSubCode.no_leading_trailing_whitespace)] });
     });
 
     test("When first name is more than 100 characters long, then validation fails", () => {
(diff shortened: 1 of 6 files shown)


tests/unit/schemas.test.ts
```diff
@@ -33,7 +33,11 @@ function schemaTest(
             result.success,
             `Input validation failed for '${input}': ${getErrorMessage(result.error)}`,
         ).toBe(true);
-        expect(result.data, "Validation data doesn't match expected output").toBe(expectedOutput);
+        if (result.success) {
+            expect(result.data, "Validation data doesn't match expected output").toBe(
+                expectedOutput,
+            );
+        }
```
Slartibartfass2: Why the if condition? Don't we check for `result.success` in the line above?

tests/unit/schemas.test.ts
```diff
@@ -112,10 +116,14 @@ function customCode(subCode: ZodIssueSubCode): string {
 }
 
 schemaTest("First Name Schema", Schema.firstName, (testValid, testInvalid) => {
-    test.each(["", " ", "  "])("When first name is blank, then validation fails", (input) => {
+    test.each([""])("When first name is empty, then validation fails", (input) => {
         testInvalid(input, { codes: [z.ZodIssueCode.too_small] });
     });
```
Slartibartfass2:
```suggestion
    test("When first name is empty, then validation fails", () => {
        testInvalid("", { codes: [z.ZodIssueCode.too_small] });
    });
```

tests/unit/schemas.test.ts
```diff
@@ -135,24 +143,30 @@ schemaTest("First Name Schema", Schema.firstName, (testValid, testInvalid) => {
...
+            });
         },
     );
 });
 
 schemaTest("Last Name Schema", Schema.lastName, (testValid, testInvalid) => {
-    test.each(["", " ", "  "])("When last name is blank, then validation fails", (input) => {
+    test.each([""])("When last name is empty, then validation fails", (input) => {
         testInvalid(input, { codes: [z.ZodIssueCode.too_small] });
     });
```
Slartibartfass2:
```suggestion
    test("When last name is empty, then validation fails", () => {
        testInvalid("", { codes: [z.ZodIssueCode.too_small] });
    });
```

tests/unit/schemas.test.ts
```diff
@@ -274,8 +290,12 @@ schemaTest("Password Schema", Schema.password, (testValid, testInvalid) => {
 });
 
 schemaTest("Project name Schema", Schema.projectName, (testValid, testInvalid) => {
-    test.each(["", " ", "  "])("When project name is blank, then validation fails", (input) => {
-        testInvalid(input, { codes: [z.ZodIssueCode.too_small] });
+    test.each([""])("When project name is empty, then validation fails", (input) => {
+        testInvalid(input, { codes: [ZodIssueCode.too_small] });
+    });
```
Slartibartfass2:
```suggestion
    test("When project name is empty, then validation fails", () => {
        testInvalid("", { codes: [ZodIssueCode.too_small] });
    });
```

Slartibartfass2 (changes requested):
Only these minor points. Good work.

tests/unit/schemas.test.ts
mowi12: Oh man ... I clearly did not pay attention writing those tests : )
I updated all the tests regarding this error.

tests/unit/schemas.test.ts
mowi12: At that moment, it made sense, but after reading it again I agree, it doesn't make a difference.

My thought process went something like this:
We only need to check if the result data and the expected output are a match if the validation itself is successful. If the validation is not successful, there is no need to check for the expected output since the test is already failed.

I will remove the part because it really is not needed.

mowi12 (commented):
I squashed the small changes into their according commits since the changes are very small, and I think you will find the appropriate positions in the code.

Slartibartfass2 (approved):
LGTM

Slartibartfass2 (approved):

PR #402, opened 2025-05-23
Added umlauts and sharp-s to valid password characters

### LINKED ISSUE #376: [Bug]: Faulty Implementation of Password Check

_No response_

### What happened?

While testing the authentication in the mock backend, I used the new sign up page. I wanted to use the following. credentials:

Image

Nothing happend.

Same details, other password ('SuperPasswort_00!') worked.

After removing the 'ß' from the password, everything worked.

The solution is probably to add ß to the list of valid character or to mention that Umlaute and 'ß' are not allowed.

### Steps to reproduce

See above.

### What did you expect?

The password should allow 'ß' as valid character or an added highlight, that 'ß' is not valid.

### Anything else?

_No response_

Slartibartfass2:
Closes #376 

## What I have made

- I added umlauts and sharp-s to the valid password characters

## Checklist

  - [x] unit tests
  - ~~[ ] integration tests~~ (not necessary)
  - ~~[ ] end-to-end tests~~ (not necessary)

### DIFF
--- a/src/lib/schemas.ts
+++ b/src/lib/schemas.ts
@@ -52,10 +52,10 @@ const lastNameSchema = z
  */
 const emailSchema = z.string().email({ message: "a valid format" });
 
-const upperCaseLetters = "A-Z";
-const lowerCaseLetters = "a-z";
+const upperCaseLetters = "A-ZÄÖÜ";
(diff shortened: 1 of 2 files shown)


mowi12 (approved):
Nice work. LGTM

PR #404, opened 2025-05-24
Replace checkbox with switch

### LINKED ISSUE #396: [Refactor]: Replace single checkboxes with switches

_No response_

In the case, there is only a singular checkbox for a setting, a switch should be used instead. Checkboxes should be used if there is more than 1 setting.

In the case, there is only a singular checkbox for a setting, a switch should be used instead.

_No response_

### Anything else?

_No response_

mowi12:
Closes #396 

## What I have made

- Replaced shortcuts visibility checkbox with switch
- Renamed review mode checkbox to witch (no other changes)

## Checklist

- [x] I have updated the tests


Slartibartfass2 (approved):
Delightful!

PR #406, opened 2025-05-25
Remove lighthouse runner

### LINKED ISSUE #403: [Bug]: Lighthouse job isn't signed in

_No response_

### What happened?

Now that we have the authentication done, the lighthouse job isn't automatically signed in and always audits the sign in page because it is redirected to it.

### Steps to reproduce

1. Run the lighthouse tool `npm run lighthouse:all`
2. See that all routes audited the sign in page

### What did you expect?

The audits should be from their according pages.

### Anything else?

_No response_

Slartibartfass2:
Closes #403

## What I have made

This was open for far too long. With the migration from the mock backend to the real backend in mind, I don't see the reason for keeping the runner anymore. Accessing the real backend with a signed-up user will be way more effort than the benefits we get from this runner. We didn't use it frequently in the past year, so why keep using it? We still can audit the application using Chrome if we'd like to. If somebody wants to add this back in in the future, they can do so, but I don't want to keep this open for any longer.

## Checklist


Merseleo:
Do we need this?

Slartibartfass2:
> Do we need this?

Yes, the lighthouse job only runs on the login screen currently because it cannot access the pages.

mowi12 (approved):

PR #407, opened 2025-05-26
Added reusable loading button

### LINKED ISSUE #375: [Feature]: Add reusable loading button

_No response_

In some cases, we have buttons that make a request and while loading are in a disabled state. There were some inconsistencies when it comes to how such a button should look like.

Create a reusable component that wraps a button and controls its loading state.

We decided on the following properties:
- button is disabled while the request is executed
- the label should switch, e.g., "Invite Users" becomes "Inviting Users"
- a loading circle should be shown as an indicator, i.e., the `LoadingCircle` svg
- the button should have a fixed size so that the width doesn't change in the loading state (this may need to be configured from the outside)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #397: [Feature]: Replace all buttons with loading buttons

_No response_

While the frontend is waiting for a response from the backend, a clicked button should display a loading state.

Replace all button with loading buttons from shadcn.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #375 
Closes #397 

## What I have made

Review this PR commit-by-commit.

- First I added the `LoadingButtonComponent`, which handles the loading state of a button
- Then I refactored existing buttons to use this functionality
- As `AlertDialog`s have a separate action button, we have to customize this ourselves instead of using the `LoadingButton` component
- Same goes for the `Tooltip` component

## Checklist

  - ~~[ ] unit tests~~ (not required)
  - [x] integration tests (only minor adjustments)
  - [x] end-to-end tests (only minor adjustments)

### DIFF
--- a/src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
@@ -8,10 +8,10 @@
     import { shortcuts } from "$lib/global-state/shortcuts-visibility-state.svelte";
     import { backendService } from "$lib/grpc-api";
     import { type Review, type Review_Create, ReviewDecision } from "$lib/model/api/review";
-    import { LoaderCircle } from "lucide-svelte";
     import { getSelectedReviewCriteriaContext } from "$lib/utils/custom-context";
     import { toast } from "svelte-sonner";
     import { shortcut, type ShortcutTrigger, type ShortcutEventDetail } from "@svelte-put/shortcut";
+    import { loadingWrapper } from "$lib/utils/common-helper";
 
     interface ButtonContent {
         name: string;
@@ -22,19 +22,27 @@
     interface PaperDecisionButtonProps {
         projectPaperId: string;
         variant: PaperDecisionButtonVariant;
-        isSubmittingReview?: boolean;
+        isSubmittingReview?: { value: boolean };
         userReview?: Review;
     }
 
     let {
         projectPaperId,
         variant,
-        isSubmittingReview = $bindable(false),
+        isSubmittingReview = $bindable({ value: false }),
         userReview,
     }: PaperDecisionButtonProps = $props();
 
     const wasAlreadyReviewed = userReview !== undefined;
-    let showLoadingSpinner = $state(false);
+    const showLoadingSpinner = $state({ value: false });
+
+    $effect(() => {
+        // Update `isSubmittingReview` every time `showLoadingSpinner` is updated
+        // `isSubmittingReview` is used as a state in the `PaperView` component to disable all decision buttons while
+        // the API call is made. However we need a different sate for displaying the loading state of the single
+        // decision button. Otherwise, all would have a spinner and label with "Submitting Review".
+        isSubmittingReview.value = showLoadingSpinner.value;
+    });
 
     /**
      * Returns the content of the button, i.e. the button name, the shortcut and the tooltip text
@@ -99,7 +107,6 @@
      * and a loading spinner is shown. After the review was submitted, a confirmation toast is shown.
      */
     async function submitReview() {
-        isSubmittingReview = showLoadingSpinner = true;
         try {
             const review: Review_Create = {
                 projectPaperId: projectPaperId,
@@ -117,7 +124,6 @@
             });
             console.error("Could not submit the review:", err);
         }
-        isSubmittingReview = showLoadingSpinner = false;
     }
 
     const selectedReviewCriteriaState = getSelectedReviewCriteriaContext();
@@ -132,9 +138,9 @@
                 const keyboardEvent = detail.originalEvent;
                 keyboardEvent.preventDefault();
 
-                submitReview();
+                loadingWrapper(showLoadingSpinner, submitReview, {});
             },
-            enabled: !wasAlreadyReviewed,
+            enabled: !wasAlreadyReviewed && !showLoadingSpinner.value,
         },
     }}
 />
@@ -174,21 +180,20 @@ Usage:
         paperDecisionButtonVariants({ variant }),
     )}
     data-testid={`decision-button-${variant}`}
-    disabled={isSubmittingReview || wasAlreadyReviewed}
-    onclick={submitReview}
+    disabled={isSubmittingReview.value || wasAlreadyReviewed || showLoadingSpinner.value}
+    loading={showLoadingSpinner.value}
+    onclick={(args) => loadingWrapper(showLoadingSpinner, submitReview, args)}
     triggerSize="default"
     triggerVariant="default"
 >
     {#snippet trigger()}
-        {#if showLoadingSpinner}
-            <LoaderCircle class="animate-spin" />
-            Submitting review
-        {:else}
-            <p>{getButtonContent().name}</p>
-            {#if shortcuts.isVisible}
-                <p>{getButtonContent().shortcut}</p>
-            {/if}
+        <p>{getButtonContent().name}</p>
+        {#if shortcuts.isVisible}
+            <p>{getButtonContent().shortcut}</p>
         {/if}
+    {/snippet}
+    {#snippet loadingTrigger()}
+        Submitting review
     {/snippet}
     {#snippet content()}
         {getButtonContent().tooltipText}
--- a/src/lib/components/composites/project-components/ProjectListEntry.svelte
+++ b/src/lib/components/composites/project-components/ProjectListEntry.svelte
@@ -2,12 +2,24 @@
     import { Progress } from "$lib/components/primitives/progress";
     import { getNames, handleSingleOrDoubleClick } from "$lib/utils/common-helper";
     import {
+        MemberRole,
         Project,
         Project_Information,
         Project_Member_List,
         ProjectStatus,
     } from "$lib/model/api/project";
     import { goto } from "$app/navigation";
+    import { env } from "$env/dynamic/public";
+    import { UserContextKey } from "$lib/global-context/userContext";
+    import { getContext } from "svelte";
+    import type { User } from "$lib/model/api/user";
+    import { Badge } from "$lib/components/primitives/badge/index.js";
+
+    const isDevMode = env?.PUBLIC_IS_DEV_MODE === "true";
+    let user: User | undefined;
+    if (isDevMode) {
+        user = getContext<() => User>(UserContextKey)();
+    }
 
     export interface ProjectListEntryInterface {
         project: Project;
@@ -52,7 +64,15 @@ Usage:
     {...!onClick ? { href: href } : { type: "button" }}
 >
     <div class="flex h-fit min-w-0 flex-col">
-        <h2 class="truncate">{project.name}</h2>
+        <h2 class="truncate">
+            {project.name}
+            <!-- Show admin badge in dev mode when current user is admin of project -->
+            {#if isDevMode && membersList.members.some((member) => member.user?.id === user?.id && member.role === MemberRole.ADMIN)}
+                <Badge class="ml-2" title="You are an admin of this project" variant="outline">
+                    admin
+                </Badge>
+            {/if}
+        </h2>
 
         {#if membersList.members.length > 0}
             <span class="text-hint truncate">
--- a/src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
+++ b/src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
     import SettingsSection from "$lib/components/composites/settings/SettingsSection.svelte";
     import Input from "$lib/components/composites/input/Input.svelte";
-    import { Button } from "$lib/components/primitives/button";
     import { backendService } from "$lib/grpc-api";
     import { onMount } from "svelte";
     import { Project } from "$lib/model/api/project";
@@ -9,9 +8,10 @@
     import { generateFieldMask } from "protobuf-fieldmask";
     import type { ApiError } from "$lib/model/general";
     import { invalidate } from "$app/navigation";
-    import { LoaderCircle } from "lucide-svelte";
     import { toast } from "svelte-sonner";
-    import Alert from "$lib/components/composites/utils/Alert.svelte";
+    import Alert, { type AlertVariant } from "$lib/components/composites/utils/Alert.svelte";
+    import LoadingButton from "$lib/components/composites/button/LoadingButton.svelte";
+    import { loadingWrapper } from "$lib/utils/common-helper";
 
     interface Props {
         projectId: string;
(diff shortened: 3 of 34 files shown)


src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -33,26 +36,28 @@
             lastName: lastNameInput.getValue(),
         };
 
-        const updatedUser = User.create(userData);
-        const maskPaths = generateFieldMask(userData).filter((path) => path !== "id");
+        if (userData.firstName === user.firstName && userData.lastName === user.lastName) {
+            toast.warning("Please provide a new name");
+            return;
+        }
```
mowi12: This toast doesn't really fit the other error messages. If the input is invalid, a red box appears below the input. If the input is the current name, a toast appears?
I don't think this is a great solution. Maybe also use the updateUserError variable for this.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -33,26 +36,28 @@
...
                 toast.success("Successfully updated your name.");
 
-                // Make sure the input fields are not focused after submitting
+                // Make sure that the input fields are not focused after submitting
                 (document.activeElement as HTMLElement)?.blur();
             })
             .catch((error) => {
                 updateUserError = { errorTitle: "Something went wrong while updating user." };
-                console.error(error);
+                console.error(`Couldn't update user name: ${error}`);
```
mowi12, line 67: Just a question because I'm curious. There is no difference between
```
`Couln't update user name: ${error}`
```
and
```
"Couldn't update user name:", error
```

Or is there?

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -98,7 +105,13 @@ Usage:
             type="text"
             value={user.lastName}
         />
-        <Button class="text-md w-full md:mt-5.5 md:w-25" type="submit">Rename</Button>
+        <LoadingButton
+            class="text-md w-full md:mt-5.5 md:w-80 lg:w-72 xl:w-68"
```
mowi12, line 116: Do we really need the `lg` and `xl` variants? I don't really see any difference.

tests/integration/settings/user-settings/change-name-settings.test.ts
```diff
@@ -67,6 +68,17 @@ describe("ChangeNameSettings", () => {
         expect(mockCall).toHaveBeenCalled();
     });
 
+    test("When the button is clicked with the same name, then the api is not called", async () => {
```
mowi12, line 71: If you agree to not use the toast in the change name component, maybe check if the error message is visible.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -53,39 +54,42 @@
...
+
+        const projectData: Partial<Project> = {
+            id: projectId,
+            name: projectNameInput.getValue(),
+        };
+
+        if (projectName === projectData.name) {
+            toast.warning("Please enter a new project name.");
+            return;
         }
```
mowi12: Same opinion as in the change name component for the user. I think a constant error message is better than a toast.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -102,33 +106,28 @@ Usage:
...
     >
         <Input
             bind:this={projectNameInput}
             class="h-full w-full"
-            disabled={loading}
+            disabled={loadingProjectName || loading.value}
             inputId="project-name-input"
             label="Project Name"
-            placeholder={loading ? "Loading" : "New Project Name"}
+            placeholder={loadingProjectName ? "Loading" : "New Project Name"}
```
mowi12, line 123: Could you test if this works?
With the current issue, that we wait for the user context, I never see "Loading". When the component has loaded, I directly see the correct name of the project.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
mowi12, line 1: Is there a reason you updated and even added new test cases for the use change name component, but not this component. I explicitly mean the e2e tests, since there are only 3 simple tests and none test the error messages or that no api call has been made.

tests/integration/dialog/TestAlertDialog.svelte
```diff
@@ -7,20 +7,16 @@
         open?: boolean;
     }
 
-    let {
-        loading = $bindable(false),
-        error = $bindable(undefined),
-        open = $bindable(false),
-    }: Props = $props();
+    let { loading = false, error = $bindable(undefined), open = false }: Props = $props();
```
mowi12: Didn't you remove the ability to bind error?

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -22,19 +22,23 @@
...
+        isSubmittingReview = $bindable({ value: false }),
         userReview,
     }: PaperDecisionButtonProps = $props();
 
     const wasAlreadyReviewed = userReview !== undefined;
-    let showLoadingSpinner = $state(false);
+    const showLoadingSpinner = $state({ value: false });
+
+    $effect(() => {
+        isSubmittingReview.value = showLoadingSpinner.value;
```
mowi12, line 44: I don't get why we need both states if they get set to the same value and are almost always used together. In which case is this actually making any difference?

src/lib/components/composites/project-components/ProjectListEntry.svelte
mowi12, line 1: I know this is only for development, but what do you think about using [badge](https://next.shadcn-svelte.com/docs/components/badge) component from shadcn. If you think this is too much, you can ignore this, but Imo this component is exactly what we want.

mowi12 (changes requested):
First of all, I really like the changes you've made! You always improve the code in every place you take some action. I like that.
There are only some small changes and questions you need to answer.

But the main thing: Is there a reason you updated added tests for the first one or two commits, and then stopped or only updated the integration tests?

I didn't look at all the tests, but do they all test the partially new behaviour, or are they really that complete?

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
Slartibartfass2: I see why this is confusing. I added a comment that should clarify the circumstances. Do you miss anything?

src/lib/components/composites/project-components/ProjectListEntry.svelte
Slartibartfass2: Great idea. I replaced it with the badge. I also removed the `flex items-center` part, as this disables the truncation.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
Slartibartfass2: No, there isn't.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
Slartibartfass2: Not necessarily as `md:w-80` is enough. However, I noticed that the width required somehow changes when the viewport widens. I always set the width to a value, which is just enough to display the loading label.

tests/integration/dialog/TestAlertDialog.svelte
Slartibartfass2: Great catch! That's obviously outdated.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
Slartibartfass2: I see why this feels weird and agree that a toast is a bad choice in this case. However, I don't think the error alert is a good choice too as this isn't really an error, but only a hint and in my opinion, the error message looks too much like a tragic error happened:
image

I searched for some examples:
image
image

In some cases, the action button is disabled as long as the input is not different.

I made some proposals, what are your thoughts?:
1. <br /> image
2. <br /> image
For the second case, we should provide a hint that this is because the input is the same.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: They do, aren't they? There are integration and e2e tests that check if the same project name is entered that an error message is shown/the API is not called.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: I will attend to this as soon as we agree on a solution.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: I added a test for this.

tests/integration/settings/user-settings/change-name-settings.test.ts
Slartibartfass2: I will.

Slartibartfass2 (commented):
Thank you for your review and the kind words.

To your question: I only added tests for the "same input" use case for the project name and user name settings. The other ones, don't require these tests.

I think we can discuss whether we should test die loading behavior for all affected components.

Edit: I will squash the new commits as soon as we're finished.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: I agree, the error alert is to much as this is no error. I generalized the error alert we are currently using so we are able to use all the variants with the same colors as the toasts. This would allow you to display a warning alert which would fit perfectly in this case.

I will remove the one commit from the current branch and make a separate one so you can cherry pick it or review it so you can use it in this pr.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
mowi12: Okay, I see you have thought about these values. If they make sense and have actual meaning, then they should stay.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
mowi12: Now, this makes more sense to me, thanks!

src/lib/components/composites/project-components/ProjectListEntry.svelte
mowi12: Perfect!

mowi12 (changes requested):
Perfect! There is only one discussion left, the error alert.

Regarding your comment about the testing about all the loading states.
Imo you don't need to test each component if the loading state is working, since the loading button is tested itself and you implemented them all the same way, they should work and those tests are only bloating the current code.

Slartibartfass2 (commented):
I now used the alerts for displaying the warning. What are your thoughts?

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
```diff
@@ -53,39 +55,46 @@
...
+        if (!projectNameInput.validate()) return;
+
+        const projectData: Partial<Project> = {
+            id: projectId,
+            name: projectNameInput.getValue(),
+        };
+
+        if (projectName === projectData.name) {
+            updateProjectError = {
+                errorTitle: "Please provide a different name.",
```
mowi12: What do you think about (based on this: [How to write error messages that actually help users rather than frustrate them](https://piccalil.li/blog/how-to-write-error-messages-that-actually-help-users-rather-than-frustrate-them/)):

```suggestion
                errorTitle: "Please provide a different name than the current name.",
```
Or something like this. Just different name doesn't tell the user why.

src/lib/components/composites/settings/user-settings/ChangeNameSettings.svelte
```diff
@@ -33,26 +36,34 @@
             lastName: lastNameInput.getValue(),
         };
 
-        const updatedUser = User.create(userData);
-        const maskPaths = generateFieldMask(userData).filter((path) => path !== "id");
+        if (userData.firstName === user.firstName && userData.lastName === user.lastName) {
+            updateUserError = {
+                errorTitle: "Please provide a different name.",
```
mowi12: Same as in project name : )

tests/e2e/project/settings/general/project-name-settings.test.ts
```diff
@@ -26,7 +26,7 @@ test.describe("Renaming a project", () => {
         await projectSettingsPage.checkForErrors();
         const projectNameHeader = page.getByRole("heading", { name: "Project 1" });
         await expect(projectNameHeader).toBeVisible();
-        const toast = page.getByText("Please enter a new project name.");
+        const toast = page.getByText("Please provide a different name.");
```
mowi12: Is this correct?
I thought there is an alert?

mowi12 (changes requested):
Almost done. Only one small thing about the phrasing of the warning.
And there are one or two unanswered conversations.

tests/e2e/project/settings/general/project-name-settings.test.ts
mowi12: Nvm, wrong test.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
Slartibartfass2: I've updated it. Tell me what you think.

tests/integration/settings/user-settings/change-name-settings.test.ts
```diff
@@ -67,6 +68,22 @@ describe("ChangeNameSettings", () => {
...
+
+        const renameButton = screen.getByRole("button", { name: "Rename" });
+
+        await userEvent.click(renameButton);
+        expect(mockCall).not.toHaveBeenCalled();
+        expect(
+            screen.getByText(
+                "To successfully change your name, you must provide a new one that is different from your current one.",
+            ),
+        ).toBeInTheDocument();
```
mowi12:
```suggestion
        expect(
            screen.getByRole("heading", {
                name: "To successfully change your name, you must provide a new one that is different from your current one.",
            }),
        ).toBeInTheDocument();
```

Why not like this? The test below this one is testing for the error alert in this way.

src/lib/components/composites/settings/project-settings/general/ProjectNameSettings.svelte
mowi12: Perfect, this is better!

mowi12 (changes requested):
Only one last small thing, then we are good to go!

tests/integration/settings/user-settings/change-name-settings.test.ts
Slartibartfass2: I changed it to use the alert role for both tests.

mowi12 (approved):
Nice. Another chuck done!

PR #413, opened 2025-05-27
feat: refactor the folder structure of the e2e tests

### LINKED ISSUE #389: [Feature]: Refactor the e2e into a folder structure
### Is your feature request related to a problem? 

The structure of the e2e tests gets worse the more test are added.

Create a folder structure for the e2e tests. The suggestion here is to structures the folders similar to the svelte routing. Meaning, that each test case takes place on one routed site. The site, where the test takes place determines in which folder ends up. For the folder structure have a look on the PR description.

domienderle:
Closes #389 

## What I have made

I have created a folder structure for the e2e test. The folder structure is based on the svelte routing (each test case ends up in the folder that includes all the test cases of the given route).

The folder structure looks like this: 

```
.
├── archivedprojects
├── auth/
│   ├── newpassword
│   ├── resetpassword
│   ├── signin
│   ├── signout
│   └── signup
├── homepage
├── paper
├── project/
│   ├── dashboard
│   ├── paper
│   ├── papers
│   └── settings/
│       ├── general
│       ├── members
│       ├── review
│       ├── slr
│       └── statistics
├── readinglist
├── settings
└── utils
```

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ only refactoring
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ only refactoring
  - ~~[ ] unit tests~~ only refactoring
  - ~~[ ] integration tests~~ only refactoring
  - ~~[ ] end-to-end tests~~ only refactoring


mowi12 (changes requested):
In general, I really like this structure!
But imo, the pom/fixture/tests structure inside the last level is too much.

To put one or sometimes 2 tests inside a tests folder is not necessary, because the files already end with .tests.ts. In VSC they are even highlited differently.

Same for the pom and fixture files. The fixtures all end with -fixture and the poms all with -model.
And since there will be only one fixture and model for each path, I don't think that an additional folder is necessary.

What do you think?

mowi12 (changes requested):
This structure looks good!
But I think you missed a spot. In the settings directory, the account and the shortcuts still have the subfolders. Please remove them.

And could you note which e2e tests need to be included in the other issue, where we want to add the missing tests?

domienderle:
Good catch, Didn't see that the settings folder was missing.

mowi12 (approved):
Perfect! Thanks for the fast replies!

PR #419, opened 2025-05-27
Add option for 'NamedList' component to group items

### LINKED ISSUE #394: [Feature]: Group papers in lists by projects

_No response_

All papers in lists (except in the papers tab inside a project) are not sorted. Since one paper with a global paper ID of 4 can be in multiple projects, this paper can be listed multiple times inside a list, sometimes not looking correctly because the user cannot differentiate between the papers (in which project they are).

Group papers by project. Maybe even insert spacers or anything that separates papers from different projects.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #394 

## What I have made

- I added two options to the `NamedList` component that allow setting a grouping function and group labels
   - If set, the items in the named list are grouped using the `groupBy` method and a group header (followed by a separator) is added before each group

## Checklist

  - ~~[ ] unit tests~~
  - [x] integration tests
  - ~~[ ] end-to-end tests~~

### DIFF
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -3,6 +3,7 @@
     import ErrorIndicator from "../utils/ErrorIndicator.svelte";
     import { groupBy } from "$lib/utils/common-helper";
     import { Separator } from "$lib/components/primitives/separator";
+    import { cn } from "$lib/utils/shadcn-helper";
 
     type T = $$Generic; /* eslint-disable-line no-undef */
 
@@ -139,13 +140,15 @@ is not found in the map, the group will be labeled as "Unknown".
             <span class="text-hint italic">{emptyHint}</span>
         {:else}
             <ul class="scroll-box space-y-4 pb-1">
-                {#if groupSelector !== undefined}
-                    {#each Object.entries(groupBy(loadedItems, groupSelector)) as [key, values] (key)}
-                        <div class="space-y-1">
+                {#if groupSelector}
+                    {#each Object.entries(groupBy(loadedItems, groupSelector)) as [key, values], index (index)}
+                        <!-- show group header before each group and add a gap to the previous group
+                             except for first group header that has no previous group -->
+                        <div class={cn("space-y-1", index >= 1 ? "mt-6" : "")}>
                             {#await groupLabels}
-                                <h3 class="italic">Loading</h3>
+                                <h2 class="italic">Loading</h2>
                             {:then loadedGroupLabels}
-                                <h3>{loadedGroupLabels?.[key] ?? "Unknown"}</h3>
+                                <h2>{loadedGroupLabels?.[key] ?? "Unknown"}</h2>
                             {/await}
                             <Separator />
                         </div>
--- a/src/lib/components/composites/list/NamedList.svelte
+++ b/src/lib/components/composites/list/NamedList.svelte
@@ -1,10 +1,12 @@
 <script lang="ts">
     import type { Snippet } from "svelte";
     import ErrorIndicator from "../utils/ErrorIndicator.svelte";
+    import { groupBy } from "$lib/utils/common-helper";
+    import { Separator } from "$lib/components/primitives/separator";
 
     type T = $$Generic; /* eslint-disable-line no-undef */
 
-    interface NamedListProps {
+    interface BaseProps<T> {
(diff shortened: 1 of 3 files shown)


src/lib/components/composites/list/NamedList.svelte
```diff
@@ -97,11 +139,21 @@ This can be e.g. a search bar.
...
-                {#each loadedItems as item (keySelector(item))}
-                    <li>
-                        {@render listItemComponent?.(item)}
-                    </li>
-                {/each}
+                {#if groupSelector !== undefined}
+                    {#each Object.entries(groupBy(loadedItems, groupSelector)) as [key, values] (key)}
+                        <div class="space-y-1">
+                            {#await groupLabels}
+                                <h3 class="italic">Loading</h3>
```
domienderle:
```suggestion
                                <h1 class="italic">Loading...</h1>
```
What do you think about this?

src/lib/components/composites/list/NamedList.svelte
```diff
@@ -97,11 +139,21 @@ This can be e.g. a search bar.
...
         {:else}
             <ul class="scroll-box space-y-4 pb-1">
-                {#each loadedItems as item (keySelector(item))}
-                    <li>
-                        {@render listItemComponent?.(item)}
-                    </li>
-                {/each}
+                {#if groupSelector !== undefined}
+                    {#each Object.entries(groupBy(loadedItems, groupSelector)) as [key, values] (key)}
+                        <div class="space-y-1">
```
domienderle:
```suggestion
                        <div class="space-y-1 mt-10">
```
What do you think about this?

src/lib/components/composites/list/NamedList.svelte
```diff
@@ -97,11 +139,21 @@ This can be e.g. a search bar.
...
+                            {#await groupLabels}
+                                <h3 class="italic">Loading</h3>
+                            {:then loadedGroupLabels}
+                                <h3>{loadedGroupLabels?.[key] ?? "Unknown"}</h3>
```
domienderle:
```suggestion
                                <h1>{loadedGroupLabels?.[key] ?? "Unknown"}</h1>
```

domienderle (changes requested):

src/lib/components/composites/list/NamedList.svelte
Merseleo: In my opinion the group header should not have a high-level heading the header of the entire list. So semantically `h1` is wrong imo. Additionally, it seems too big in my opinion. But if you want me to increase the size of the group headers and can change the text size style separately.

src/lib/components/composites/list/NamedList.svelte
Merseleo: I thought we agreed on not using "..." when loading something? But if you want this for clarity I can add them.

src/lib/components/composites/list/NamedList.svelte
Merseleo: Good idea, but in my opinion 40px is bit too much. What do you think of the new solution?

src/lib/components/composites/list/NamedList.svelte
Merseleo: But as I tested some different header sizes, `h2` seems to look the best and is more correct, as the `PaperListEntry` or other components use `h2` as well and this text should not be bigger than the heading as well. Are you ok with that?

Merseleo (commented):
Thank you for the feedback. I played around with some different designs and I think based on your feedback I found a good solution. What do you think?

src/lib/components/composites/list/NamedList.svelte
domienderle: That is true. Since the loading is italic it should be enough then.

src/lib/components/composites/list/NamedList.svelte
domienderle: This is also fine. I guess that's a matter of taste in the end. But looks good as it is now.

src/lib/components/composites/list/NamedList.svelte
domienderle: Yeah, that's a good point. Since it looks good as it is now, I think we are good here.

domienderle (approved):
Yes I am fine with that. Looks good now. Thank you.

PR #421, opened 2025-05-28
Use sharding for e2e tests to reduce testing duration

### LINKED ISSUE #418: [CI]: Don't execute e2e fail fast when no changes were made in the tests/e2e directory

_No response_

Currently, when no changes to the e2e tests were made, the fail fast job is still executed by building the server and then terminating because no changed tests have been found. This costs us approx 40s per run.

As the e2e tests are our bottleneck when it comes to finishing all status checks, we could reduce the long run time by skipping those fail fast tests when no changes have been made.

AFAIK, there's no native way to do that, but there is a GitHub action which can be used for such use case.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #420: [CI]: Use sharding for E2E Tests to reduce testing duration

_No response_

Currently, running the E2E tests take a lot of time to finish. By using sharding, we could reduce this time as the tests run in parallel.

Add sharding.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #420 
Closes #418 

## What I have made

To reduce the time, it takes to run the e2e tests, I introduced sharding, i.e., the tests run on several machines, and they only run a certain share of all tests.

Currently, the number of shards is 3. This means each runner executes 56 tests. While the previous job ran for 7-9 minutes, with this change it takes 4-5 minutes. We could even increase the number of shards, but as the setup also takes some time, at a certain point we don't get any improvements.

I previously had the idea that the fast failing tests should be executed as separate jobs before the not-fast-failing tests, but as the setup takes up to 3 minutes or sometimes more, I don't see that this is more efficient.

Additionally, we check if there were changes to the e2e tests directory, and if so, skip the fast failing tests. Previously, if there were no changes, the fast failing job still built the server and then didn't execute any tests. This time is now also reduced.

This should make the testing less cumbersome for use.

## Checklist

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -1,0 +1,47 @@
+name: End-to-End Tests
+
+on:
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+    workflow_dispatch:
+
+env:
+    node_version: ${{ vars.NODE_VERSION }}
+    PUBLIC_API_BASE_URL: ${{ vars.PUBLIC_API_BASE_URL }}
+
+jobs:
+    e2e-test:
+        name: End-to-End Testing with Playwright
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v4
+              with:
+                  # Force a non-shallow checkout, so that we can reference $GITHUB_BASE_REF.
+                  # See https://github.com/actions/checkout for more details.
+                  fetch-depth: 0
+                  submodules: "recursive"
+
+            - name: E2E-Setup
+              uses: ./.github/workflows/e2e-setup/
+              with:
+                  node_version: ${{ env.node_version }}
+
+            - name: Run only tests for changed code (fail fast)
+              if: github.event_name == 'pull_request'
+              run: npm run test:e2e -- --only-changed=origin/${{ github.base_ref }}
+
+            - name: Run Playwright
+              run: npm run test:e2e
+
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+                      e2e-tests:
+                        - 'tests/e2e/**'
+
+            - name: Run only tests for changed code (fail fast)
+              # Skip fast failing test run when not triggered by a pull request and no changes were made to the e2e test directory
+              if: ${{ github.event_name == 'pull_request' && steps.changes.outputs.e2e-tests == 'true' }}
+              run: npm run test:e2e -- --shard ${{ matrix.shard }}/${{ strategy.jobs-total }} --only-changed=origin/${{ github.base_ref }}
+
+            - name: Run Playwright
+              run: npm run test:e2e -- --shard ${{ matrix.shard }}/${{ strategy.job-total }}
```
mowi12: At the top you are using ...`jobs-total`... and at the bottom you are using ...`job-total`...
Is this correct or do you have a type in one of the versions?

mowi12 (changes requested):
This looks good. You refactored the old code and modularized it, as well as adding the sharding functionality successfully.
There is only one possible type you need to fix, but that's it!

.github/workflows/e2e_tests.yml
Slartibartfass2: Great catch! `jobs-total` is wrong. This would've failed. Thank you!

mowi12 (approved):
Perfect, this looks great!

PR #422, opened 2025-05-28
Toggle whether maybe is available decision

### LINKED ISSUE #26: [Use Case]: Toggle Whether Maybe Is Available Decision

Project Admin

### Describe the use case

As project admin, I want to toggle whether "maybe" is an available decision on a paper.

### Acceptance criteria

- when this option is toggled on, a project member can set their decision on a paper to "maybe"
- when this option is toggled off, a project member can't set their decision on a paper to "maybe"
- the project admin is warned that they are about to change this option when the project is past stage zero, as this might affect the outcome of the SLR
- the project admin has to confirm their decision to change this option (e.g. press confirm button)

### Anything else?

Part of #15

mowi12:
Closes #26 

## What I have made

- Refactored error alert to a customizable alert
  - updated components
  - updated tests
- Added maybe as decision section to project slr setting
  - added tests

## Checklist

  - [ ] ~~unit tests~~
  - [x] integration tests
  - [x] end-to-end tests


src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
```diff
@@ -0,0 +1,179 @@
...
+    let updateSLRSettingsError: ApiError | undefined = $state(undefined);
+
+    async function toggleIsMaybeAsDecisionSettingStatus(targetCheckedState: boolean) {
+        isUpdatingMaybeAsDecisionSettingStatus = true;
+
+        await loadingProject
+            .then((project) => {
+                const projectSettings = project.settings ?? Project_Settings.create();
+                projectSettings.reviewMaybeAllowed = targetCheckedState;
+                project.settings = projectSettings;
```
domienderle:
```suggestion
                project.settings ??= Project_Settings.create();
                project.settings.reviewMaybeAllowed = targetCheckedState;
```
What dou you think about this. It's a bit more compact.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
```diff
@@ -0,0 +1,179 @@
...
+    let isConfirmDialogOpen = $state(false);
+    let title = $state("");
+    let dialogDescription = $state("");
+    let pendingActionConfirmCallback: (() => void) | null = $state(null);
+
+    const disabled = $derived(slrSettingsLocked || isUpdatingMaybeAsDecisionSettingStatus);
+
+    let updateSLRSettingsError: ApiError | undefined = $state(undefined);
+
+    async function toggleIsMaybeAsDecisionSettingStatus(targetCheckedState: boolean) {
```
domienderle, line 42: Please add a short comment on the function. I know it is obvious what the handles, but for longer methods I would prefer a short comment.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
```diff
@@ -0,0 +1,179 @@
...
+            pendingActionConfirmCallback();
+        }
+        pendingActionConfirmCallback = null;
+    }
+
+    function handleCancelClick() {
+        pendingActionConfirmCallback = null;
+    }
+
+    onMount(async () => {
```
domienderle, line 124: Same here. A short comment would be great.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
```diff
@@ -0,0 +1,179 @@
...
+
+<!--
+@component
+This component renders a section within the SLR project settings. It allows administrators to toggle this setting for the project.
+When enabled, reviewers can gain the ability to select 'Maybe' as a decision on a paper, in addition to the standard options. If disabled, the 'Maybe' option will be removed from the decision options.
+Before the setting is changed, a confirmation dialog is presented to the admin, requiring them to confirm their choice to either enable or disable the 'Maybe as Decision' functionality.
+
+Usage:
+```svelte
+  <MaybeAsDecisionSetting {projectId} {loadingProject} />
```
domienderle:
```suggestion
  <MaybeAsDecisionSetting {projectId} {loadingProject} {slrSettingsLocked}/>
```
Update the usage with the slr settings

src/lib/global-state/maybe-as-decision-state.svelte.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const STORAGE_KEY = "maybeAsDecision";
+const initialValue = getLocalStorageItem<boolean>(STORAGE_KEY, true);
+
+/**
+ * Stores, whether 'Maybe' is allowed as decision as review or not.
+ * This is only used to perfect the visualization of the review mode and is not actually used to en- or disable the review buttons.
+ *
+ * Teh default state is active, so if no preference is changed, the 'Mabye' option is enabled.
+ */
```
domienderle:
```suggestion
/**
 * Stores, whether 'Maybe' is allowed as the decision for a review or not.
 * This is only used to perfect the visualization of the review mode and is not used to en- or disable the review buttons.
 *
 * The default state is active, so if no preference is changed, the 'Maybe' option is enabled.
 */
```
I know this is nitpicky, but it's what WebStorms grammar corrector suggests.

src/routes/project/[projectId]/settings/slr/+page.svelte
```diff
@@ -1,8 +1,42 @@
...
     import ProjectSettingsLayout from "$lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte";
+    import MaybeAsDecisionSetting from "$lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte";
+    import Alert from "$lib/components/composites/utils/Alert.svelte";
+    import { MemberRole, ProjectStatus } from "$lib/model/api/project.js";
+    import { resource } from "$lib/resource.svelte.js";
+    import { onMount } from "svelte";
 
     let { data } = $props();
-    const { projectId, loadingProject } = data;
+    const { user, projectId, loadingProject, loadingMemebers } = data;
```
Comment (domienderle) [inline src/routes/project/[projectId]/settings/slr/+page.svelte]: Please refactor the variable `loadingMemebers` to `loadingMembers`.

src/routes/project/[projectId]/settings/slr/+page.svelte
```diff
@@ -1,8 +1,42 @@
...
+    );
+
+    onMount(() => {
+        async function checkIfSLRSettingsAreLocked() {
+            return await loadingProject
+                .then((project) => {
+                    return project.status !== ProjectStatus.ACTIVE;
+                })
+                .catch(() => {
+                    return false; // If loading fails, assume settings are not locked
```
Comment (domienderle) [inline src/routes/project/[projectId]/settings/slr/+page.svelte]: I am not sure if this is a good behaviour. Even though it is more restrictive to the user I would prefer to lock the settings if the loading fails. Otherwise it takes the risk that something unforseen happens if the user is able to change the settings in case he is normallly not longer allowed to.

src/routes/project/[projectId]/settings/slr/+page.svelte
```diff
@@ -15,4 +49,15 @@
     {/await}
 </svelte:head>
 
-<ProjectSettingsLayout {projectId} selectedTab="slr">SLR content</ProjectSettingsLayout>
+<ProjectSettingsLayout {projectId} selectedTab="slr">
+    {#if slrSettingsLocked}
```
Comment (domienderle) [inline src/routes/project/[projectId]/settings/slr/+page.svelte]: Currently this is never shown, because I am still able to change the settings, even after reviewing the first paper.

src/routes/project/[projectId]/settings/slr/+page.ts
```diff
@@ -0,0 +1,13 @@
+import type { PageLoad } from "../../$types";
+import { loadMembers } from "../members/helper";
+
+export const load: PageLoad = ({ params }) => {
+    const loadingMemebers = loadMembers({ id: params.projectId });
```
Comment (domienderle) [inline src/routes/project/[projectId]/settings/slr/+page.ts]: Same here. Change to `loadingMembers`

tests/e2e/project/settings/slr/project-slr-settings.test.ts
domienderle, line 1: Since I am currently refactoring the e2e tests I free you of refactoring the tests, because I can do this too. But please add test cases for the following points:
- If the confirm dialog is shown 
- If the project is already active it is not possible to change the settings (but this doesn't work correctly at the moment)

tests/example-data.ts
```diff
@@ -95,21 +96,29 @@ export const Authors = {
...
+        settings: createProjectSettings(),
     },
     demoProjectArchived: {
         id: "1",
         name: "Demo Project (archived)",
         status: ProjectStatus.ARCHIVED,
         currentStage: 3n,
         maxStage: 3n,
-        settings: undefined,
+        settings: createProjectSettings(),
```
domienderle, line 121: What is this for?

tests/setupTest.ts
```diff
@@ -233,6 +233,9 @@ vi.mock("$lib/grpc-api", () => {
             getPaperPdf: vi.fn(),
             setPaperPdf: vi.fn(),
             updateProjectMemberRole: vi.fn(),
+            getNextPaper: vi.fn(),
+            getNextPaperToReview: vi.fn(),
+            getPreviousPaper: vi.fn(),
```
domienderle: That's weird. Why isn't this already on dev? Shouldn't the linter have errors on dev then because of this?

tests/integration/settings/project-settings/slr/project-slr-settings.test.ts
domienderle, line 1: Add a test that checks, if the settings option is disabled if the project is already active.

domienderle (changes requested):
Looks good. But a couple of changes have to be made.
- I am not sure what the final decision was about toggeling the maybe settings button after at least one project is marked with maybe. Because in the current implementation the `Accept` and `Decline` button are both disabled if I submit a review with `Maybe` and disabled the maybe option afterwards. Let me now the current decision on this behaviour. Is this intended to be so at the moment?
- I would propose to disable the button after the setting is changed. If the confirm button in the dialog is pressed the user is allowed to press the button again until the change is applied. This allows to "change" the setting again without anything happening. You can have a look on this behaviour if you slow down the backend.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
mowi12: You're right. This is much more convenient. I think, I needed the separate projectSettings variable for something else, but this piece of code doesn't exist any more. Therefore, the settings of the project can be updated directly.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
mowi12: I'm not quite sure which function you mean, since you marked the ending of the `handleCancelClick()` function and the start of the `onMount()` function.
Imo the `handleActionClick()` and the `handleCancelClick()` functions are short enough that comments are not necessary.

And the `onMount()` function only makes one single call to the backend.
And since `onMount()` functions almost always fetch something at the start of the component, there is no need to explicitly tell the user that this functions fetches something at the start of the component.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
mowi12: You're correct, the `slrSettingsLocked` prop is missing. I added it to the usage and added a sentence to explain its' functionality.

src/lib/global-state/maybe-as-decision-state.svelte.ts
mowi12: No, this is good. I agree that your (WebStorms) suggestion sounds better than mine!

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/slr/+page.svelte:10]: Sure.
I also renamed the same symbol in the `+page.ts` file in the same directory. It had the same issue.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: See reply comment.

tests/example-data.ts
mowi12: Well, in order to be able to use the settings, mainly the `reviewMaybeAllowed` setting, a project needs to have settings. Therefore, I create settings instead of using undefined.

tests/setupTest.ts
mowi12: No, that's perfectly fine.
The current checked out API version was 0.2.0, which doesn't inlcude these calls. Only v0.3.0 (you're current PR I'm reviewing) and my v0.4.0 needs these calls.

tests/integration/settings/project-settings/slr/project-slr-settings.test.ts
mowi12: I renamed this file since it's not testing the project slr settings but only the maybe as decision setting.
Therefore, the check if the warning is visible should not go here.

But since this warning does not belong to any component. It cannot be checked for in a integration test that easily. I tried it and didn't find a way. Maybe you have an idea.

tests/e2e/project/settings/slr/project-slr-settings.test.ts
mowi12: Regarding 1.: Why should I test this? I test this by using the button dialog in two tests.
Regarding 2.: See reply comment.

mowi12 (commented):
Regarding the general workflow:
Each project can be in several states: ACTIVE, ACTIVE_LOCKED, etc.
A project starts in the ACTIVE state. As soon as a review gets made, the state changes to ACTIVE_LOCKED. This is fully handled by the backend (not implemented in the mock-backend since its too much work for a feature that only locks settings and can simply be tested with integration and e2e tests).

Regarding some of your comments to this issue.
You're right. Although projects already have reviews, their state is still ACTIVE. Therefore, we can still change the setting, although this should not be allowed. So simply ignore this and just closely look at the tests that they are working correctly and test the wanted behaviour.

In two converstations I wrote "see reply comment". I mean this comment.
If there are any other questions please ask me.

src/lib/components/composites/settings/project-settings/slr/MaybeAsDecisionSetting.svelte
domienderle: Since this comment would only be for the sake of more documentation and is not needed necessarily it is ok for me with your argument.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (domienderle) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: You didn't comment on this one. What are your thoughts about this?

tests/e2e/project/settings/slr/project-slr-settings.test.ts
domienderle: Regarding 1: Never mind. I don't know what I saw there. You do already test it.
Regarding 2: All right. In this case add at least a `TODO` for this test case once it is implemented in the backend.

tests/setupTest.ts
domienderle: Ah true. Didn't see that. Then it is fine.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (domienderle) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: Does it make sense to show this alert for every project member? Imo it only makes sense to show this to the admin, how is the only one that sees the settings and is able to change them.

tests/integration/settings/project-settings/slr/project-slr-settings.test.ts
domienderle: I see your problem here. At the moment I don't have a better idea to do this. But then add a `TODO` to the e2e tests, that we at least test this in the e2e test if it is possible due to the backend.

domienderle (changes requested):
A couple smaller changes are still left.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: You're right. But in this case I have another idea. But this needs to be discussed with the team.
The idea I have, is to simply not show the SLR settings tab in the left menu when the user is not an admin since all the settings are for the admin, or not?

Then there would be no `if admin` snippet in this file at all.
What do you think about that? I will also ask the team.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: Sorry, I missed this. You're right. This makes more sense. I will update it.

src/routes/project/[projectId]/settings/slr/+page.svelte
Comment (mowi12) [reply src/routes/project/[projectId]/settings/slr/+page.svelte]: I already implemented this, as it was already approved.

domienderle (changes requested):
Looks really good. Can you refactor the method `isCurrentUserAdmin`? Because at the moment you use the same code in four different files.

domienderle (changes requested):
All right. Now just have a look at the test. If the run through the pipeline we are ready to merge.

mowi12:
While refactoring, I chose the wrong variant to extract and lost one important feature. This is now back.
I had to make some changes to the components, but they work now!

domienderle (changes requested):
You can still toggle the button after I confirmed to change the settings. The scenario I mean is: 
- Click on the toggle button
- Confirm the change
- While updating the settings click the toggle button again (this shouldn't be possible)
- If you now confirm you get another success toast, but the toggle button has the same value as before

To fix this modify the toggle button, that it is disabled while the settings are updated.

domienderle (approved):

PR #427, opened 2025-05-30
Add Alert Variants

### LINKED ISSUE #426: [Feature]: Alert Variants

_No response_

We currently only have the ErrorAlert as component. There is no warning variant, or else.

Similar to the shadcn toasts, i want variants for the alert.

_No response_

### Anything else?

_No response_

mowi12:
Closes #426 

## What I have made

- Refactored error alert to a customizable alert
  - updated components
  - updated tests

## Checklist

  - [ ] ~~unit tests~~ (not required)
  - [x] integration tests
  - [ ] ~~end-to-end tests~~(not required)


src/lib/components/primitives/button/button.svelte
```diff
@@ -8,9 +8,8 @@
         variants: {
             variant: {
                 default: "bg-primary text-primary-foreground hover:bg-primary/90",
-                destructive: "bg-destructive text-destructive-foreground hover:bg-destructive/90",
-                destructiveSubtle:
-                    "bg-gray-100 text-red-600 border border-gray-300 hover:bg-gray-200",
+                error: "bg-error text-error-foreground hover:bg-error/90",
+                errorSubtle: "bg-gray-100 text-red-600 border border-gray-300 hover:bg-gray-200",
```
Slartibartfass2: I think `destructive` is fine here. `error` as a variant for a button seems a bit weird in my opinion. Why should anyone use an error button? `destructive` signals the consequences of clicking the button. The same goes for the alert dialog

Slartibartfass2 (changes requested):
Looks fine so far. Only this minor discussion.
Could you also add an example of each variant to the theme page?

src/lib/components/primitives/button/button.svelte
mowi12: You're right. I will change it back to 'destructive'.

src/lib/components/composites/dialog/AlertDialog.svelte
```diff
@@ -93,10 +93,10 @@ Usage:
     <AlertDialog
         title="This is the title of the dialog"
         actionProps={{
-            variant: "destructive",
+            variant: "error",
```
Slartibartfass2:
```suggestion
            variant: "destructive",
```

src/lib/components/composites/dialog/AlertDialog.svelte
```diff
@@ -93,10 +93,10 @@ Usage:
...
             onclick: () => console.log("Action button clicked"),
         }}
-        actionButtonText="Execute Destructive Action"
+        actionButtonText="Execute Error Action"
```
Slartibartfass2:
```suggestion
        actionButtonText="Execute Destructive Action"
```

Slartibartfass2 (changes requested):
Minor
Fix the linting errors.

Slartibartfass2 (approved):

PR #436, opened 2025-06-03
Fix breaking chips input

### LINKED ISSUE #377: [Bug]: Breaking Chips Input

_No response_

### What happened?

When entering a very long input into a chips input field, this can sometimes break the UI.

### Steps to reproduce

Go to a project, and add a user to the project. (You need to be an admin). Enter a very long email address.

Image

You should receive something like this.

### What did you expect?

I have no solution for this issue, but maybe the input scrolls (similar to while entering the long word).

Or to wrap it, but I don't know how this would look.

At least, the buttons, should stay inside the container!

### Anything else?

_No response_

Merseleo:
Closes #377

## What I have made

- I added the class `overflow-x-auto` to the form around the `InviteUsersInput` containing a `ChipsInput` component so long content does not break the UI but is hidden and can be scrolled into view

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - ~~[ ] end-to-end tests~~

### DIFF
--- a/src/lib/components/composites/settings/project-settings/members/InviteUsersDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/members/InviteUsersDialog.svelte
@@ -117,7 +117,11 @@ Usage:
         Search for an existing user or invite a new user by email.
     {/snippet}
     {#snippet content()}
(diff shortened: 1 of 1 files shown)


mowi12 (approved):
LGTM

PR #444, opened 2025-06-04
Fix outdated wiki

### LINKED ISSUE #443: [Bug]: Fix wiki job

_No response_

### What happened?

The wiki is currently out-of-date.

### Steps to reproduce

Fix it.

### What did you expect?

A working wiki.

### Anything else?

_No response_

Slartibartfass2:
Closes #443
## What I have made

The current wiki is outdated for several reasons:
- the job failed due to broken links
- the job wasn't triggered when important changes were made

To fix this, we also run the linting job in PRs and the default branches.
Additionally, we change the GitHub URLs so that they point to the current branch. This way, we can detect broken links with the current state of the wiki. Otherwise, we would only notice this when it's too late and the changes were already pushed to develop.

## Checklist

### DIFF
--- a/.github/workflows/wiki.yml
+++ b/.github/workflows/wiki.yml
@@ -2,8 +2,14 @@ name: Lint and Publish Wiki
 
 on:
     push:
-        paths:
-            - "**/*.md"
+        branches:
+            - main
+            - develop
+    pull_request:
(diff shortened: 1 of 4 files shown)


mowi12 (approved):
LGTM

PR #445, opened 2025-06-05
Fix misleading hint shown when no search results could be found

### LINKED ISSUE #374: [Bug]: Misleading hint shown when no search results could be found

_No response_

### What happened?

When you search through a list using the custom search component and no results could be found, then the normal `emptyHint` is shown. Actually this hint should only be used if initially the list is empty and not if it is empty after searching or filtering.

### Steps to reproduce

1. Go to the reading list page 
2. Search for an paper that does not exist on the reading list

### What did you expect?

I would expect a more precise hint, e. g. "No search results" or "No results found".

### Anything else?

Image

Merseleo:
Closes #374

## What I have made

- I changed the hint displayed when no search results could be found
  - on the reading list page
  - on the project papers overview page
  - in the lists of the references and citations on the paper view page

## Checklist

- [x] I have adapted tests that prove my fix is effective or that my feature works
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
@@ -54,7 +54,7 @@ Usage:
         <NamedList
             emptyHint={noSearchResultsForBackwardReferencedPapers
                 ? "No references match your search."
-                : "No references found."}
+                : "No references exist or have been added."}
             errorHint="Couldn't load references."
             items={backwardReferencedPapers}
             keySelector={(paper) => paper.id}
@@ -77,7 +77,7 @@ Usage:
         <NamedList
             emptyHint={noSearchResultsForForwardReferencedPapers
                 ? "No citations match your search."
-                : "No citations found."}
+                : "No citations exist or have been added."}
             errorHint="Couldn't load citations."
             items={forwardReferencedPapers}
             keySelector={(paper) => paper.id}
--- a/src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
@@ -18,17 +18,25 @@
 
     let backwardReferencedPapers = $state<Promise<Paper[]>>(allBackwardReferencedPapers);
     let forwardReferencedPapers = $state<Promise<Paper[]>>(allForwardReferencedPapers);
+    let noSearchResultsForBackwardReferencedPapers = $state(false);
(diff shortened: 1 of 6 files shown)


src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -44,7 +52,9 @@ Usage:
 <section class="flex h-full flex-col gap-5">
     <div class="flex h-full flex-[1_1_0] overflow-hidden">
         <NamedList
-            emptyHint="No references found."
+            emptyHint={noSearchResultsForBackwardReferencedPapers
+                ? "No references match your search."
+                : "No references found."}
```
mowi12: 'No references match your search' and 'No references found' both sound like the search yielded no result. I think 'No references found' should be something like 'No references exist'. It should be more clear that there simply are no references, not that they couldn't have been found.

'No references exist' is not perfect, maybe you can come up with something better!

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
```diff
@@ -65,7 +75,9 @@ Usage:
     </div>
     <div class="flex h-full flex-[1_1_0] overflow-hidden">
         <NamedList
-            emptyHint="No citations found."
+            emptyHint={noSearchResultsForForwardReferencedPapers
+                ? "No citations match your search."
+                : "No citations found."}
```
mowi12: Same here.

mowi12 (changes requested):
I like your changes, well done! There is only one sentence, where I don't like the exact wording. Maybe you don't see it that way, so please let me know what you think.

And regarding the wiki linting, I don't understand the error. Locally, I cannot reproduce the error. The file isn't even that long in that specific line.

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
Merseleo: You are right, it sounds quite similar. But imo 'No references exist' is really good, as this actually the case.

Merseleo (commented):
Thank you for your feedback and suggestions! What do you think of my adaption of your suggestion?

src/lib/components/composites/paper-components/paper-view/cards/ReferencesAndCitationsCardContent.svelte
mowi12: I guess you're right. I like the phrasing you chose!

mowi12 (approved):
Perfect, nicely done!

PR #447, opened 2025-06-06
Use linux-sp runner

### LINKED ISSUE #446: [CI]: Change runner to `linux-sp`

_No response_

We can now use the custom university runner.

Change the workflow files so that they use the custom runner.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #446

## What I have made

- Changed the runner to `linux-sp`

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (not required)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ env:
 jobs:
     build:
         name: Build
-        runs-on: ubuntu-latest
+        runs-on: linux-sp
         steps:
             - name: Checkout code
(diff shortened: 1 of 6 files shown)


Merseleo:
Why do want to change the runner? Do we still need this?

Slartibartfass2:
> Why do want to change the runner? Do we still need this?

The idea was that the university runner is better but we currently don't have problems with the GitHub runners.

PR #448, opened 2025-06-07
Suboptimal user context handling

### LINKED ISSUE #440: [Bug]: Suboptimal User Context Handling

_No response_

### What happened?

There are two things which are not implemented very efficient.
First, on the initial start of the mock-backend and the frontend, the 'Create Project'. Button is disabled because the user in the context is null. This shouldn't be, or at least, it should update as soon as the user is set.

Second, the user gets called before every new page load since it is being fetched in the root layout.ts file.

### Steps to reproduce

1. Close mock-backend and frontend, and all tabs in the browser which belong to the SnowballR application (frontend)
2. Start the mock-backend
3. Start the frontend
4. Log in (with alice, e.g.)

### What did you expect?

Regarding (1): The button should enable as soon as the user gets loaded

Regarding (2): As soon as the user is set in the context, it only needs to be fetched again if invalidated.

### Anything else?

_No response_

mowi12:
Closes #440 

## What I have made

- Introduced a new user session cache
- Updated root layout.ts and page.ts to use this cache and reduce fetches
- Fix error output in page.ts on unCheckedPaths
- Fix disabled 'Create Project' button due to initial user being null

## Checklist

  - [ ] ~~unit tests~~ not needed
  - [ ] ~~integration tests~~ everything still works
  - [ ] ~~end-to-end tests~~ everything still works


mowi12:
For the reviewer (probably @Slartibartfass2 or @Merseleo) as explanation why I did not use a state directly to store the user:
A state should always be initialized. Since I would create the state in a separate file, the initial state would be `undefined`. This would also mean that we have to modify all files which use the user context. This way we have better handling of the cached user and do not need to modify all file which use the cached user.

I ask AI for an approach, and it also used this way and not a state to handle the user directly.
Online, I did not find any other way how to correctly use this, so Imo, this solution is one of the best ways to do this.
I'm very curious, what your opinion to this solution is!

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -34,11 +34,30 @@
...
+            isErrorOnUsersLoading = false;
+            return;
+        }
+
+        isLoadingUsers = true;
+        loadUsers(currentUser)
+            .then((result) => {
+                initialPossibleMembers = result.initialPossibleMembers;
+                isErrorOnUsersLoading = result.isErrorOnUsersLoading;
+                isLoadingUsers = false;
```
Merseleo: I think this is not necessary, as you finally set it to false and finally runs always?!

src/lib/components/composites/project-components/CreateProjectDialog.svelte
```diff
@@ -34,11 +34,30 @@
     let isLoadingUsers = $state(true);
     let initialPossibleMembers: User[] = $state([]);
 
-    onMount(async () => {
-        const result = await loadUsers(user);
-        initialPossibleMembers = result.initialPossibleMembers;
-        isErrorOnUsersLoading = result.isErrorOnUsersLoading;
-        isLoadingUsers = false;
+    $effect(() => {
```
Merseleo, line 37: Why this have to be an effect? Because the component can be mounted before the user gets loaded?

src/lib/user-state/userContext.ts
Merseleo, line 1: Why did you renamed it? Because I thought it is still a context?

src/lib/user-state/userSessionStore.ts
```diff
@@ -0,0 +1,39 @@
+import { invalidate } from "$app/navigation";
+import type { User } from "$lib/model/api/user";
+
+export const USER_SESSION_DEPENDENCY_KEY = "data:getCurrentUser";
+
+// Internal cache for the user object.
+// undefined: State is unknown, never fetched, or cache explicitly invalidated
+// null: Fetch attempted, but user is confirmed not authenticated, or an error implies no user
+// User: Fetch successful, user is authenticated
+let SVELTE_KIT_INTERNAL_cachedUser: User | null | undefined = undefined;
```
Merseleo: What is this for a strange naming convention?

src/lib/user-state/userSessionStore.ts
```diff
@@ -0,0 +1,39 @@
+import { invalidate } from "$app/navigation";
+import type { User } from "$lib/model/api/user";
+
+export const USER_SESSION_DEPENDENCY_KEY = "data:getCurrentUser";
+
+// Internal cache for the user object.
+// undefined: State is unknown, never fetched, or cache explicitly invalidated
+// null: Fetch attempted, but user is confirmed not authenticated, or an error implies no user
+// User: Fetch successful, user is authenticated
```
Merseleo:
```suggestion
/** Internal cache for the user object.
 * | Value | Description |
 * | ----- | ----------- |
 * | undefined | the state is unknown, the user never was fetched, or the cache has explicitly invalidated |
 * | null | the user was attempted to be fetched, but user is confirmed not authenticated, or an error implies no user |
 * | User | the user was fetched successfully and is authenticated |
 */
```

src/routes/+page.ts
```diff
@@ -42,7 +43,17 @@ export const load: PageLoad = async ({ depends, parent }) => {
...
+    const user = parentData.user as User | null;
+
+    // If the user is not logged in, we return empty arrays for projects and open reviews
+    // This is to ensure that the page can still render without errors
+    if (!user) {
+        return {
+            projectsMetadata: Promise.resolve([]),
+            openReviews: Promise.resolve([]),
+        };
+    }
```
Merseleo, line 55: But how is this possible that the user is null on this route? I thought it can only be null if a user is on unchecked routes?

src/lib/user-state/userSessionStore.ts
```diff
@@ -0,0 +1,39 @@
...
+ * This is primarily for use by the root layout load function after a successful fetch or auth check.
+ *
+ * @param user - The user object to cache, or null if the user is not authenticated.
+ */
+export function setCachedUser(user: User | null): void {
+    SVELTE_KIT_INTERNAL_cachedUser = user;
+}
+
+/**
+ * Invalidate the client-side user cache and trigger a re-fetch of the user session in the root layout.
```
Merseleo:
```suggestion
 * Invalidates the client-side user cache and trigger a re-fetch of the user session in the root layout.
```
Maybe user sessions is not the best wording as it could be confusing with normal login sessions?

src/lib/user-state/userSessionStore.ts
Merseleo: And, however, I do not quite understand why we do not use a global state here? Is it not possible to initialize a global state with undefined as well? And because we only have a CSR application now it should be save?!

src/lib/user-state/userSessionStore.ts
```diff
@@ -0,0 +1,39 @@
...
+// null: Fetch attempted, but user is confirmed not authenticated, or an error implies no user
+// User: Fetch successful, user is authenticated
+let SVELTE_KIT_INTERNAL_cachedUser: User | null | undefined = undefined;
+
+/**
+ * Retrieves the internal cached user object.
+ * This is primarily for use by the root layout load function.
+ *
+ * @returns The ached user object, or null if the user is not authenticated.
+ *          Returns undefined if the user state is unknown (e.g., not yet fetched or cache invalidated).
```
Merseleo:
```suggestion
 * @returns The cached user object, or null if the user is not authenticated or
 *                  undefined if the user state is unknown (e.g. not yet fetched or cache invalidated).
```

src/lib/user-state/userSessionStore.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function getCachedUser(): User | null | undefined {
+    return SVELTE_KIT_INTERNAL_cachedUser;
+}
+
+/**
+ * Updates the internal cached user cache.
```
Merseleo:
```suggestion
 * Updates the internal user cache.
```

src/routes/+layout.ts
```diff
@@ -18,41 +23,67 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
...
     const uncheckedPaths = ["/signin", "/signup", "/resetpassword"];
     const onUncheckedPath = uncheckedPaths.includes(url.pathname);
     if (onUncheckedPath) {
         return { user: nullUser };
     }
 
+    // If the user is already cached, we return the cached user
+    // If the user is not cached (undefined or null), we will fetch the user
+    const cachedUser = getCachedUser();
+    if (cachedUser !== undefined && cachedUser !== null) {
```
Merseleo:
```suggestion
    if (!cachedUser) {
```
Actually undefined and null should be falsy values and the user object should always be non empty and truthy?

src/routes/+layout.ts
```diff
@@ -18,41 +23,67 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const nullUser: User = null as any;
```
Merseleo: In general this seems like a very bad solution imo. First the use of any instead of `as unknown as User` and secondly why not explicitly use the type `User | null` or set it to an empty user object without id etc.?

src/routes/+layout.ts
```diff
@@ -18,41 +23,67 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
...
     );
 
     if (authStatusCall === undefined) {
+        setCachedUser(null);
         return { user: undefinedUser };
     }
 
     if (authStatusCall.status.code !== StatusCodes.OK) {
+        setCachedUser(null);
         return { user: undefinedUser };
```
Merseleo: Why do you return the `undefinedUser` but set the cached user to `nullUser`

Merseleo (changes requested):
So as general feedback: As far as I tested your changes it seems to work and the _CreateProjectDialog_ button is not disabled forever anymore and the user does not get refetched every time, but only if something changes, so great work! However there are some things considering the way you implemented the fix I do not quite understand or are not happy (without further explanation). Can you please explain some decisions to me in the conversations? 

Besides there are some spelling mistakes / inconsistencies to fix ;)

src/lib/components/composites/project-components/CreateProjectDialog.svelte
mowi12: Oh you're right. This, I didn't see.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
mowi12: Yes, correct! I don't understand why though. I added some `console.log` outputs just to see at which point the component gets mounted and this confirmed my suspicions. I don't know why, and I also did not find any other solution, so this is the best solution in my opinion.

src/lib/user-state/userContext.ts
mowi12: It's still a context, but I moved this file into one combined directory which contains everything for the user management.

mowi12: I renamed the directory to a better name so it doesn't user any svelte specific words.

src/lib/user-state/userSessionStore.ts
mowi12: You're right. Session is not the best wording. It only caches the current user, so I try to reflect that.

src/lib/user-state/userSessionStore.ts
mowi12: Yeah, this name is definitely no great choice. I'm sorry for that!

mowi12: > For the reviewer (probably @Slartibartfass2 or @Merseleo) as explanation why I did not use a state directly to store the user:
A state should always be initialized. Since I would create the state in a separate file, the initial state would be undefined. This would also mean that we have to modify all files which use the user context. This way we have better handling of the cached user and do not need to modify all file which use the cached user.
>
> I ask AI for an approach, and it also used this way and not a state to handle the user directly.
Online, I did not find any other way how to correctly use this, so Imo, this solution is one of the best ways to do this.
I'm very curious, what your opinion to this solution is!

I don't know if you saw the comment I wrote at the top. Here I tried to explain my initial thoughts.
In addition, we would have problems for the invalidation. Now we can just remove the cached user from the state and through that get a new user. If we directly store the user inside a state and only use that, we could not check if the current user exists and based on that, refresh the user. I hope you understand my point.
And setting a flag to indicate, that the user needs to be refreshed is not as good of a solution, imo.

src/routes/+page.ts
mowi12: As I understand it, the root page.ts is always executed, same as the root layout.ts. Therefore, the use can be null or undefined. I don't think there is any other solution than to check if the user object exists.

src/routes/+layout.ts
mowi12: You're right!

src/routes/+layout.ts
mowi12: This was luca-schlecker's idea. As I reviewed it and also didn't like this solution, but luca convinced me to keep it because it doesn't make any difference, since there either is a user or there is none but we are not interested what this non-existing user looks like.

But I agree, this is a bad solution. I will try to use a better solution!

mowi12: If I remember correctly, the initial idea was to differentiate between null user and undefined user. But I checked the other auth routes, and we never check anything. So imo, we can simply use an empty user object for all cases. If we need to differentiate, then we can add other versions.

src/routes/+layout.ts
mowi12: Not important any more as we now always return an empty user object.

src/lib/components/composites/project-components/CreateProjectDialog.svelte
Merseleo: Ok, that's actually a bit strange, but I am ok with that.

src/lib/user-state/userContext.ts
Merseleo: Ok.

src/lib/user-state/userSessionStore.ts
Merseleo: In my opinion you improved the wording a lot, however, you still are inconsistent with _Invalidate_ and _Retrieve**s**_ at the beginning of your comments.

src/routes/+layout.ts
Merseleo: Sorry my mistake: of course it need to be 
```typescript
if (cachedUser) {
     return { user: cachedUser };
}
``` 
and not 
```typescript
if (!cachedUser) {
     return { user: cachedUser };
}
```

src/lib/user-state/userSessionStore.ts
Merseleo: I read your comment but did not understand it at the beginning. But especially your second argument with the invalidation makes sense to me.

src/routes/+page.ts
Merseleo: Yeah it is always executed and if the user is null we redirect to the login page, dont we? Did you face problems when you did not make this check?

src/routes/+layout.ts
Merseleo: Ok.

src/routes/+layout.ts
```diff
@@ -30,29 +35,47 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
...
+                setCachedUser(null);
+                await goto("/signin");
+                return { user: emptyUser };
+            }
+            // Renewal successful, proceed to fetch current user
+        } catch (error) {
+            console.error(`Session renewal failed: ${error}`);
+            setCachedUser(null);
+            await goto("/signin");
+            return { user: emptyUser };
```
Merseleo: So now you very often use this code snippet of setting the cached user, optionally navigate and return the emptyUser. Do you want to outsource this into a separate function to reduce the duplicate code? Not necessary, only a suggestion.

Merseleo (changes requested):
Ok, I think there are just a few cosmetic issues to fix and a mistake on my part.

PS: Why did you cancel the E2E tests? Or was there another problem canceling them?

Slartibartfass2:
Hi, as I was also requested here my only point that, I think, is also important:

Please provide a brief description in the wiki on how to use the user context.

src/routes/+page.ts
mowi12: Yes, before we got an error because the user is null, and we try to get the projects for the user.
As this won't work, we need to return these Promises.

If you have a better solution, please let me know!

src/routes/+layout.ts
mowi12: Ahh. My bad. This is also the reason why the E2E tests fail. Because we now always return null or undefined and never the cached user.

src/routes/+layout.ts
mowi12: Yeah, why not, DRY is always a good principle to use.

mowi12 (commented):
As mentioned in a conversation, the E2E failed because of the `!cachedUser` check, which led to always returning a null/undefined user.

I added an entry in the wiki explaining the most important parts, as @Slartibartfass2  requested.

I also removed the backend unreachable as we never reach this state.
We either have not user -> fetch/redirect
We have a cached user -> not needed

src/routes/+layout.ts
```diff
@@ -30,29 +35,52 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
...
+        return await redirectToSignIn();
     }
 };
+
+/**
+ * Redirects to the sign-in page and clears the user cache.
+ * This function is used when the user is not authenticated or when an error occurs
+ * that implies the user is not authenticated.
+ *
+ * @returns Redirects to the sign-in page and returns an empty user object.
```
Merseleo:
```suggestion
 * @returns An empty user object.
```

Merseleo: You already wrote in your comment, that you redirect ...

src/routes/+layout.svelte
Merseleo, line 1: Why did you removed this alert dialog?

src/routes/+layout.ts
Merseleo: Right.

src/routes/+page.ts
Merseleo: No not really. Probably the best solution considering our knowledge of Svelte

src/routes/+layout.svelte
Merseleo: I agree with you that checking whether the user is undefined is not correct anymore, however, when the backend is not reachable because of whatever reason we does not show anything in the application or at least there is no interaction possible but the user doesn't really know why. So either we check this separately and navigate to an error page, that we are unavailable at the moment. What do you think?

wiki/Contributing.md
```diff
@@ -75,6 +79,100 @@ create branches from an issue as it already provides `<issue-number>-<short-desc
...
+### How the User is Fetched and Made Available
+
+The initial user loading logic lives in the root layout.ts file. It checks whether a user is already cached using the
+userCache. If no cached user is found, it attempts to determine the current authentication status and fetch the user
+accordingly. If the user is unauthenticated or the session is invalid, it tries to renew the session or redirects to
+the sign-in page if necessary.
+
+The resulting user object—whether fetched, cached, or an explicit "empty" user—is returned from layout.ts and injected
+into the user context by the root layout.svelte file. This ensures the user state is consistent and accessible
+throughout the app.
```
Merseleo, line 116: I think this logic would be easier to understand if you read it in a diagram, e. g. state diagram or activity diagram.

wiki/Contributing.md
```diff
@@ -75,6 +79,100 @@ create branches from an issue as it already provides `<issue-number>-<short-desc
...
+              toast.error("Failed to update user details.");
+          }
+   
+    }
+   </script>
+   ```
+
+   Calling `triggerCurrentUserRefresh()` invalidates the `USER_DEPENDENCY_KEY`, by calling
+   `invalidate(USER_DEPENDENCY_KEY)`. This prompts SvelteKit to re-run the root `load` function which re-fetches the
+   updated user from the backend.
```
Merseleo: I think this is not necessary as it is a detail interested people can look up on their own.

Merseleo (changes requested):
I still have some little questions and suggestions, sorry that this PR is taking so long.

src/routes/+layout.ts
mowi12: You're right.

src/routes/+layout.svelte
mowi12: Exactly, we cannot check if the user is null/undefined as this should never be the case.
The problem really is only, how do we find out if the backend is reachable or not?
Currently if the user cannot be fetched we are simply redirected to the sign-in page and stay there and get an error message if the user cannot be fetched.

I don't know how to differentiate in this case.
Either way, imo this should be a new issue ("Add Error Alert - Backend not Reachable") or something.
if it is even needed.

wiki/Contributing.md
mowi12: You're right. I will add one.

wiki/Contributing.md
mowi12: You're right. This is not really needed.

src/routes/+layout.svelte
Merseleo: Yeah, you are right. Let's think about this problem as soon as we finish the possible error codes and we need to integrate the new specifications in the frontend.

wiki/Contributing.md
```diff
@@ -75,6 +79,116 @@ create branches from an issue as it already provides `<issue-number>-<short-desc
...
+accordingly. If the user is unauthenticated or the session is invalid, it tries to renew the session or redirects to
+the sign-in page if necessary.
+
+The resulting user object—whether fetched, cached, or an explicit "empty" user—is returned from layout.ts and injected
+into the user context by the root layout.svelte file. This ensures the user state is consistent and accessible
+throughout the app.
+
+Components can safely assume that the user object is always valid and never null or undefined. In cases where the user
+is not authenticated, the app will redirect to the sign-in page before components are rendered. This eliminates the
+need for null checks in downstream components that rely on user data.
```
Merseleo:
```suggestion
The initial user loading logic lives in the root `layout.ts` file and is described by flow chart above.

If the session is expired, the sessions is tried to be renewed. It this fails, the user is unauthenticated or any other error occurred, the user is redirected to the sign-in page.

Components throughout the entire application can safely access the user context and assume that the user object is always valid and never `null` or `undefined`. In cases where the user
is not authenticated, the app will redirect to the sign-in page before components are rendered. This eliminates the need for null checks in downstream components that rely on user data.
```

Merseleo (dismissed):
Looks good, I think we are finished. I added a suggestion to the description of the user loading process, but can also ignore this comment and merge.

wiki/Contributing.md
mowi12: I added the `null`, and `undefined` suggestions. The formatting I had to dismiss, since the line length is capped and 120 chars.

Merseleo (approved):

PR #449, opened 2025-06-09
Added missing tests

### LINKED ISSUE #390: [Refactor]: Add missing e2e tests

_No response_

A lot of components were finished before e2e tests were added. These components/workflows are not tested yet.

E2E tests for all untested parts of the project. As well as integration tests if needed.

- Until now the `reviewMode` in the `paper-navigation-button.tests.ts` e2e test is toggled via 
`await page.evaluate(() => {
    localStorage.setItem("reviewMode", JSON.stringify(false));
});`
As soon as the test suit for the review mode toggle button exist, adjust this tests to use the methods of the created model.

_No response_

### Anything else?

_No response_

domienderle:
Closes #390 

## What I have made

I have refactored all e2e tests, fixtures and models. The goal was to create a strict separation of the page models, the fixtures and the test cases. Therefore I created a model for all existing pages and dialogs and tried to include as many `page.getSomething()` calls into the models, so they can be managed and maintained centralized in the specific models. All of the `page.getSomething()` calls that are still in the test cases are test case specific calls. 

The whole setup and teardown `beforeAll`, `afterAll`, ... is now handled in the `fixtures`. Additionally, each page model that is used in the tests gets included as a single fixture, so each test only uses the ressources it really needs.

Thus, the test classes only include the test cases. Due to the fixtures and the page models, the test cases now only must perform the actions according to the workflow and don't have to handle anything else. 

## Checklist

  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - [x] end-to-end tests


tests/e2e/navigation-bar-model.ts
mowi12, line 1: Why did you choose `false` as default value for the `shortcuts` parameter?
The initial value and therefore the default value in the `shortcuts-visibility-state.svelte.ts` file is `true`.

This would make the usage of these functions a lot easier, since only the tests for the shortcuts visibility would need to actively set the `shortcuts` variable to `false` if needed.

tests/e2e/navigation-bar-model.ts
```diff
@@ -0,0 +1,90 @@
...
+        this.page = page;
+        this.userButton = this.page.getByRole("button", { name: /^[A-Z]{2}$/ });
+        this.goBackButton = this.page.getByRole("link", { name: "Back to" });
+        this.dashboardTab = this.page.getByRole("tab", { name: "Dashboard" });
+        this.papersTab = this.page.getByRole("tab", { name: "Papers" });
+        this.statisticsTab = this.page.getByRole("tab", { name: "Statistics" });
+        this.settingsTab = this.page.getByRole("tab", { name: "Settings" });
+    }
+
+    getCustomUserButton(initials: string) {
```
mowi12: Why is this called `customUser`? The component itself is called `UserAvatar`, so why not call the function `getUserAvatar` or `getUserAvatarButton`?

Imo, `getUserAvatar` would fit best, although it is a button, since the component is called this way and happens to be a button. (It doesn't really contain a button, but a tooltip, but in the background it functions as a button and has the role of a button.)

tests/e2e/navigation-bar-model.ts
```diff
@@ -0,0 +1,90 @@
...
+    readonly userButton: Locator;
+    readonly goBackButton: Locator;
+    readonly dashboardTab: Locator;
+    readonly papersTab: Locator;
+    readonly statisticsTab: Locator;
+    readonly settingsTab: Locator;
+
+    constructor(page: Page) {
+        this.page = page;
+        this.userButton = this.page.getByRole("button", { name: /^[A-Z]{2}$/ });
```
mowi12: Why is there this variable `userButton` if you do not use it in the `getCustomUserButton`?
Don't they reference the same thing?

If you want one to be specific, rather add a parameter or just make the initials optional.

tests/e2e/navigation-bar-model.ts
mowi12: Documentation is missing.

tests/e2e/navigation-bar-model.ts
mowi12, line 1: Do you think `NavigationBarModel` is the correct name for this component?
We have different variations and this one rather corresponds to the project version.

tests/e2e/homepage/home-page-model.ts
```diff
@@ -1,48 +1,56 @@
...
     readonly page: Page;
+    readonly heading: Locator;
+    readonly userMenuButton: Locator;
     readonly createProjectDialog: Locator;
     readonly openCreateProjectDialogButton: Locator;
 
     constructor(page: Page) {
         this.page = page;
+        this.heading = page.getByRole("heading", { name: "SnowballR", exact: true });
+        this.userMenuButton = page.getByRole("button", { name: /^[A-Z]{2}$/ });
```
mowi12: Why do we need the `userMenuButton` in this model?
If I understand it correctly, the `home-page-fixture` also includes the `navigationBar` which already provides this.

tests/e2e/homepage/home-page-model.ts
```diff
@@ -1,48 +1,56 @@
...
+import { ProjectPaperViewPageModel } from "$tests/e2e/project/paper/project-paper-view-model";
+import { ProjectDashboardModel } from "$tests/e2e/project/dashboard/project-dashboard-model";
 
-type LinkName = "Reading List" | "Archived Projects" | "Invitations" | "Settings" | "Sign Out";
-type SettingName = "Account" | "Project Setup" | "Shortcuts" | "Review";
-
-export class DevHomePage {
+export class HomePageModel {
     readonly page: Page;
+    readonly heading: Locator;
```
mowi12, line 7: Maybe a bit specific, but what about `snowballRHeading` since it always references only the snowballR heading. This would make it clear which heading it references.

tests/e2e/auth/signup/sign-up-page-fixtures.ts
```diff
@@ -0,0 +1,34 @@
...
+    signUpPage: SignUpPageModel;
+    homePage: HomePageModel;
+    signInPage: SignInPageModel;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - home page
+ * - the dialog for creating a new project
+ */
```
mowi12: I think you made a copy-paste mistake, please update the comments.

tests/e2e/auth/signup/signup.test.ts
mowi12, line 1: Please add a test for a failed API call.

mowi12, line 1: I know this is not part of this issue, but could you maybe "fix" the missing `errorDetails` in line 83 in file `/routes/(auth)/signup/+page.svelte`?

Just make the title an actual title and add a more detailed description if possible.

tests/e2e/auth/signout/signout-fixture.ts
```diff
@@ -0,0 +1,27 @@
...
+type SignOutPageFixtures = {
+    signInPage: SignInPageModel;
+    navigationBarModel: NavigationBarModel;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - home page
+ * - the dialog for creating a new project
+ */
```
mowi12: Copy-paste mistake again. Please update comment accordingly.

tests/e2e/navigation-bar-model.ts
mowi12, line 1: Up until this point, I never saw one reason why we need to get the link of a navigation in the user avatar menu.
Why don't the methods itself to the click?
So, not `get...Link`, but `navigateTo...` or `click...InUserMenu` or something like that.

In addition, you can simply add the click to open the user avatar menu inside this function, so we do not need to call this click separately in the tests.

EDIT: I think you did this, because all other models/fixtures also only return the links and need to manually clicked. After thinking about it. I actually like it that way, because it actually resembles the users workflow.

You can ignore the requested changes in here and can resolve this on you own.

tests/e2e/auth/signin/sign-in-page-fixtures.ts
```diff
@@ -0,0 +1,36 @@
...
+    signInPage: SignInPageModel;
+    homePage: HomePageModel;
+    signUpPage: SignUpPageModel;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - home page
+ * - the dialog for creating a new project
+ */
```
mowi12: Again copy-paste mistake. Please update the comments.

tests/e2e/auth/signin/signin.test.ts
```diff
@@ -1,54 +1,51 @@
...
 
 test.use({ user: null });
 
 test.describe("Sign In Functionality", () => {
-    test.beforeEach(async ({ apiClient, page }) => {
-        await apiClient.register(alice);
-        await page.goto("/signin");
-    });
-
     test("When the user is unauthenticated, then they will be redirected to /signin", async ({
```
mowi12: I think we should navigate to another site to prove that we will get redirected to `/signin`.
Currently, the fixture already redirected to `/signin`.

Just add a `goto("/")` or even some other path `/project/42` or something like that.
Maybe you need to add a `waitForUrl()` but this is okay, as this is the intended way I think.
Please check how playwright handles the waiting for calls like `goto`.

tests/e2e/auth/signin/signin.test.ts
```diff
@@ -1,54 +1,51 @@
...
-        await expect(page.getByText("Create Project", { exact: true })).toBeVisible();
+    test("When the user enters valid credentials, then they are logged in", async ({
+        signInPage,
+        homePage,
+    }) => {
+        await signInPage.emailInput.fill(alice.email);
+        await signInPage.passwordInput.fill(alice.password);
+        await signInPage.signInButton.click();
+        await expect(homePage.heading).toBeVisible();
+        await expect(homePage.openCreateProjectDialogButton).toBeVisible();
```
mowi12: As soon as my PR is done, which fixes the initial error where the create project button is disabled, I think we should test if this button is `enabled`. This way we can clearly say, that everything worked. If the button were `disabled`, then we can not say, that everything has loaded, or at least, most of it.

Maybe you have to write a custom `await waitFor()` function, because sometimes it can take a while for this button to get enabled. But this function could then be used everywhere.

tests/e2e/auth/signin/signin.test.ts
```diff
@@ -1,54 +1,51 @@
...
+        signInPage,
+        signUpPage,
     }) => {
-        await page.getByRole("link", { name: "Sign Up" }).click();
-        await page.waitForURL("/signup");
+        await signInPage.signUpLink.click();
+        await expect(signUpPage.signUpHeader).toBeVisible();
     });
 
     test("When the user clicks 'Forgot Password?', then they are redirected to /forgotpassword", async ({
```
mowi12: I really liked the wording you chose for the other tests, where you used sign-in page instead of the path. Could you also do this here?

tests/e2e/auth/signin/signin.test.ts
```diff
@@ -1,54 +1,51 @@
...
         page,
+        signInPage,
     }) => {
-        await page.getByRole("link", { name: "Forgot Password?" }).click();
+        await signInPage.forgotPasswordLink.click();
         await page.waitForURL("/resetpassword");
```
mowi12: Please add a TODO so we know, that this test needs to be updated when the reset password page is implemented.
Currently, we test for the URL, but not for the heading compared to the other tests.

tests/e2e/paper/paper-view-model.ts
```diff
@@ -0,0 +1,106 @@
...
+            exact: true,
+        });
+        this.allListEntries = page.getByRole("listitem");
+        this.generalInformationLabel = page.getByText("General Information");
+        this.publicationTypeLabel = page.getByText("Publication Type");
+        this.publicationNameLabel = page.getByText("Publication Name");
+        this.externalIdLabel = page.getByText("External ID");
+
+        this.projectName = "Project 1";
+        this.paperNames = ["Paper View Paper 0", "Paper View Paper 1"];
```
mowi12:
```suggestion
        this.paperNames = ["Paper View - Paper 0", "Paper View - Paper 1"];
```

This makes it easier to understand, and it doesn't sound weird.

tests/e2e/paper/paper-view-fixtures.ts
```diff
@@ -0,0 +1,78 @@
...
+    paperViewPage: PaperViewPageModel;
+    homePage: HomePageModel;
+    projectPaperViewPage: ProjectPaperViewPageModel;
+    readingListPage: ReadingListPageModel;
+    navigationBar: NavigationBarModel;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - paper view page
```
mowi12: No copy-paste error here, but now you provide more that one page object model. Please update the comment.

tests/e2e/auth/signin/signin.test.ts
mowi12, line 1: Please add a test for a failed API call.

tests/e2e/paper/paper-view.test.ts
```diff
@@ -0,0 +1,64 @@
+import { test } from "./paper-view-fixtures";
+import { expect } from "@playwright/test";
+
+test.describe("Paper View functionality", () => {
```
mowi12:
```suggestion
test.describe("Paper View Functionality", () => {
```

tests/e2e/readinglist/reading-list-page-model.ts
```diff
@@ -37,17 +95,18 @@ export class DevReadingListPage {
...
 
     async expectNumberOfEntries(numberOfEntries: number) {
         await expect(this.readingListEntries).toHaveCount(numberOfEntries);
     }
+
+    async openReadingList() {
+        const navigationBar = new NavigationBarModel(this.page);
+        await navigationBar.userButton.click();
+        await navigationBar.getReadingListLink(true).click();
+    }
```
mowi12: I don't think this fits here, don't you think?
Shouldn't this rather be a function on the homepage or the navigation bar since you already have the user avatar menu and the reading list link there?

This is only duplicate code and makes it harder to understand which fixture/model to use in order to open something, in this case the reading list.

tests/e2e/paper/paper-view.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        paperViewPage,
+        homePage,
+        projectPaperViewPage,
+        readingListPage,
+    }) => {
+        await page.goto("/");
+
+        await homePage.openProjectPaper(paperViewPage.paperNames[0]);
+        await projectPaperViewPage.getReadingListButton(true).click();
+        await readingListPage.openReadingList();
```
mowi12: As already mentioned in the `reading-list-fixture`, this function should be on the homepage. Update accordingly after refactoring.

tests/e2e/paper/paper-view.test.ts
```diff
@@ -0,0 +1,64 @@
...
+        await readingListPage.readingListEntries.first().click();
+        await expect(paperViewPage.getHeader(paperViewPage.paperNames[0])).toBeVisible();
+    });
+
+    test.fixme(
+        "When the user clicks on the information button, then the paper information is displayed",
+        async ({ paperViewPage }) => {
+            // TODO implement that the information and document button are disabled as long as the page is loading
```
mowi12: What's going on here? Are you unable to implement this, or what's going on?

tests/e2e/project/papers/project-papers-page-fixtures.ts
```diff
@@ -0,0 +1,101 @@
...
+    homePage: HomePageModel;
+    projectDashboard: ProjectDashboardModel;
+    navigationBar: NavigationBarModel;
+};
+
+const TOTAL_PAPERS = NUM_PAPERS_PER_STAGE * 2;
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - project papers page
```
mowi12, line 26: Please add the missing object models as well.

tests/e2e/project/papers/papers-overview.test.ts
```diff
@@ -1,66 +1,28 @@
...
+        page,
+        projectPapersPage,
+        homePage,
+        navigationBar,
+    }) => {
+        // Directly navigate to the papers overview
+        await page.goto("/");
+        await homePage.openProject(projectPapersPage.projectName);
+        await navigationBar.papersTab.click();
+        await expect(projectPapersPage.showFiltersButton).toBeVisible();
```
mowi12, line 17: Any reason you chose the filter button?

tests/e2e/project/settings/general/project-general-settings-fixture.ts
```diff
@@ -1,20 +1,55 @@
...
+    projectSettingsPage: ProjectGeneralSettingsPageModel;
+    homePage: HomePageModel;
+    projectMemberSettingsPage: ProjectMemberSettingsPageModel;
+    navigationBar: NavigationBarModel;
+    projectSettingsSideBar: ProjectSettingsSidebarModel;
 };
 
 /**
  * Extends the default **custom** fixture by providing the page object models for the
  * - project settings page
```
mowi12: Again, please add the missing object models.

tests/e2e/project/settings/general/project-general-settings-model.ts
```diff
@@ -1,14 +1,28 @@
...
         this.projectNameInput = page.getByLabel("Project Name");
         this.renameButton = page.getByRole("button", { name: "Rename" });
+
+        this.projectId = "";
+        this.projectName = "Project 1";
+    }
+
+    getHeading() {
+        return this.page.getByRole("heading", { name: this.projectName });
+    }
```
mowi12: If I remember correctly, in other models, you had a `readonly` variable at the top instead of a function to check the heading. Maybe also do this here.

tests/e2e/project/settings/general/project-general-settings-model.ts
```diff
@@ -19,6 +33,7 @@ export class DevProjectSettingsPage {
     async changeProjectName(projectName: string) {
         await this.projectNameInput.fill(projectName);
         await this.renameButton.click();
+        this.projectName = projectName;
     }
 
     async checkForErrors() {
```
mowi12: I know this function is not from you, but in all other tests you also had a `readonly` variable at the top instead of a function to check. Either update all other tests or just this one.

tests/e2e/project/settings/general/project-general-settings-model.ts
```diff
@@ -1,14 +1,28 @@
...
+        this.projectName = "Project 1";
+    }
+
+    getHeading() {
+        return this.page.getByRole("heading", { name: this.projectName });
+    }
+
+    async checkHeading(heading: string) {
+        await expect(this.page.getByRole("heading", { name: heading })).toBeVisible();
     }
```
mowi12: Please add a short documentation to this method so it is clear, what this method is doing in comparison to the `getHeading` function.
Even if the `getHeading` function gets removed, this is still important.

tests/e2e/project/settings/general/project-name-settings.test.ts
```diff
@@ -1,21 +1,34 @@
 import { test } from "./project-general-settings-fixture";
 import { expect } from "@playwright/test";
 
-export let projectId: string = "";
-const projectName = "Project 1";
-
 test.describe("Renaming a project", () => {
```
mowi12: Title Case?

mowi12 (changes requested):
First of all, nice work! There is a lot of stuff going on here and much is looking really good! But of course, there are changes which we need to talk about.

I looked at the first 22 files (they are from the first 4 or 5 commits.
I tried to really focus on how you changed some parts up and decided to give you feedback now before I look at all the other files, because there are some points you probably need to work into almost all files.

1. **General Wording of Navigation Tests**

As you probably know, there are two variants for the `test` function in order to match the description of the test. The description should make sense with and without the `test` in front of it.

`test("When navigating to general settings, the page is displayed", ...` would be a way to do it, as an example.

2. **Failed API Calls**

Some tests are simply missing the failed API call test. I think for the first 22 files I mentioned it in the files, but for the rest you need to check on your own. I will try to verify it in the end.

3. **Title Case in all `describe`-functions**

This one is not as important, but Imo this should be consistent.
If you rather think they should also be sentences, then let me know and update all title case `describe`-functions to sentences.

4. **General Naming of Files**

In generall this is allready pretty consistent, but some files (reading-list e.g.) still use the word "page" in their name. Imo wither all fixtures/model/tests should have the same wording and include the word "page" in their name to make clear its about a page (if it is about a page) and not include it when its not about a page. Or remove the word "page" from all files names.
This also applies to the actual object names in the files of the fixtures/models.

5. **Test Names**

Some test names are split with a "+", some are not.
Could you try to make this consistent and either split all non-splitted lines after, Idk, 120 character? Or just remove the "+" from the few that are split.

Edit:
6. **Consistent homepage/navigationBarModel Usage**

As mentioned in some files, you use the variables/functions of the homepage model and navigation bar inconsistent.
Try to either use always the functions you already have or remove them completely.

And just to get your hopes up again: I really like the code style you know use!
There is not a single code snippet which needs refactoring or else (imo :D)

tests/e2e/navigation-bar-model.ts
domienderle: That is a good call. I should have done this before, but I recognized it too late.

tests/e2e/navigation-bar-model.ts
domienderle: You are right. This is a good suggestion. Even though I would name it `getUserAvatarButton`, because I named every tooltip that it handled like this as a button and naively I would expect the `getUserAvatar` method not to return a button. So to keep it consistent with the rest of the namings I would stay with `getUserAvatarButton`.

tests/e2e/navigation-bar-model.ts
domienderle: I only need the `getUserAvatarButton` in the `change-user-name-settings.test.ts`, to make sure, that the `userAvatarButton` really displays the right credentials. In every other test it doesn't matter what credentials this button has, because it was already tested that they shall be correct. But since both reference to the same element, you are right. It can be refactored in one method.

tests/e2e/navigation-bar-model.ts
domienderle: I know what you mean. I have know refactored the navigation bar to be split into the normal `NavigationBar` with the `userAavatarButton` and the `goBackButton`, and the `ProjectNavigationBar` with the additional tabs. Contextually it should make more sense now, because you can/have to use the correct navigation bar now in the correct context.

tests/e2e/homepage/home-page-model.ts
domienderle: Good catch. We don't need it. I was just not deleted.

tests/e2e/homepage/home-page-model.ts
domienderle: I am not sure. We always your `heading` on all the other pages as well, even though there are also fixed headings sometimes. My thought about this was to keep the naming consitent and to ensure that it is the correct, the heading it is part of the model. If you use the ``HomePageModel` it should be clear which heading is used, if you know what I mean. We can change the name of the heading, but then we have to do it every time we have a fixed heading and I am not a fan about this, because imo this wouldn't be really consistent anymore.

tests/e2e/auth/signup/sign-up-page-fixtures.ts
domienderle: Good catch. I updated all the comments.

tests/e2e/auth/signup/signup.test.ts
domienderle: Done. I also added it to the `/routes/(auth)/signin/+page.svelte`.

tests/e2e/auth/signout/signout-fixture.ts
domienderle: Should be done for each fixture now.

tests/e2e/navigation-bar-model.ts
domienderle: Exactly. I also thought about this. But since now have the models two handle all the pages, dialogs and bars it is really easy to just "click" through the website as a user would do. Due to that, I just return the elements to simulate "perfect" user behaviour.

tests/e2e/auth/signin/signin.test.ts
domienderle: You are right. I didn't catch this. Good call.

tests/e2e/paper/paper-view.test.ts
domienderle: I renamed all the `describe`-blocks, so this is implicitly done.

tests/e2e/auth/signin/signin.test.ts
domienderle: Yes you are right. I will do this as soon as your PR is done.

tests/e2e/readinglist/reading-list-page-model.ts
domienderle: You are right. This is kind of confusing. I have now removed the whole method, because in any other case, we also do both of the calls in the test, which is imo the correct way to really make the user click simultation visible in the tests.

tests/e2e/paper/paper-view.test.ts
domienderle: I don't think we need this implementation anymore. I was not sure in the beginning, but since this is a tab and not a button, which doesn't depend crucially on the loading data, it should not be necessary. What do you think?

tests/e2e/project/papers/papers-overview.test.ts
domienderle: Not really, I just wanted to make sure that at least one object of the specific side is shown. If I would check on the heading, it could also be another side. But for this check, we could also check on any other specific element of this side.

tests/e2e/project/settings/general/project-general-settings-model.ts
domienderle: You are right. With the `checkHeading` we can just create a `readonly` attribute. Good catch.

tests/e2e/project/settings/general/project-general-settings-model.ts
domienderle: Good catch. Is done.

tests/e2e/project/settings/general/project-name-settings.test.ts
domienderle: I updated all titles

domienderle:
I tried to include all your reviews and do it for all the files. I tried to fix all the comments and also add comments to all methods that were still uncommented. You can have a look on the code again and give me some feedback on it.

tests/e2e/navigation-bar-model.ts
mowi12: I like it!

tests/e2e/navigation-bar-model.ts
mowi12: Perfect!

tests/e2e/navigation-bar-model.ts
```diff
@@ -0,0 +1,88 @@
...
+    constructor(page: Page) {
+        this.page = page;
+        this.goBackButton = this.page.getByRole("link", { name: "Back to" });
+    }
+
+    /**
+     * Returns the user avatar button.
+     *
+     * @param initials - if the credentials are given it only returns the user avatar button if the credentials of the button are correct
+     */
```
mowi12:
```suggestion
/**
 * Returns the user avatar button based on the provided initials.
 *
 * If no initials are provided, it returns a button whose name consists of two uppercase letters.
 * If initials are provided, it returns the button that exactly matches those initials.
 *
 * @param initials - The expected initials on the avatar button (optional).
 */
```

mowi12: Same applies to the `project-navigation-bar-model.ts`.

tests/e2e/navigation-bar-model.ts
```diff
@@ -0,0 +1,88 @@
...
+    getUserAvatarButton(initials: string = "") {
+        if (initials === "") {
+            return this.page.getByRole("button", { name: /^[A-Z]{2}$/ });
+        }
+        return this.page.getByRole("button", { name: initials, exact: true });
+    }
```
mowi12:
```suggestion
    getUserAvatarButton(initials?: string) {
        if (!initials) {
            return this.page.getByRole("button", { name: /^[A-Z]{2}$/ });
        }
        return this.page.getByRole("button", { name: initials, exact: true });
    }
```

mowi12: Same applies to the `project-navigation-bar-model.ts`.

tests/e2e/homepage/home-page-model.ts
mowi12: True, I have not thought about the fact, that we would need to change all the variable names from `heading` to the correct variant. Then let's keep it as it is now.

tests/e2e/auth/signup/signup-page.test.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from "@playwright/test";
+import { test } from "./sign-up-page-fixtures";
+
+test.use({ user: null });
+
+test.describe("Sign Up Tests", () => {
+    test("When navigation to the sign up page, then the page is displayed", async ({
```
mowi12:
```suggestion
    test("When navigating to the sign up page, then the page is displayed", async ({
```

tests/e2e/auth/signup/signup.test.ts
mowi12: As discussed, we won't test failed API calls.

mowi12: As discussed, look at the titles of the previous errors in this file and match the heading, thanks!

tests/e2e/auth/signin/signin.test.ts
mowi12: My PR is merged now, so this should be testable now.

tests/e2e/auth/signin/signin.test.ts
mowi12: Perfect!

tests/e2e/auth/signin/signin.test.ts
mowi12: As discussed, doesn't need to be done.

tests/e2e/project/papers/project-papers-page-model.ts
```diff
@@ -42,6 +69,10 @@ export class DevProjectPapersPage {
         return this.page.getByRole("button", { name: title });
     }
 
+    getStageButton(stageIndex: number): Locator {
+        return this.page.getByRole("button", { name: `Stage ${stageIndex}` });
+    }
```
mowi12, line 75: Please add a simply comment similar to the others in this file just to keep it consistent.

tests/e2e/project/papers/papers-overview.test.ts
mowi12: Ok

tests/e2e/project/settings/general/project-general-settings-page-model.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    constructor(page: Page) {
+        this.page = page;
+        this.projectId = "";
+        this.projectName = "Project 1";
+
+        this.heading = page.getByRole("heading", { name: this.projectName });
+        this.projectNameInput = page.getByLabel("Project Name");
+        this.renameButton = page.getByRole("button", { name: "Rename" });
+        this.errorAlter = page.getByRole("alert");
```
mowi12:
```suggestion
        this.errorAlert = page.getByRole("alert");
```

tests/e2e/project/settings/general/project-general-settings-page-model.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /**
+     * Verifies that a heading with the specified name is visible on the page.
+     *
+     * @param heading - The text of the heading to check for visibility.
+     */
+    async checkHeading(heading: string) {
+        await expect(this.page.getByRole("heading", { name: heading })).toBeVisible();
+    }
```
mowi12: As discussed, this method is not needed as we can simply check if the `heading` variable at the top is visible since the project name is the heading.

tests/e2e/project/settings/review/project-review-settings-page-model.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Returns the delete button associated with the specified tag.
+     *
+     * @param tagName - The name of the tag for which the delete button is to be fetched.
+     */
+    async getTagDeleteButton(tagName: string) {
+        return this.getTag(tagName).then((tag) => tag.getByRole("button", { name: "×" }));
+    }
```
mowi12: Do we actually need this in a separate method? It's only used by the method after this one.
Imo a method for only one line of code which is only used by one method is not necessary. In this case e.g. we add about 7-8 lines of code for one line of code.

I suggest just putting this one line of code inside the other method.

tests/e2e/project/settings/slr/project-slr-settings-page.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        // Directly navigate to the project slr settings
+        await page.goto("/");
+
+        await homePage.openProject(projectSLRSettingsPage.projectName);
+        await projectNavigationBar.settingsTab.click();
+        await projectSettingsSideBar.slr.click();
+        await expect(projectSLRSettingsPage.maybeAsDecisionSwitch).toBeVisible();
+    });
+
+    // --- Maybe as Decision ---
```
mowi12: This comment can be removed since each describe block will test one settings section.

tests/e2e/readinglist/reading-list-page-model.ts
```diff
@@ -4,23 +4,83 @@ export const PREDICTABLE_PAPER_TITLE_PREFIX = "Test Paper";
...
+     * @param index - The index of the entry to check for visibility.
+     */
+    async checkEntryVisible(index: number) {
+        const entry = this.getEntryByIndex(index);
+        await entry.scrollIntoViewIfNeeded();
+        await expect(entry).toBeVisible();
+    }
+
+    /**
+     * WARNING: To use this method, the paper must already be part of the reading list.
```
mowi12: Please replace 'WRNING' with `@remarks` to use the provided annotations.

tests/e2e/settings/account/account-settings-page.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        // Navigate to the account settings page via the shortcuts settings page
+        await page.goto("/");
+
+        await navigationBar.getUserAvatarButton().click();
+        await navigationBar.getSettingsLink().click();
+        await settingsSideBar.shortcuts.click();
+        await expect(shortcutsSettingsPage.shortcutsVisibilitySwitch).toBeVisible();
+
+        await settingsSideBar.account.click();
+        await expect(accountSettingsPage.firstNameInput).toBeVisible();
```
mowi12, line 28: Why do we need this part of the test. If yes, why?

tests/e2e/settings/shortcuts/shortcuts-settings-page-model.ts
mowi12, line 1: What about adding a expect to the `setShortcutsVisibility` function similar to other toggles so the test fails if the toggle is not the expected state.

Look at the `project-slr-settings-page-model.ts`.

mowi12: Same applies for the slr settings toggle. Please add this here to if you think this is a good idea.

tests/e2e/settings/review/review-settings-page-model.ts
mowi12, line 1: Same here with the expect for the toggle.

tests/e2e/project/dashboard/project-dashboard-page-model.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * Returns a heading element based on a specified paper title.
+     *
+     * @param paperTitle - The title of the paper to locate within the navigation role.
+     */
+    getHeading(paperTitle: string) {
+        return this.page.getByRole("navigation").getByText(paperTitle, { exact: false });
+    }
```
mowi12, line 26: Shouldn't this also be a readonly variable at the top, similar to all the other headings checks?
It's again, the project name which is the heading of the dashboard.

And the description and naming of the variables is a bit wrong, isn't it?

If not, why do we get the heading of a paper here and not in the appropriate model?

tests/e2e/archivedprojects/archived-projects-page-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+type ArchivedProjectsPageFixture = {
+    archivedProjectsPage: ArchivedProjectsPageModel;
+    navigationBar: NavigationBarModel;
+};
+
+/**
+ * Extends the default **custom** fixture by providing the page object models for the
+ * - home page
+ * - the dialog for creating a new project
+ */
```
mowi12: This seems still a bit wrong.

src/routes/(auth)/signin/+page.svelte
mowi12, line 1: Please use the same format as in the error messages before.

src/routes/(auth)/signup/+page.svelte
mowi12, line 1: Please use the same format as in the error messages before.

tests/e2e/project/settings/members/promote-members.test.ts
mowi12, line 1: This file has a lot of commented out code.

What about all this?

Are this already tests that don't have the needed functionality in the mock -backend?

What about also making fixme tests out of them?

tests/e2e/project/paper/project-paper-view-page-model.ts
```diff
@@ -26,16 +39,24 @@ export class DevPaperViewPage {
...
+     * Returns a heading element based on a specified paper title.
      *
-     * @param paperId - The local project paper id
+     * @param paperTitle - The title of the paper to locate within the navigation role.
      */
-    async openPaperView(paperId: string) {
-        await this.page.goto(`/paper/${paperId}`);
-        await expect(this.referenceListEntry0).toBeVisible();
+    getHeading(paperTitle: string) {
+        return this.page.getByRole("navigation").getByText(paperTitle, { exact: false });
```
mowi12: Why does this need to be `exact: false`?

Shouldn't the name be known of the paper and therefore be exact?

tests/e2e/project/paper/project-paper-view-page-model.ts
mowi12, line 1: I know, i know ... but there are. two functions without documentation. Please add the two final comments : )

wiki/Testing.md
mowi12, line 1: Shouldn't we also inlcude the isolated fixture somewhere?

wiki/Testing.md
```diff
@@ -155,4 +157,22 @@ requiring it to be installed and running.
...
+   to access any element must be declared in the according model to ensure reusability. Only use functions that
+   click multiple elements if they are really needed. In general, the intention is to simulate the user behavior
+   as good as possible by simulating each click in the tests.
+
+   The name of the model is defined by the following guidelines:
+
+   - `custom-name-page-model.ts` (if the model represents a full page)
+   - `custom-name-navigation-bar-model.ts` (if the model represents a navigation bar)
+   - `custom-name-sidebar-model.ts` (if the model represents a sidebar)
+   - `custom-name-dialog-model.ts` (if the model represents a dialog)
```
mowi12, line 178: Maybe also inlcude the naming for the fixture and model.

mowi12 (changes requested):
First of all, great work! I really like your changes and you refactored everything as discussed. This time there are less changes than before, but still changes.

1. `@return` Parameter
We discussed this point. I don't know if this is needed, since there is not a single testing method which does this. As discussed, you can not do this. Just ignore this point.

2. TODOs
Maybe replace TODOs with the `fixme` version of a test. This makes it clear that there are tests, which are not finished. Then we don't have to search for the keyword `todo`, but already see this, when we run the tests.

3. Describe blocks instead of mutliple Files
This will be discussed in our sprint meeting. I will add a comment as soon as this is discussed.
(e.g. `keyword-settings.test.ts` move inside the `project-review-settings-page.test.ts`)

4. Where to put Navigation test
At the start of the first desribe bock?
Or, since we will use more than one block either way, we can simply use a single one which would be the best solution.

tests/e2e/navigation-bar-model.ts
domienderle: Good call. Since the methods in both classes are exactly the same the overriding method in `project-navigation-bar-model.ts` is unnecessary, which is why I removed it.

tests/e2e/navigation-bar-model.ts
domienderle: As I said, I removed it.

tests/e2e/auth/signup/signup-page.test.ts
domienderle: Good catch.

tests/e2e/project/settings/general/project-general-settings-page-model.ts
domienderle: Good catch.

tests/e2e/project/settings/general/project-general-settings-page-model.ts
domienderle: All right. It doesn't work as we expected it to work. Since the heading is defined in the constructor, it always searches for the project with the name `Project 1`. If we would like to dynamically referenciate on locators we must use methods. Neveretheless, you had a good point. Since the tests for the general project settings test a couple of test cases with wrong input (so the project name doesn't change), we can only check on the heading there. Since the only case in which the heading is different is the case, when the inputs are correct, I added a manual check for the new heading there. 

Have a look on it. I think this should be a good solution.

tests/e2e/project/settings/review/project-review-settings-page-model.ts
domienderle: Valid point. When I implemented this method I thought I would have to use it more often. But you are right. At the moment it doesn't make sense.

tests/e2e/settings/account/account-settings-page.test.ts
domienderle: The thought about this was the following: The plan should be to test all possible navigations to this page. The first route to this page is by just clicking on the `userAvatarButton `and on `settings` afterwards. Now you are directly on the `account-settings-page`. But another way to get there would be to just navigate from another user setting to the account settings. But since we get directly routed to this page on click on settings, we have to navigate to another setting first before we navigate back. It is not the pretties solution, since we don't have this case for the other settings, but imo it would make sense for the sake of complenteness.
But if you consider this as unnecessary, let me know.

tests/e2e/settings/shortcuts/shortcuts-settings-page-model.ts
domienderle: Good idea. Done.

tests/e2e/project/dashboard/project-dashboard-page-model.ts
domienderle: Yes, you are right.

tests/e2e/archivedprojects/archived-projects-page-fixture.ts
domienderle: Right. Must have forgotten this.

tests/e2e/project/settings/members/promote-members.test.ts
domienderle: This test are not testable at the moment, because it is not possible to promote an invite member. Nevertheless, you are right. We can change them to fixme. I changed them and adjusted them.

tests/e2e/project/paper/project-paper-view-page-model.ts
domienderle: True.

wiki/Testing.md
domienderle: Good point.

wiki/Testing.md
domienderle: Yes, good catch.

domienderle:
All right. I tried to include everything. I rebased the most things in the single files, but the refactoring for all files like the `TODOs` or the `describe` blocks are in extra commits. The last commit for the `project-member-settings-page.test.ts` should have also been rebased, but every time I rebased my changes into the last commits my changes got discarded, so I just commited it extra.

tests/e2e/auth/signup/signup.test.ts
mowi12: I should have been more specific the first time. What about

```
registrationError = {
                        errorTitle: "Registration Failed",
                        errorDetails:
                            "Something went wrong during registration. Please make sure your internet connection is stable, then try again.",
                    };
```

for the error in signup. And what about

```
signinError = {
                        errorTitle: "Sign-In Failed",
                        errorDetails:
                            "Something went wrong during sign-in. Please make sure your internet connection is stable, then try again.",
                    };
```

for the sign-in page.

tests/e2e/paper/paper-view.test.ts
mowi12: I think you can remove the TODO. But keep the test itself. Also remove the fixme, so the test is actually counting.
I think we should keep it, since there are two different views, and depending on which tab we are, the user should see different things.

tests/e2e/navigation-bar-model.ts
mowi12: True. But in this case, all methods can be removed, or not?

tests/e2e/navigation-bar-model.ts
mowi12: Yes, you removed the code in the `project-navigation-bar-model.ts` file, but in the normal one, this code still exists. And I think, the `?` is the better variant.

tests/e2e/project/settings/general/project-general-settings-page-model.ts
mowi12: Yes, I think this is a good solution.
Generally, if there is only one line of code in a method, how good can this method really be.

The way we test it now is a good compromise by being able to check if the heading is visible via the variable and manually if the name is changing.

tests/e2e/project/settings/review/project-review-settings-page-model.ts
mowi12: So do you want to keep it or what?

I don't think we'll ever need to only get the delete button and not use it.

tests/e2e/readinglist/reading-list-page-model.ts
mowi12: I think this needs to be formated another way. Normally there is an empty line between the normal description and the `@remarks` part. And the `@remarks` is in a line above, not in the same line.

tests/e2e/settings/account/account-settings-page.test.ts
mowi12: Okay, sounds good.

tests/e2e/project/dashboard/project-dashboard-page-model.ts
mowi12: I don't think you changed this.

wiki/Testing.md
```diff
@@ -141,12 +141,22 @@ requiring it to be installed and running.
 
 1. **Use the test fixture**:
    Instead of importing the `test` fixture directly from "@playwright/test" directly,
```
mowi12:
```suggestion
   Instead of importing the `test` fixture directly from "@playwright/test",
```

wiki/Testing.md
```diff
@@ -141,12 +141,22 @@ requiring it to be installed and running.
...
-   use the `test` variable exported from [`utils/fixtures/shared-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/tests/e2e/utils/fixtures/shared-fixture.ts).
+   use the `test` variable exported from one of the following:
+
+   - [`utils/fixtures/shared-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/tests/e2e/utils/fixtures/shared-fixture.ts)
+     (a new mock backend is started for each worker)
+   - [`utils/fixtures/isolated-fixture.ts`](https://github.com/SE-UUlm/snowballr-frontend/blob/develop/tests/e2e/utils/fixtures/isolated-fixture.ts)
+     (a new mock backend is started for each test)
+
    This ensures that the API calls to the mock backend are automatically redirected to the correct mock backend
```
mowi12, line 151:
```suggestion
   This ensures that the API calls to the mock backend are automatically redirected to the correct mock backend instance
```

wiki/Testing.md
mowi12: You did not do this, or?

tests/e2e/project/settings/slr/project-slr-settings-page.test.ts
mowi12, line 1: Extract nagivation into own describe block

mowi12 (changes requested):
Nicely done. There are only one or two conversations which you missed.

Then there are two last things, then we are done!

1. Completely remove TODOs
I think you can completely remove the TODO strings above the `fixme` tests. There is no need for the text since the fixme is saying the same.
If the TODO contains something important, e.g. what needs to be implemented in order to create this test, then leave it.

2. `describe`block for navigating
I think there are still one or two files where the navigation is not in a separate describe block. Please check this again. I found one, but I thought there are two.

tests/e2e/navigation-bar-model.ts
domienderle: That is right. Don't know why its not already the case-

tests/e2e/navigation-bar-model.ts
domienderle: That is right. That is a better implementation.

tests/e2e/project/settings/review/project-review-settings-page-model.ts
domienderle: No, I removed it, but I don't know why it was not changed. Should be removed now.

tests/e2e/readinglist/reading-list-page-model.ts
domienderle: I am not sure if I got you, but I think I adjusted it as you said.

tests/e2e/project/dashboard/project-dashboard-page-model.ts
domienderle: Sorry, I forgot to answer. The problem here is, that other tests than the dashboard tests also use the heading to check, if the dashboard page is correctly loaded. Since they maybe use other project names, it makes sense to keep this method imo.

wiki/Testing.md
domienderle: Yes I did. But I added it in the according section. Have a look on the last sentences of the sections for the test classes and the fixtures.

tests/e2e/project/settings/slr/project-slr-settings-page.test.ts
domienderle: Done.

wiki/Testing.md
mowi12: Ah, i missed this.

tests/e2e/readinglist/reading-list-page-model.ts
mowi12: Yes, perfect!

tests/e2e/project/dashboard/project-dashboard-page-model.ts
mowi12: True. I looked at the code and this looks plausible.

mowi12 (approved):
Perfect. This went pretty fast for the amount of conversations we had. Now the e2e tests are finally nice to look at : )

PR #451, opened 2025-06-20
Added option to sort project papers

### LINKED ISSUE #39: [Use Case]: Sort Papers

Merseleo:
Closes #39 

## What I have made

- I added a new button on the project papers overview page so the user can select the sorting option
- Accordingly to the selected sort option, the paper are sorted within the stages; currently the sorting options "Paper Id", "Paper title", "Year" and "Decision" are supported
- When the sort options changes, the new sort option is persisted in the URL and loaded from the URL as well

In general, most of this functionality can be applied to other pages or lists as well, as long as they deal with (project papers).

## Checklist

  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests


src/lib/components/composites/select/SortOptionsSelect.svelte
```diff
@@ -0,0 +1,60 @@
...
+        SortDirection,
+        type SortOptionLabel,
+    } from "$lib/model/sortCriteria";
+    import {
+        ArrowDown,
+        ArrowDown10,
+        ArrowDownZA,
+        ArrowUp,
+        ArrowUp10,
+        ArrowUpAZ,
```
luca-schlecker: The current icon selection is a bit confusing imo.
Sometimes the arrow changes direction, sometimes the decoration beside it.

What do you think about having it always pointing downwards and just swapping the order of the decorations?

| Id: low to high | Id: high to low | Title: low to high | Title: high to low |
| :---: | :---: | :---: | :---: |
| image | image |  image | image |

src/lib/components/composites/select/SortOptionsSelect.svelte
```diff
@@ -0,0 +1,60 @@
...
+                <ArrowUp10 class="size-4" />
+            {/if}
+        {:else if ALLOWED_SORT_OPTIONS[selectedSortOption].criterion === SortCriteria.PAPER_TITLE}
+            <ArrowDownZA class="size-4" />
+        {:else if ALLOWED_SORT_OPTIONS[selectedSortOption].criterion === SortCriteria.DECISION}
+            <ArrowDown class="size-4" />
+        {:else}
+            <ArrowDown10 class="size-4" />
+        {/if}
+    </Select.Trigger>
```
luca-schlecker: The trigger seems to have a slight border:
image

src/lib/model/sortCriteria.ts
luca-schlecker, line 1: Files should be `kebab-case`.

src/lib/model/sortCriteria.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "Year: Oldest to Newest": {
+        criterion: SortCriteria.YEAR,
+        direction: SortDirection.ASC,
+    },
+    "Year: Newest to Oldest": {
+        criterion: SortCriteria.YEAR,
+        direction: SortDirection.DESC,
+    },
+};
```
luca-schlecker: Is there a reason why this is in place to potentially prevent a user from sorting any criterion in the direction they'd like?

I feel like this should be inverted: a function taking a `SortOption` and returning the text currently used as a key.
The text input would then just list every possible combination of `SortCriteria` and `SortDirection`.

Let me know your thought's on this.

src/lib/utils/sorters.ts
```diff
@@ -0,0 +1,65 @@
+import { PaperDecision, type Project_Paper } from "$lib/model/api/project";
+import { SortCriteria, SortDirection, type SortOption } from "$lib/model/sortCriteria";
+
+/**
+ * Compares two paper decisions based on a custom priority order:
+ * ACCEPTED \> IN_REVIEW \> DECLINED \> UNREVIEWED
```
luca-schlecker: I think this line is wrong. To me, it suggests that `ACCEPTED` > `UNREVIEWED`.
Yet, the current implementation orders `ACCEPTED` **<** `UNREVIEWED`, as `UNREVIEWED` has a higher numeric value.

src/lib/utils/sorters.ts
```diff
@@ -0,0 +1,65 @@
...
+            case SortCriteria.YEAR:
+                compareResult = (a.paper?.year ?? 0) - (b.paper?.year ?? 0);
+                break;
+
+            case SortCriteria.DECISION:
+                compareResult = comparePaperDecisions(a.decision, b.decision);
+                break;
+
+            case SortCriteria.PAPER_ID:
+                compareResult = a.localId.localeCompare(b.localId);
```
luca-schlecker: I know that `id` and `localId` currently are strings, however, I believe a user would expect `4` to come before `11` if sorted by id from low to high.

If I recall correctly, currently the plan for local ids is to be only of numerical nature.

I'd suggest trying to parse ids as integers. If both ids of two papers are integers, the lower one is ordered before. If both ids are strings, the current implementation can be used to order them. If, however, one is numerical and one not, the integer value should be ordered before the non-numerical.

tests/unit/utils/search-parameters.test.ts
```diff
@@ -116,3 +115,40 @@ describe("Helper filters URL query parameters", () => {
...
 });
+
+describe("Helper sort URL query parameter", () => {
+    test("When the query parameter for the sort option is present, then it will be successfully read and returned.", async () => {
+        page.url = new URL("http://localhost/test?sort=Title&order=desc");
+
+        expect(getSortOptionFromURL()).toBe("Title: Z to A");
+    });
+
+    test("When the query parameter for the sort option is not present, then undefined will be returned.", async () => {
```
luca-schlecker: I cannot see how "undefined" plays any role in this test?
This should probably state that the default is expected.

luca-schlecker (changes requested):
Great work! Works well, is documented and has tests.
There are a few minor questions and issues, though.
I am looking forward to hearing your opinion on them.

tests/unit/utils/search-parameters.test.ts
Merseleo: Good catch, this is a leftover from the implementation phase. First, I returned undefined and checked somewhere else for undefined and handled the case, but corrected it to return a default value, if something went wrong. Thank you for the hint.

src/lib/components/composites/select/SortOptionsSelect.svelte
Merseleo: I also tested this solution and in my opinion it is more difficult to recognize the sorting direction when only the decorations change and not the arrow, but actually it doesn't matter to me, so I can change it, if you prefer this solution.

src/lib/components/composites/select/SortOptionsSelect.svelte
Merseleo: You have a real eagle eye, I only realised that the second time I looked at it.

src/lib/components/composites/select/SortOptionsSelect.svelte
Merseleo: Do have any further ideas for the decoration for arrow for indicating in which order the decisions are sorted?

src/lib/model/sortCriteria.ts
Merseleo: You are right

src/lib/model/sortCriteria.ts
Merseleo: How does this prevent a user from sorting any criterion? I think I do not quite understand what you want to tell me. Can you elaborate?

src/lib/utils/sorters.ts
Merseleo: You are right, but if you look at the implementation of the sort function, lower values are sorted before higher values, so if you want to sort something ascending, things with a _higher_ priority should have _lower_ numeric values. However, if this is confusing in your opinion, if can assign ACCEPTED to 4, unspecified to 0 and invert the return statement: 
```typescript
function comparePaperDecisions(a: PaperDecision, b: PaperDecision): number {
    const priority: Record<PaperDecision, number> = {
        [PaperDecision.UNSPECIFIED]: 0,
        [PaperDecision.UNREVIEWED]: 1,
        [PaperDecision.DECLINED]: 2,
        [PaperDecision.IN_REVIEW]: 3,
        [PaperDecision.ACCEPTED]: 4,
    };

    // subtract the priority of b from the priority of a as if a has the higher priority the return value must be negative  for the sorting function
    return priority[b] - priority[a];
}
```

src/lib/utils/sorters.ts
Merseleo: Good catch, I will implement your suggestion or even better I just found a function from @Slartibartfass2 that already compares two paper ids. Are you ok with the implementation of this function?

Merseleo (commented):
Thank you for the feedback and good ideas. The review was very attentive and I didn't notice some mistakes on my part, very good 👍

src/lib/components/composites/select/SortOptionsSelect.svelte
luca-schlecker: I'd prefer either my solution or where the order of the decorations is fixed and only the arrow changes.

src/lib/model/sortCriteria.ts
luca-schlecker: You currently maintain a list of all possible combinations the user is allowed to sort by (in `ALLOWED_SORT_OPTIONS`). IMO this is not necessary as all possible combinations are valid.

Maybe the dropdown could look something like this?

image

This is only a rough sketch though, it should just demonstrate how a composition could look like. I am looking forward to your opinion on this.

src/lib/utils/sorters.ts
luca-schlecker: No no, I agree with your decision to do it this way around with `ACCEPTED` having a value of `0`. Only the comment is misleading imo. Swap the `>` signs:
```
ACCEPTED < IN_REVIEW < DECLINED < UNREVIEWED
```

`ACCEPTED < IN_REVIEW` as $0 - 1 = -1 < 0$

src/lib/utils/sorters.ts
luca-schlecker: I agree with the implementation. Failing to parse a numerical id will not sort it by lexicographical order, though.

I'll leave it up to you whether non-numerical ids are seen as equal or should be compared using lexicographical order. This function would implement the ordering I suggested, but you can modify it however you like:
```ts
function comparePaperId(a: string, b: string): number {
    const idA = parseInt(a.slice(1), 10);
    const idB = parseInt(b.slice(1), 10);

    const aNan = isNaN(idA);
    const bNan = isNaN(idB);

    if (!aNan && !bNan) return idA - idB;
    else if (aNan && bNan) return a.localCompare(b);
    else if (aNan && !bNan) return 1;
    else if (!aNan && bNan) return -1;
}
```

luca-schlecker (changes requested):

src/lib/components/composites/select/SortOptionsSelect.svelte
Merseleo: Ok, I always used an arrow pointing downwards, what do you think?

src/lib/model/sortCriteria.ts
Merseleo: I did some further research and in my opinion this would be very unusual to have two buttons, but way more usual to have one select with all options.

src/lib/utils/sorters.ts
Merseleo: Ok, I adapted the comment. What do you think of it?

src/lib/utils/sorters.ts
Merseleo: Ok I though a while about your suggestion and did not found anything better.

src/lib/model/sortCriteria.ts
Merseleo: Furthermore I tried to implement your idea with the function and in my opinion this would complicate everything, especially when we want it type safe and we would loose flexibility when we do not want symmetric sort options.

Merseleo (commented):
I still have struggles to understand your ideas correctly, because in my opinion they would introduce further complexity and worsen user experience.

src/lib/components/composites/select/SortOptionsSelect.svelte
luca-schlecker: Way better. :+1:

src/lib/model/sortCriteria.ts
luca-schlecker: Ok. We'll leave it like this then. Asymmetric options seem rather weird, but I guess it would be more complex then. :+1:

src/lib/utils/sorters.ts
luca-schlecker: Better now. :+1:

luca-schlecker (dismissed):
Everything seems to be addressed. Great work :+1:.
After a rebase, this can be merged. :rocket:

Merseleo:
I also asked Alex for his opinion on the design: He said that a select component in which every option is described in detail is probably easier to understand, especially with the _Decision_ option.

luca-schlecker (approved):
Great work, LGTM :rocket:

PR #455, opened 2025-06-25
Removed noop catch statements

### LINKED ISSUE #452: [Feature]: Remove noop-catch

_No response_

Now that we use SvelteKit's fetch we can remove the noop-catch statements and with this clean up the code a bit.

For reference see [SvelteKit docs](https://svelte.dev/docs/kit/load#Streaming-with-promises).

Remove all noop-catch statements.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #452 

## What I have made

Since we use SvelteKit's `fetch` function, we don't need the noop-catch statements anymore. I removed them.

Waits for #454 to see the decrease in findings.

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (not required)

### DIFF
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -58,9 +58,6 @@ export const load: PageLoad = async ({ depends, parent }) => {
 
     const allUserProjects = backendService.getAllProjectsForUser({ id: userId }).response;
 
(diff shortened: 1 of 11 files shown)


Merseleo (approved):
You are right, there are useless if we use the `fetch` function from SvelteKit

PR #463, opened 2025-07-09
fix: correctly show loading spinner for navigation buttons

### LINKED ISSUE #462: [Bug]: Flickering review buttons

#52

### What happened?

If one of the decision buttons is clicked, then the buttons disappear until the data for the next paper is loaded. At the moment the buttonsseem to flicker because of the short disappearance.

### Steps to reproduce

1. Go to the paper view of a project paper
2. Click on one of the review decision buttons

### What did you expect?

The expected behaviour would be, to show the loading spinner as long as the current paper is shown and the disabled buttons as long as the paper data is not fully loaded on the next papers page. 

### Anything else?

_No response_

domienderle:
Closes 

## What I have made

- I removed the `await`-statement in the `PaperView.svelte`, which causes the flickering of the review decision buttons
- I have added a global state for the buttons in the project paper view, which handles the loading of the ressources for all five buttons at once. 
- Additionally, I fixed a couple of bugs regardings the loading spinner of the paper navigation buttons and the review decision buttons.

## Checklist

  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
@@ -14,6 +14,7 @@
     import { shortcut, type ShortcutEventDetail, type ShortcutTrigger } from "@svelte-put/shortcut";
     import { navigatePaper } from "$lib/utils/paper-navigation";
     import type { Project, Project_Paper } from "$lib/model/api/project";
+    import { projectPaperLoading } from "$lib/global-state/project-paper-loading-state.svelte";
 
     interface ButtonContent {
         name: string;
@@ -22,25 +23,21 @@
     }
 
     interface PaperDecisionButtonProps {
-        projectPaperId: string;
         variant: PaperDecisionButtonVariant;
         isSubmittingReview?: { value: boolean };
         userReview?: Review;
         loadingProject?: Promise<Project>;
-        loading?: boolean;
         loadingProjectPaper: Promise<Project_Paper | undefined>;
         paperQueue?: Project_Paper[];
         nextProjectPaper?: Project_Paper;
     }
 
     let {
-        projectPaperId,
         variant,
         isSubmittingReview = $bindable({ value: false }),
         userReview = $bindable(undefined),
         loadingProjectPaper,
         loadingProject,
-        loading = $bindable(false),
         paperQueue = $bindable([]),
         nextProjectPaper = $bindable(undefined),
     }: PaperDecisionButtonProps = $props();
@@ -54,7 +51,6 @@
         // the API call is made. However we need a different sate for displaying the loading state of the single
         // decision button. Otherwise, all would have a spinner and label with "Submitting Review".
         isSubmittingReview.value = showLoadingSpinner.value;
-        loading = showLoadingSpinner.value;
     });
 
     /**
@@ -121,8 +117,10 @@
      */
     async function submitReview() {
         try {
+            projectPaperLoading.isLoading = true;
+            const paper = await loadingProjectPaper;
             const review: Review_Create = {
-                projectPaperId: projectPaperId,
+                projectPaperId: paper!.id,
                 decision: getDecision(),
                 selectedCriteriaIds: selectedReviewCriteriaState.criteria,
             };
@@ -140,6 +138,7 @@
                 nextProjectPaper,
                 undefined,
             );
+            if (!nextProjectPaper) projectPaperLoading.isLoading = false;
         } catch (err) {
             toast.error("Could not submit the review!", {
                 description: "Please check your connection to the server.",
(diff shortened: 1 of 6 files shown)


src/lib/global-state/project-paper-loading-state.svelte.ts
mowi12, line 1: This looks perfect, but please a comment, explaining what this stores exactly. Similar to all the other global states.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
```diff
@@ -23,19 +23,23 @@
...
-        loading = showLoadingSpinner.value;
-    });
+    let buttonStillLoading: boolean = $state(true);
+
+    const isDisabled: boolean = $derived(
+        projectPaperLoading.isLoading ||
+            (direction === "right" ? buttonRightDisabled : buttonLeftDisabled) ||
+            buttonStillLoading,
+    );
+    let isLoading: boolean = $derived(projectPaperLoading.isLoading || buttonStillLoading);
```
mowi12: What about:

```suggestion
    const isLoading: boolean = $derived(projectPaperLoading.isLoading || buttonStillLoading);
    const isDisabled: boolean = $derived(
        isLoading || (direction === "right" ? buttonRightDisabled : buttonLeftDisabled),
    );
```

1. You can use `const` for `isLoading` since it's never reassigned.
2. You can reuse the `isLoading` in the evaluation of `isDisabled`, since it uses the same variables to compute the `isDisabled` state.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -22,40 +23,35 @@
...
-    $effect(() => {
+    /*  $effect(() => {
         // Update `isSubmittingReview` every time `showLoadingSpinner` is updated
         // `isSubmittingReview` is used as a state in the `PaperView` component to disable all decision buttons while
         // the API call is made. However we need a different sate for displaying the loading state of the single
         // decision button. Otherwise, all would have a spinner and label with "Submitting Review".
         isSubmittingReview.value = showLoadingSpinner.value;
-        loading = showLoadingSpinner.value;
-    });
+    });*/
```
mowi12: If this is not used any more, why not remove it completely?

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -209,8 +212,11 @@ Usage:
         paperDecisionButtonVariants({ variant: variant }),
     )}
     data-testid={`decision-button-${variant}`}
-    disabled={isSubmittingReview.value || wasAlreadyReviewed || showLoadingSpinner.value || loading}
-    loading={showLoadingSpinner.value}
+    disabled={isSubmittingReview.value ||
+        wasAlreadyReviewed ||
+        showLoadingSpinner.value ||
+        projectPaperLoading.isLoading}
+    loading={projectPaperLoading.isLoading || isSubmittingReview.value}
```
mowi12: This is pretty clunky. What about creating a `isLoading` and `isDisabled` variable at the top, similar to the other files, and only using these variables here?

mowi12 (changes requested):
The logic works and for me, everything looks fine in the UI.
There are only some refactorings you need to look at.

src/lib/global-state/project-paper-loading-state.svelte.ts
domienderle: Good catch. I completely missed this after it worked.

src/lib/components/composites/paper-components/paper-view/PaperNavigationButton.svelte
domienderle: Good catch. Looks good.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
domienderle: I just forgot to remove it.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
domienderle: Good idea. Is done.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
```diff
@@ -22,40 +23,35 @@
...
-        // decision button. Otherwise, all would have a spinner and label with "Submitting Review".
-        isSubmittingReview.value = showLoadingSpinner.value;
-        loading = showLoadingSpinner.value;
-    });
+    const isLoading = $derived(projectPaperLoading.isLoading || isSubmittingReview.value);
+    const isDisabled = $derived(
+        isLoading ||
+            wasAlreadyReviewed ||
+            showLoadingSpinner.value ||
+            projectPaperLoading.isLoading,
```
mowi12: I think this is duplicated, since it is also included in `isLoading`

mowi12 (changes requested):
Nicely done!
Only one last thing.

src/lib/components/composites/paper-components/paper-view/PaperDecisionButton.svelte
domienderle: Shit, you are right. Thanks.

mowi12 (approved):

PR #465, opened 2025-07-13
Incomplete error handling in layoutts fails to redirect on auth errors

### LINKED ISSUE #464: [Bug]: Incomplete error handling in `+layout.ts` fails to redirect on auth errors

_No response_

### What happened?

The root `+layout.ts` load function does not handle all authentication failure scenarios. When the `getAuthenticationStatus` RPC fails (either at the network level or with a non-OK application status), the user is not redirected to `/signin` and is left on a protected page in a broken state.

### Steps to reproduce

1. Start services (db, proxy, backend, frontend)
2. Go to `/` (while Issue 159 in the backend is not merged)
3. Receive wrongly closed call and not get redirected

### What did you expect?

To get redirected.

### Anything else?

_No response_

mowi12:
Closes #464 

## What I have made

The load function is responsible for ensuring that only authenticated users can access protected routes. Any ambiguity or error encountered while checking the authentication status should result in a redirect to the sign-in page.

Currently, there are two unhandled failure cases:

1. RPC/Network Failure: If the `backendService.getAuthenticationStatus` promise is rejected, `authStatusCall` becomes `undefined`. The code logs "Authentication status could not be determined" but does not perform a redirect.
2. Application-Level Error: If the RPC succeeds but the response wrapper contains a non-OK status code (e.g., `authStatusCall.status.code !== StatusCodes.OK`), the code logs an error but also fails to redirect.

Both of these scenarios indicate that a valid user session cannot be confirmed, and the user must be redirected to ensure application integrity and a good user experience. The existing `redirectToSignIn` helper function should be used to handle these cases.

## Checklist

  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests

### DIFF
--- a/src/routes/+layout.ts
+++ b/src/routes/+layout.ts
@@ -34,14 +34,18 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
         () => undefined,
     );
 
+    // Redirect on gRPC/network failure
     if (authStatusCall === undefined) {
-        setCachedUser(null);
-        return { user: emptyUser };
+        console.error("Authentication status could not be determined. Redirecting to sign-in.");
+        return await redirectToSignIn();
     }
 
+    // Redirect on backend business logic failure
     if (authStatusCall.status.code !== StatusCodes.OK) {
-        setCachedUser(null);
-        return { user: emptyUser };
+        console.error(
+            `Authentication status call failed with status: ${authStatusCall.status.code}. Redirecting to sign-in.`,
+        );
+        return await redirectToSignIn();
     }
 
     const authStatus = authStatusCall.response.authenticationStatus;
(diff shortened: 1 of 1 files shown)


domienderle (changes requested):
Looks good so far. The only thing I was thinking about was, if the console ouput should really be an error, or more a warning. What do you think?

mowi12 (commented):
I also thought about that, but I'm not sure how and if we should update the logs.
I checked each of the console outputs and am quite happy with how they are now:

1. If the gRPC/network fails -> error
2. business logic failure -> error
3. Session renewal failed -> error
4. Current user could not be loaded -> error

Each of these are definitely errors and should be logged as one.

domienderle (approved):
All right. I just wanted to hear you opinion about it, because I was not sure either. Then we will keep it as it is, I am also fine with the errors.

PR #466, opened 2025-07-15
Add custom 404 not found page

### LINKED ISSUE #194: [Feature] Add custom 404 Not Found page

_No response_

At the moment we do not really load data from the backend, but hardcoded all objects in the `load` functions. As soon as we start to really load data, something could go wrong, especially that the resource cannot be found.

Therefore you can throw an error using `error(404, { message: "your message" })` which causes svelte to render a custom +error.svelte file. 
However, this file is very basic at the moment and need a redesign.
It should have a design matching the general snowballr design and a speaking error message.

_No response_

### Anything else?

_No response_

mowi12:
Closes #194 
Closes #472 

## What I have made

- Updated the `+error.svelte` file
- Added e2e test

In order to test custom `throw error(405, "message")` error pages, we would need to create a separate file in `routes` in order to test these. As these files would also be visitable in production without explicitly not including them, I decided not to test them.

Since the 404 error page works, other cases should also work.

## Checklist

  - [ ] unit tests
  - [ ] integration tests
  - [x] end-to-end tests


src/routes/+error.svelte
```diff
@@ -1,5 +1,34 @@
...
+                return "Internal Server Error";
+            default:
+                return "Error";
+        }
+    });
+
+    const bodyMessage = $derived(() => {
+        if (page.error?.message === "Not Found") {
+            return "The page you are looking for does not exist.";
+        } else {
```
domienderle: In headTitleMessage you look check for error code 500, in the body message you don't. Maybe add something like this:
```kotlin
const bodyMessage = $derived(() => {
        if (page.error?.message === "Not Found") {
            return "The page you are looking for does not exist.";
        } else if (page.error?.message === "Internal Server Error") {
            return "An internal server error occurred.";
        } else {
            return page.error?.message || "An unexpected error occurred.";
        }
});
```

src/routes/+error.svelte
domienderle, line 1: I have looked into the HTTP error codes. I think we don't need to handle most of the specifically, but maybe the `501 Not implemented` and `502 Bad Gateway` error would be useful to have handled separately. What do you think?

domienderle (changes requested):
Looks good so far. Only two minor things.

src/routes/+error.svelte
mowi12: I refactored the whole file and added some more cases. This should look better now.

src/routes/+error.svelte
mowi12: I added some more placeholders and default messages for all other cases.

src/routes/+error.svelte
```diff
@@ -1,5 +1,81 @@
...
+                break;
+            case 404:
+                title = "Not Found";
+                headline = "Page Not Found";
+                body = "Sorry, the page you are looking for does not exist.";
+                break;
+            case 500:
+                title = "Server Error";
+                headline = "Internal Server Error";
+                body = "Something went wrong on our end. Our team has been notified.";
```
domienderle: I am not sure, if this is a proper message. It is not even true that our team has been notified right? I would propose something like this:
```suggestion
                body = "A server-side error occurred. Please retry or contact support."
```
or
```suggestion
                body ="The server encountered an error and could not complete your request."
```
But I guess this is a matter of taste. We can discuss about those formulations.

tests/e2e/error/error-page-fixture.ts
```diff
@@ -0,0 +1,23 @@
...
+import { HomePageModel } from "../homepage/home-page-model";
+import { ErrorPageModel } from "./error-page-model";
+
+type ErrorPageFixtures = {
+    errorPage: ErrorPageModel;
+    homePage: HomePageModel;
+};
+
+/**
+ * Extends the dafault **custom** fixture by providing the page object models for the
```
domienderle:
```suggestion
 * Extends the default **custom** fixture by providing the page object models for the
```

domienderle (changes requested):
I really like how you handle the error codes now. It is a bit unlucky, that we can only test the 404 error, but since I don't know how we would be able to test the rest in e2e tests, I guess it is ok for now. 

Two little things are left, then we are good to go.

domienderle:
All right looks good now. As soon as the faild e2e run correctly we can merge this.

domienderle (approved):

PR #469, opened 2025-07-16
Added more robust wiki auto-gen command

### LINKED ISSUE #468: [Feature]: Use more robust wiki bash command

_No response_

In https://github.com/SE-UUlm/snowballr-backend/issues/165, we noticed that the wiki command produces a wrong output if `\n` is used in the wiki markdown files. While this repo doesn't have these problems yet. We should also use the better command to prevent such issues in the future.

Same as in https://github.com/SE-UUlm/snowballr-backend/issues/165.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #468 

## What I have made

Same as in https://github.com/SE-UUlm/snowballr-backend/pull/166

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no tests required; execute the bash script to see that nothing has changed)


Merseleo (approved):
Preventive maintenance ;)

PR #472, opened 2025-07-16
Improve +layout.ts

### LINKED ISSUE #471: [Refactor]: Improve Initial Load

#194

Currently, the initial `+layout.ts` file handles the authentication process quite well. But there are some cases in which we want to redirect to the error page or need to error log.

A better and overhauled `+layout.ts` file.

_No response_

### Anything else?

_No response_

mowi12:
Closes #471 

## What I have made

- Moved usage of cached user after authentication check

The initial idea was to refactor the whole file and to use `redirect()` and `throw error()` to show custom error page. But since we do not use ssr or `layout.server.ts` files, we cannot use these methods inside the `load` function. Therefore no real refactoring is possible.

This means, we will always redirect via `goto()` to the sign in page.

In the future we can think about a custom error page, no default `+error.svelte` file. And redirect to this path.
But as this wasn't the focus of this issue, I didn't do this here as this is not important.

The small refactorings that are left, are just some small improvements I found during experimenting.

## Checklist

  - [ ] ~~unit tests~~
  - [ ] ~~integration tests~~
  - [ ] ~~end-to-end tests~~

### DIFF
--- a/src/routes/+layout.ts
+++ b/src/routes/+layout.ts
@@ -9,24 +9,16 @@ import { getCachedUser, setCachedUser, USER_DEPENDENCY_KEY } from "$lib/current-
 
 export const ssr = false;
 
-const emptyUser: User = null as unknown as User;
+const PUBLIC_PATHS = ["/signin", "/signup", "/resetpassword"];
 
 export const load: LayoutLoad = async ({ depends, url, fetch }) => {
     depends(USER_DEPENDENCY_KEY);
     setFetch(fetch);
 
-    // If the user is on a path that does not require authentication, we return an empty user
-    const uncheckedPaths = ["/signin", "/signup", "/resetpassword"];
-    const onUncheckedPath = uncheckedPaths.includes(url.pathname);
-    if (onUncheckedPath) {
-        return { user: emptyUser };
-    }
-
-    // If the user is already cached, we return the cached user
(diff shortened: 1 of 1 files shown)


src/routes/+layout.ts
```diff
@@ -66,6 +58,12 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
         return await redirectToSignIn();
     }
 
+    // If the user is cached, return it
+    const cachedUser = getCachedUser();
+    if (cachedUser) {
+        return { user: cachedUser };
+    }
```
domienderle, line 65:
```suggestion
    if (cachedUser) return { user: cachedUser };
```

domienderle (changes requested):
Since it was just little changes there is not much to improve. One little suggestion I have left.

But there is one thing: Since you implemented the custom 404 error page and this is almost merged, I would propose to add the `goto` statements in this PR as well. As I understood you just wanted to add `goto` statements if some errors occurred on the `+layout.ts` page to show our custom error page. If I understood this right this should not be much effort and I would propose to do this here as soon as the 404 page is added, then we don't have to open a new PR. What do you think?

mowi12 (commented):
Good idea. I will update the 404 pr to also close this issue and cherry pick the current commit with your suggestion.

mowi12:
Will be merged in #466

PR #473, opened 2025-07-16
Added edit paper details feature

### LINKED ISSUE #61: [Use Case]: Edit Paper Details

Project Member

### Describe the use case

As project member, I want to edit the details of a paper (title, authors, year, publisher) in case the information is incorrect or not complete.

### Acceptance criteria

- ability to change title, authors, year, publisher
- all inputs must be non-blank
- year must be a positive number
- authors must contain at least one author

### Anything else?

Related to #34 and #41

Slartibartfass2:
Closes #61

## What I have made

This PR adds the update paper feature.
Previously, two edit buttons were shown; now they have moved into one in the tab list.

When the user clicks the edit button, a disabled save button is shown, and the user can modify the paper details.
The save button is enabled when there are modifications to the original paper (unsaved changes).
When the user clicks the save button, a loading state is shown and the API call is made.

Currently, we only validate the year input. Thorough input validation will be done in a different issue.

## Checklist

  - ~~[ ] unit tests~~ (not required)
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/input/ToggleableInput.svelte
+++ b/src/lib/components/composites/input/ToggleableInput.svelte
@@ -5,9 +5,16 @@
 
     type Props = WithElementRef<HTMLInputAttributes> & {
         isEditable: boolean;
+        onInputChange: (input: string) => void;
     };
 
-    let { isEditable = $bindable(false), value = $bindable(), class: className }: Props = $props();
+    let {
+        isEditable = $bindable(false),
+        value = $bindable(),
+        onInputChange,
+        placeholder,
+        class: className,
+    }: Props = $props();
 </script>
 
 <!--
@@ -29,6 +36,8 @@ Usage:
         className,
     )}
     data-testid="toggleable-input"
+    oninput={(event) => onInputChange((event.target as HTMLTextAreaElement)?.value as string)}
+    {placeholder}
     readonly={!isEditable}
     rows="1"
     {value}
--- a/src/lib/components/composites/input/ToggleableInput.svelte
+++ b/src/lib/components/composites/input/ToggleableInput.svelte
@@ -4,11 +4,13 @@
     import { cn } from "$lib/utils/shadcn-helper";
 
     type Props = WithElementRef<HTMLInputAttributes> & {
+        key: string;
         isEditable: boolean;
         onInputChange: (input: string) => void;
     };
 
     let {
+        key,
         isEditable = $bindable(false),
         value = $bindable(),
         onInputChange,
@@ -35,7 +37,7 @@ Usage:
         isEditable ? "border-input rounded-md" : "border-transparent",
         className,
     )}
-    data-testid="toggleable-input"
+    data-testid={`toggleable-input-${key}`}
     oninput={(event) => onInputChange((event.target as HTMLTextAreaElement)?.value as string)}
     {placeholder}
     readonly={!isEditable}
--- a/src/lib/components/composites/input/ToggleableInput.svelte
+++ b/src/lib/components/composites/input/ToggleableInput.svelte
@@ -23,12 +23,11 @@
 @component
 A textarea that can be toggled between read-only and editable mode.
 
-The `inputAction` and `inputActionProps` props are used to pass an action to the textarea.
-This is used in `AbstractToggleableInput` and probably won't be needed elsewhere.
+The `onInputChange` method is called whenever the input of the input changes.
 
 Usage:
 ```svelte
-    <ToggleableInput {isEditable} {value} />
+    <ToggleableInput {isEditable} {key} onInputChange={(input) => foo(input)} />
 ```
 -->
 <textarea
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -10,9 +10,18 @@
     export interface PaperDetailProp {
         key: keyof Paper;
         label: string;
-        transform?: (
+        /**
+         * The method to transform the paper property value to a displayable string.
+         */
+        toDisplayValue?: (
             value: any /* eslint-disable-line @typescript-eslint/no-explicit-any */,
         ) => string;
+        /**
+         * The method to transform the displayed string to the paper property value.
+         */
+        fromDisplayValue?: (
+            value: string,
+        ) => any /* eslint-disable-line @typescript-eslint/no-explicit-any */;
     }
 
     type Props = WithElementRef<HTMLAttributes<HTMLDivElement>> & {
@@ -24,7 +33,12 @@
     };
 
     let { prop, index = 0, loadingPaper, paper = $bindable(), isInEditMode }: Props = $props();
-    const { key, label, transform = undefined } = prop;
+    const {
+        key,
+        label,
+        toDisplayValue: toDisplayValue = (v) => v.toString(),
+        fromDisplayValue = (v) => v,
+    } = prop;
 
     const skeletonValues = [
         "w-[6rem] sm:w-[7.5rem] md:w-[11rem] lg:w-[19.8rem]",
@@ -36,13 +50,8 @@
         "w-[2.5rem] sm:w-[3.25rem] md:w-[5rem] lg:w-[9rem]",
     ];
 
-    function applyValue() {
-        const value = paper[key];
-        return transform ? transform(value) : value;
-    }
-
     function updateValue(newValue: string) {
-        paper = { ...paper, [key]: newValue };
+        paper = { ...paper, [key]: fromDisplayValue(newValue) };
     }
 </script>
 
@@ -72,7 +81,7 @@ Usage:
             isEditable={isInEditMode}
             onInputChange={updateValue}
             placeholder={`No ${label} available`}
-            value={applyValue()}
+            value={toDisplayValue(paper[key])}
         />
     {:catch}
         <ErrorIndicator errorMessage={`Couldn't load ${label}`} />
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -11,10 +11,10 @@
         key: string;
         value: unknown;
         loadingPaper: Promise<Paper>;
-        areDetailsInEditMode: boolean;
+        isInEditMode: boolean;
     };
 
-    const { key, value, loadingPaper, areDetailsInEditMode }: Props = $props();
+    const { key, value, loadingPaper, isInEditMode }: Props = $props();
 </script>
 
 <!--
@@ -37,7 +37,7 @@ Usage:
             <Skeleton class={cn("flex h-[1.625rem] rounded-full", value as string)} />
         </div>
     {:then}
-        <ToggleableInput isEditable={areDetailsInEditMode} {value} />
+        <ToggleableInput isEditable={isInEditMode} {value} />
     {:catch}
         <ErrorIndicator errorMessage={`Couldn't load ${key}`} />
     {/await}
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -6,39 +6,23 @@
     import type { HTMLAttributes } from "svelte/elements";
     import type { Paper } from "$lib/model/api/paper";
     import ErrorIndicator from "../../utils/ErrorIndicator.svelte";
+    import type { StringifiedPaper } from "$lib/model/general";
 
     export interface PaperDetailProp {
-        key: keyof Paper;
+        key: keyof StringifiedPaper;
         label: string;
-        /**
-         * The method to transform the paper property value to a displayable string.
-         */
-        toDisplayValue?: (
-            value: any /* eslint-disable-line @typescript-eslint/no-explicit-any */,
-        ) => string;
-        /**
-         * The method to transform the displayed string to the paper property value.
-         */
-        fromDisplayValue?: (
-            value: string,
-        ) => any /* eslint-disable-line @typescript-eslint/no-explicit-any */;
     }
 
     type Props = WithElementRef<HTMLAttributes<HTMLDivElement>> & {
         prop: PaperDetailProp;
         index?: number;
         loadingPaper: Promise<Paper>;
-        paper: Paper;
+        paper: StringifiedPaper;
         isInEditMode: boolean;
     };
 
     let { prop, index = 0, loadingPaper, paper = $bindable(), isInEditMode }: Props = $props();
-    const {
-        key,
-        label,
-        toDisplayValue: toDisplayValue = (v) => v.toString(),
-        fromDisplayValue = (v) => v,
-    } = prop;
+    const { key, label } = prop;
 
     const skeletonValues = [
         "w-[6rem] sm:w-[7.5rem] md:w-[11rem] lg:w-[19.8rem]",
@@ -51,7 +35,7 @@
     ];
 
     function updateValue(newValue: string) {
-        paper = { ...paper, [key]: fromDisplayValue(newValue) };
+        paper = { ...paper, [key]: newValue };
     }
 </script>
 
@@ -81,7 +65,7 @@ Usage:
             isEditable={isInEditMode}
             onInputChange={updateValue}
             placeholder={`No ${label} available`}
-            value={toDisplayValue(paper[key])}
+            value={paper[key]}
         />
     {:catch}
         <ErrorIndicator errorMessage={`Couldn't load ${label}`} />
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -41,11 +41,11 @@
 
 <!--
 @component
-Paper Detail component to display a single detail of a paper.
+Paper detail component to display a single detail of a paper, e.g., the title or the publisher.
 
 Usage:
 ```svelte
-    <PaperDetail {key} {label} {loadingPaper} {isInEditMode} bind:paper />
+    <PaperDetail prop={{ key, label }} {loadingPaper} {isInEditMode} bind:paper />
 ```
 -->
 <div id={key} class="flex flex-row gap-2" data-testid="paper-detail">
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -63,6 +63,7 @@ Usage:
     {:then}
         <ToggleableInput
             isEditable={isInEditMode}
+            {key}
             onInputChange={updateValue}
             placeholder={`No ${label} available`}
             value={paper[key]}
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -7,38 +7,74 @@
     import type { Paper } from "$lib/model/api/paper";
     import ErrorIndicator from "../../utils/ErrorIndicator.svelte";
 
+    export interface PaperDetailProp {
+        key: keyof Paper;
+        label: string;
+        transform?: (
+            value: any /* eslint-disable-line @typescript-eslint/no-explicit-any */,
+        ) => string;
+    }
+
     type Props = WithElementRef<HTMLAttributes<HTMLDivElement>> & {
-        key: string;
-        value: unknown;
+        prop: PaperDetailProp;
+        index?: number;
         loadingPaper: Promise<Paper>;
+        paper: Paper;
         isInEditMode: boolean;
     };
 
-    const { key, value, loadingPaper, isInEditMode }: Props = $props();
+    let { prop, index = 0, loadingPaper, paper = $bindable(), isInEditMode }: Props = $props();
+    const { key, label, transform = undefined } = prop;
+
+    const skeletonValues = [
+        "w-[6rem] sm:w-[7.5rem] md:w-[11rem] lg:w-[19.8rem]",
+        "w-[4rem] sm:w-[5rem] md:w-[7.3rem] lg:w-[13rem]",
+        "w-[2rem] sm:w-[2.5rem] md:w-[3rem] lg:w-[3.5rem]",
+        "w-[5rem] sm:w-[6rem] md:w-[8.6rem] lg:w-[15rem]",
+        "w-[2rem] sm:w-[2.5rem] md:w-[3rem] lg:w-[3.5rem]",
+        "w-[3.5rem] sm:w-[4.8rem] md:w-[7rem] lg:w-[12.5rem]",
+        "w-[2.5rem] sm:w-[3.25rem] md:w-[5rem] lg:w-[9rem]",
+    ];
+
+    function applyValue() {
+        const value = paper[key];
+        return transform ? transform(value) : value;
+    }
+
+    function updateValue(newValue: string) {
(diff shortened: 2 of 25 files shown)


mowi12:
I started the review, but won't get to it in the next week, so it's to review again.

src/lib/components/composites/input/ToggleableInput.svelte
Merseleo, line 1: But your `onInputChange` does not solve the "problem" you described [here](https://github.com/SE-UUlm/snowballr-frontend/issues/501) right?

Merseleo: And is it possible that the component comment isn't up-to-date?
Because the usage is not valid anymore (key and onInputChange are necessary, placeholder as well but could theoretically be changed to be optional; the description with the `AbstractToggleableInput` is inaccurate as no such component exists) and could describe the use of the `onInputChange` callback better.

src/lib/utils/model-helper.ts
```diff
@@ -13,4 +15,18 @@ function asProjectPaper(paper: Project_Paper | Paper): Project_Paper | undefined
...
+        const paperKey = key as keyof Paper;
+        if (key === "authors") {
+            stringifiedPaper[paperKey] = getNames(paper[paperKey] as unknown as Author[]);
+        } else {
+            stringifiedPaper[paperKey] = paper[paperKey].toString();
+        }
+    }
+
+    return stringifiedPaper;
+}
```
Merseleo, line 30: I think you didn't test this method? Do you think it isn't necessary or did you forgot it?

src/routes/project/[projectId]/papers/+page.svelte
```diff
@@ -218,7 +220,7 @@
             <Card.Root
                 class="border-container-border-grey relative flex h-full w-full flex-col gap-5 p-5 shadow-lg"
             >
-                <PaperDetailsCardContent {loadingPaper} />
+                <PaperDetailsCardContent {loadingPaper} paper={stringifyPaper(Paper.create())} />
```
Comment (Merseleo) [inline src/routes/project/[projectId]/papers/+page.svelte]: I think this should not be the final solution, as now the paper details card on the papers overview page doesn't work anymore. If you open it, you only get "No ... available" and no details could be loaded, probably as the stringified paper is not updated as soon as it is loaded.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
```diff
@@ -6,39 +6,69 @@
...
+    const { key, label } = prop;
+
+    const skeletonValues = [
+        "w-[6rem] sm:w-[7.5rem] md:w-[11rem] lg:w-[19.8rem]",
+        "w-[4rem] sm:w-[5rem] md:w-[7.3rem] lg:w-[13rem]",
+        "w-[2rem] sm:w-[2.5rem] md:w-[3rem] lg:w-[3.5rem]",
+        "w-[5rem] sm:w-[6rem] md:w-[8.6rem] lg:w-[15rem]",
+        "w-[2rem] sm:w-[2.5rem] md:w-[3rem] lg:w-[3.5rem]",
+        "w-[3.5rem] sm:w-[4.8rem] md:w-[7rem] lg:w-[12.5rem]",
+        "w-[2.5rem] sm:w-[3.25rem] md:w-[5rem] lg:w-[9rem]",
```
Merseleo, line 34: Maybe it is too long since I used CSS, but why do you use w-[90%] or so for the abstract but different media queries for the paper detail skeletons?

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
```diff
@@ -6,39 +6,69 @@
...
+    ];
+
+    function updateValue(newValue: string) {
+        paper = { ...paper, [key]: newValue };
+    }
 </script>
 
 <!--
 @component
 Paper Detail component to display a single detail of a paper.
```
Merseleo:
```suggestion
Paper detail component to display a single detail of a paper, e.g., the title or the publisher of the paper.
```

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
```diff
@@ -6,39 +6,69 @@
...
 
-- `value` is used as class value for the skeleton loader while the promise `loadingPaper` is pending.
-    When the promise is resolved, the value is displayed as a text.
-
 Usage:
 ```svelte
-    <PaperDetail id={key} {key} {value} {loadingPaper} {areDetailsInEditMode} />
+    <PaperDetail {key} {label} {loadingPaper} {isInEditMode} bind:paper />
```
Merseleo: Is this really correct? I think it should be {prop} as key and label only exist inside this object.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
```diff
@@ -1,120 +1,61 @@
...
 Content of the `PaperDetailsCard`, i.e. displays the details.
 
-- `allowEditModeToggle` (default: false) determines whether the paper details, e.g. the abstract or title
-are editable
-- `startInEditMode` (default: false) determines whether the paper details are rendered initially with the edit mode on
-
 Usage:
 ```svelte
-    <PaperDetailsCardContent {paper} {allowEditModeToggle} {startInEditMode} />
+    <PaperDetailsCardContent {paper} />
```
Merseleo: `loadingPaper` is necessary as well and maybe you could explain why you pass the loading paper and the paper as bindable.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
```diff
@@ -138,23 +79,24 @@ Usage:
...
-                onclick={() => (isAbstractInEditMode = !isAbstractInEditMode)}
-                size={20}
-            />
-        {/if}
     </div>
     {#await loadingPaper}
-        {#each [100, 95, 70, 82, 50, 75, 90] as width, i (i)}
-            <Skeleton class="flex h-[1.625rem] rounded-full w-[{width}%]" />
-        {/each}
-    {:then paper}
```
Merseleo, line 153: Isnt this solution more readable?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -40,4 +131,36 @@ Usage:
...
+                    {/if}
+                {/if}
+                <Pencil
+                    class="select-none hover:cursor-pointer"
+                    aria-label="Toggle Edit Paper Mode"
+                    data-testid="toggle-edit-paper-mode-btn"
+                    onclick={() => (isInEditMode = !isInEditMode)}
+                    size={24}
+                />
+            </div>
```
Merseleo, line 186: This button list appears not inside the card, if the width is very small: 

Image

Do you have an idea how to fix this? Or should we start to set a minimum width for the application to prevent such problems, as the application is not usable on mobile anyway?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,21 +1,112 @@
...
+    let isPaperModified = $state(false);
+    let isMakingApiCall = $state(false);
+
+    loadingPaper.then((p) => {
+        originalPaper = stringifyPaper(p);
+        paper = stringifyPaper(p);
+    });
+
+    $effect(() => {
+        if (paper === undefined || originalPaper === undefined) {
```
Merseleo: Can `paper` become undefined?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,21 +1,112 @@
...
+        });
+
+        await updateCall;
+
+        isMakingApiCall = false;
+    }
+
+    function isEqual(obj1: unknown, obj2: unknown): boolean {
+        return JSON.stringify(obj1) === JSON.stringify(obj2);
+    }
```
Merseleo: You only use this once, why not directly in the effect? Especially since it now looks more like a helper function that doesn't belong here?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: And one general question? Why you dont use derived or derived.by? Isn't this more correct?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,21 +1,112 @@
...
+            .updatePaper({
+                paper: paperObject,
+                mask: {
+                    paths: generateFieldMask(paperObject),
+                },
+            })
+            .response.then((updatedPaper) => {
+                originalPaper = stringifyPaper(updatedPaper);
+                paper = stringifyPaper(updatedPaper);
+            });
```
Merseleo, line 106: What happens in the catch case, i.e., when the API call fails?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,21 +1,112 @@
...
+                // TODO: this assumption is very error prone and we should use structured HTML to fix this
+                firstName: parts.slice(0, parts.length - 1).join(" "),
+                lastName: parts[parts.length - 1],
+                orcid: "",
+            };
+            return person;
+        });
+
+        return authors;
+    }
```
Merseleo, line 76: Do you want to keep this implementation this way for now and fix it later or adapt it to the solution we agreed on in the last meeting (authors are separated by semicolon and if a comma appears, then it is interpreted as last name, first name, otherwise as first name last name and add a help behind authors explaining the syntax)?

tests/e2e/project/paper/project-paper-view-page.test.ts
```diff
@@ -203,3 +206,51 @@ test.describe("Decide on Paper Tests", () => {
...
+        await titleInput.fill("Updated Title");
+
+        // Save
+        await projectPaperViewPage.savePaperChangesButton.click();
+
+        // Expect success toast
+        await expect(projectPaperViewPage.updatedPaperSuccessToast).toBeVisible();
+    });
+
+    test("When the user enters an invalid year and saves, then an error toast is shown", async ({
```
Merseleo, line 234: Maybe you could perform a reload of the page and check, that the updated wrong year isn't displayed as it was not saved?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo, line 1: Maybe you could update the comment of the component, as it is responsible for displaying and editing the paper details, right?

tests/integration/paper-components/paper-view/cards/paper-details-card.test.ts
Merseleo, line 1: Please add a test for the case, the API call fails or the paper could not be updated.

Merseleo (changes requested):
At first, I thought it would be quite a big PR, but it actually went well and was very pleasant to review because the work had been done thoroughly. The comments are mostly questions or minor things that could be improved, but don't necessarily need to be. 

However, while reviewing it, I noticed a few things that are still missing: input validation for the paper details, useful suggestions for inputs, better input for the authors, and the problem with the paper details on the paper overview page. I understand that these are a lot of tasks and don't necessarily have to be done in this PR, but have you already written issues for them or could you still do so and link them to the relevant places (or point them out with comments in the code or something)?

src/lib/components/composites/input/ToggleableInput.svelte
Slartibartfass2: No, it does not.

I updated the documentation of the `ToggleableInput`.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: I updated the documentation.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Not anymore. Good catch. I changed it to `paper.id === ""`.

`derived` reacts to changes to a single state; here we want to react to changes to two states for which `effect` can be used.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Yes, I would keep this for now. Sophisticated input validation should be done separately. I added [this issue](https://github.com/SE-UUlm/snowballr-frontend/issues/513) to track this feature.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: A toast is shown with an error message. See below.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: I moved it to `common-helper.ts` and added tests.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: I couldn't find a solution right away, but since we once said that anything below `md` is not critical to us, I would keep it this way.

src/lib/utils/model-helper.ts
Slartibartfass2: I didn't think it was that necessary, but I added some.

src/routes/project/[projectId]/papers/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/papers/+page.svelte]: Good catch, I didn't notice this. I changed it so that the selected paper is used.

tests/e2e/project/paper/project-paper-view-page.test.ts
Slartibartfass2: Great idea, I added it.

tests/integration/paper-components/paper-view/cards/paper-details-card.test.ts
Slartibartfass2: I added a test case.

Slartibartfass2 (commented):
Thank you for taking the time to review this feature.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Slartibartfass2: I updated it. I changed it while developing. Documentation just ages so bad.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Slartibartfass2: Mainly because these are the values from before and it looks good, so I didn't change anything.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: Ah ok, i was not familar with this syntax, but I understand it now.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -99,9 +99,9 @@
             error: "Failed to update the paper.",
         });
 
-        await updateCall;
-
-        isMakingApiCall = false;
+        await updateCall.finally(() => {
+            isMakingApiCall = false;
+        });
```
Merseleo: I think this change is not working properly. I edited paper details, killed the backend so the `updatePaper` call will fail, saved the changes to trigger the call and got the error toast, but the loading spinner didn't disappear.

Merseleo: Can you reproduce this behavior or did I something wrong?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: Ok, maybe we should then set the minimum width and height of the global layout, what do you think?

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Merseleo: I recognized this later too.

src/lib/utils/common-helper.ts
```diff
@@ -334,6 +334,17 @@ function callDebounced(fn: (...args: unknown[]) => void, delay = 500, args?: unk
     }
 }
 
+/**
+ * Checks whether two objects are equal by comparing their JSON string representations.
+ *
+ * @param obj1 - First object to compare
+ * @param obj2 - Second object to compare
+ * true if the objects are equal, false otherwise.
```
Merseleo:
```suggestion
 * @returns True if the objects are equal, false otherwise.
```

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Merseleo: Yeah 😔

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
Merseleo: I think you missed this comment?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: I thought always when `orginialPaper` is changed, `paper` is changed as well to `isPaperModified` is dependent on only one state, but actually it doesnt matter.

Merseleo (changes requested):
Ok only little things left and I think you missed two conversations. 
In general I want to thank you for constant refactoring and improvement of the code, even if it do not touch your changes. 

One question on how we want to deal with user experience regarding updates: it is always possible to leave the edit mode without saving the changes, i.e. locally you see the updated paper details, but they are not saved. Isnt this a bit confusing or do you think this is normal behavior if a save button exists?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
Slartibartfass2: Ah yes, I must've missed this. I updated the comment.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
Slartibartfass2: I'm still not sure if this works correctly, since Tailwind searches for all used classes in the code and only sends the used ones to the client, so if the classes are dynamically constructed, they might not be sent to the client. This is the safe way.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Not sure what you mean by that. We cannot configure the minimum window size.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: I couldn't reproduce this. Can you give me more details or make a screen recording?

Slartibartfass2 (commented):
Thank you for the review and the kind words.

I added a confirmation when the page is exited and unsaved changes exist. Do you think this is enough, or should we already show this when the edit mode is left?

The last two commits are related to your last review.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: But a minimum size of the entire layout, right? However, nothing to discuss in this PR.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCardContent.svelte
Merseleo: Ok that makes sense, so then lets keep it as you changed.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -1,35 +1,132 @@
...
+    }
+
+    beforeNavigate(({ cancel }) => {
+        if (!isPaperModified) return;
+        if (!confirm('Are you sure you want to leave this page? You have unsaved changes that will be lost.')) {
+            cancel();
+        }
+    });
```
Merseleo, line 131: This seems not work completely (see Mattermost)

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: I think it is fixed now.

Merseleo (changes requested):
I think only one problem left with the beforeNavigate, after the fix it can be merged.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Merseleo: And I think I found a solution to your problem (and the problem, that if you navigate between the papers, the paper details are not updated correctly): 

```typescript
$effect(() => {
  const currentPaperPromise = loadingPaper;
  if (!currentPaperPromise) return;

  currentPaperPromise
    .then((resolvedPaper) => {
      // Skip if a new promise has been assigned while waiting
      if (currentPaperPromise !== loadingPaper) return;

      originalPaper = stringifyPaper(resolvedPaper);
      paper = stringifyPaper(resolvedPaper);
    })
    .catch(() => {});
});
```

in the `PaperDetailsCard` instead of the 
```typescript
loadingPaper.then((p) => {
        originalPaper = stringifyPaper(p);
        paper = stringifyPaper(p);
    });
```
code snippet, because if `loadingPaper` is updated, the component doesnt react to the change.

Merseleo:
And there's another thing I noticed when I took a closer look at it again. Your solution always sends the entire new paper object, and the field mask always contains all fields. Wouldn't it be smarter to keep the original paper as a paper object and then recast the string paper (which you do anyway) and then determine the diff object and only send that?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Great work!

Merseleo (approved):
> And there's another thing I noticed when I took a closer look at it again. Your solution always sends the entire new paper object, and the field mask always contains all fields. Wouldn't it be smarter to keep the original paper as a paper object and then recast the string paper (which you do anyway) and then determine the diff object and only send that?
I think it is ok to not deal with this "problem" in this PR, so I created #521 to handle this, but I think this isn't that important.

I tested your bug fix and it works, so let's finally merge this.

PR #475, opened 2025-07-16
Added script to retrieve all used api calls

### LINKED ISSUE #474: [Feature]: Add script to retrieve all used API calls

_No response_

To know which API calls we need to implement in the backend and which may be removed from the API, we require a script that searches the whole frontend for API call usages.

Add a script that counts the used API calls.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #474 

## What I have made

Added a script that retrieves all used API calls in the frontend.

## Checklist

### DIFF
--- a/scripts/get-all-used-calls.js
+++ b/scripts/get-all-used-calls.js
@@ -1,0 +1,49 @@
+import fs from "fs";
+import path from "path";
+import { sync as globSync } from "glob";
+
+// Regular expression to match backendService calls
+const apiCallRegex = /backendService\s*\.\s*(\w+)/g;
+
+// Function to read file and find API calls
+const findApiCalls = (filePath) => {
+    const content = fs.readFileSync(filePath, "utf-8");
+    const matches = [...content.matchAll(apiCallRegex)];
+    return matches.map((match) => ({
+        method: match[1],
+        file: filePath,
+        line: content.substring(0, match.index).split("\n").length,
+    }));
+};
+
+// Search all files in src directory
+const srcPath = path.join(path.dirname(new URL(import.meta.url).pathname), "..", "src");
+const files = globSync(path.join(srcPath, "**/*.{js,ts,svelte}"));
+
+// Collect all API calls
+const allCalls = files.flatMap(findApiCalls);
+
+// Group by method name
+const groupedCalls = Object.groupBy(allCalls, (call) => call.method);
+
+// Print results
+console.log("API Calls Usage:");
+Object.entries(groupedCalls).forEach(([method, usages]) => {
+    console.log(`\n${method}:`);
+    usages.forEach((usage) => {
+        console.log(`  ${usage.file}:${usage.line}`);
+    });
+});
+
+// Print total and distinct counts
+const distinctMethodsCount = Object.keys(groupedCalls).length;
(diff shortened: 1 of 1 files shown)


scripts/get-all-used-calls.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Regular expression to match backendService calls
+const apiCallRegex = /backendService\s*\.\s*(\w+)/g;
+
+// Function to remove comments from code
+const removeComments = (content) => {
+    // Remove single line comments
+    content = content.replace(/\/\/.*/g, "");
+    // Remove multi-line comments
+    content = content.replace(/\/\*[\s\S]*?\*\//g, "");
```
luca-schlecker, line 21: Why did you opt for `[\s\S]` instead of `.`?

scripts/get-all-used-calls.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Search all files in src directory
+const srcPath = path.join(path.dirname(new URL(import.meta.url).pathname), "..", "src");
+const files = globSync(path.join(srcPath, "**/*.{js,ts,svelte}"));
+
+// Collect all API calls
+const allCalls = files.reduce((acc, file) => {
+    const calls = findApiCalls(file);
+    return [...acc, ...calls];
+}, []);
```
luca-schlecker: I think `.flatMap` is to be preferred here.
```patch
- const allCalls = files.reduce((acc, file) => {
-     const calls = findApiCalls(file);
-     return [...acc, ...calls];
- }, []);
+ const allCalls = files.flatMap(findApiCalls);
```

scripts/get-all-used-calls.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Group by method name
+const groupedCalls = allCalls.reduce((acc, call) => {
+    if (!acc[call.method]) {
+        acc[call.method] = [];
+    }
+    acc[call.method].push({ file: call.file, line: call.line });
+    return acc;
+}, {});
```
luca-schlecker: I think `Object.groupBy` is to be preferred here.
```patch
- const groupedCalls = allCalls.reduce((acc, call) => {
-     if (!acc[call.method]) {
-         acc[call.method] = [];
-     }
-     acc[call.method].push({ file: call.file, line: call.line });
-     return acc;
- }, {});
+ const groupedCalls = Object.groupBy(allCalls, (call) => call.method);
```

luca-schlecker (changes requested):
Works as intended, just three minor things I noticed. 👍 This script is very helpful.

scripts/get-all-used-calls.js
Slartibartfass2: First, I have to say that I usually create such scripts via AI, as this is the best use case for it.

`[\s\S]` is not the same as `.` because the former matches newlines and the latter does not. In this case, we want to match multi-line comments, which contain newlines.

Slartibartfass2 (commented):
Thank you for the review. Great suggestions.

scripts/get-all-used-calls.js
luca-schlecker: Learned something new today. Didn't know `.` ignored newlines, the description "matches any single character" lead me to believe it wouldn't. Thank you ^^

luca-schlecker (approved):

PR #479, opened 2025-07-17
Show archived user projects on homepage

### LINKED ISSUE #467: [Bug]: Archived projects not shown on homepage

_No response_

### What happened?

When you open the homepage and have an archived project, then this project will not be shown anymore as the `getAllUserProjects` call used to retrieve these projects only returns active projects

### Steps to reproduce

1. Assert that you have an archived project
2. Go to the homepage

### What did you expect?

I would except that also the archived projects for the user are loaded (using `getAllArchivedProjectsForUser`) and shown in the frontend.

### Anything else?

_No response_

Merseleo:
Closes #467 

## What I have made

- I added the request to the (mock) backend for the archived user projects, so on the homepage not only active but also archived projects are displayed

**Note:** With the current example data of the mock backend it is very unlikely that a user have an archived project, so to properly test this fix I recommend to increase the probability. Furthermore, we should discuss what should be possible when we click on an archived project and maybe fix access in this PR too.

## Checklist

  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests

### DIFF
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -56,13 +56,21 @@ export const load: PageLoad = async ({ depends, parent }) => {
 
     const userId = user.id;
 
-    const allUserProjects = backendService.getAllProjectsForUser({ id: userId }).response;
+    const allActiveUserProjects = backendService.getAllProjectsForUser({ id: userId }).response;
+    const allArchivedUserProjects = backendService.getAllArchivedProjectsForUser({
+        id: userId,
+    }).response;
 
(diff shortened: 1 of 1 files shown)


luca-schlecker (approved):
I really like the grayed-out look archived projects have. Your implementation also honors that archived projects should come last, which I find really important.
As to what should happen on click: I think the current behavior is fine. You should still be able to look at it, and modification is (eventually) denied through the backend. Though, it would probably be nice to disable/gray-out interactions that would fail. But I think that would be its own PR.

LGTM! 🚀

PR #486, opened 2025-07-24
Use `GetInviteCandidates` instead of `GetAllUsers` to get suggestions when inviting users to projects

### LINKED ISSUE #437: [Bug]: Invite User Dialog has Wrong Focus in Firefox

_No response_

### What happened?

See: https://github.com/user-attachments/assets/893987bc-ea3d-49a0-91f2-008492e2848c

As soon as the user uses his arrow keys (Down, in the beginning), the focus switches to the whole list. You cannot select another user by clicking the down arrow. The list is scrolling in this case.

### Steps to reproduce

1. Go into a project in which you are admin
2. Enter any letter
3. As soon as any suggestions pop up, click the arrow down button

### What did you expect?

The same behaviour as in the Safari browser.
Select another user by up or down arrow, confirm with enter.

### Anything else?

_No response_

### LINKED ISSUE #478: [Bug]: Request all users instead of invite candidates in InviteUserDialog

#16

### What happened?

When I invite other users to my project, then I get some suggestions. To get candidates to invite, we currently request all users existing on the server and (fuzzy) searching them in the frontend. This is not very privacy friendly and very performant. That's why we outsourced the search task into the backend.

**Note:** It is not really a bug, what previously intended. However, the new API version introduced a new call that should be implemented now.

### Steps to reproduce

/

### What did you expect?

I would expect that not all users are requested but only the invite candidates.

### Anything else?

_No response_

### LINKED ISSUE #489: [Bug]: Make open project button a loading button

_No response_

### What happened?

If a new project is created and should be openend, then this button can be pressed multiple times, because it is now loading button.

### Steps to reproduce

1. Start the mock backend with response delay
2. Start the frontend
3. Click on `Create Project` to open the project creation dialog
4. Type in a name an click on `Create Project` to really create a project
5. Click the `Open` button

### What did you expect?

I expected the button to be a loading button, which allows only one click, then goes into loading mode, until the project is opened, but instead the button is now loading button and can be clicked multiple times.

### Anything else?

_No response_

Merseleo:
Closes #478

## What I have made

- I bumped the api version to v0.5.1
- I removed the usage of `GetAllUsers` to initially request all users and filter them in the frontend for invitation suggestion
  - To still provide suggestions, I integrated the new `GetInviteCandidates` call and every time when the users input some strings into the `InviteUserInput` after 200ms or so invite candidates are requested and stored for suggesting users to the user
- I made the button for opening a project after it was created a "loading button", i.e., added a loading state to the `CreateProjectDialog` indicating whether this component is currently navigating to the project dashboard (Closes #489)
- I fixed the problem, that when input something in the `InviteUserInput` in Firefox or Chrome and trigger suggestions to show up, it was not possible to navigate through these suggestions using the up / down arrows (no I do not focus the list anymore, but only highlight the selected item and scroll it into view, Closes #437)
- I added and integrated the helper functions `debounce` and `callDebounced` to create / call a function after a certain delay
- furthermore I did some minor refactorings and deleted old, unused or unnecessary code
- I turned off some tests that are either super flaky and we need further investigation or that are not supported with the v0.5.1 and a mock backend running on version v0.6.0 (used in the e2e tests in the pipeline)

**Note for the reviewer:** I suggest to test this PR using 1) the mock backend e.g. for long suggestion lists when using the search input "@example" or for testing the exclusion of users in projects and 2) the real backend (see `feat/75-feature-implement-getinvitecandidates-call` to test the real matching behavior, however then manually some test users must be inserted into the database).

And maybe you could have a look at the new tests (especially at the ones that were missing before this PR); please check on your own, whether I missed anything.

## Checklist

  - [x] unit tests
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (only refactor, did not add any new functionality)


src/lib/components/composites/input/ChipsInput.svelte
```diff
@@ -282,6 +303,14 @@ Usage:
         </div>
     {/if}
     <!-- suggestions list -->
+    <!-- TODO: note for the reviewer: i am not very happy with this loading text
+        and its position: any suggestions? -->
```
domienderle: The first thing coming to my mind would be to simply add a loading spinner to the right of the input field as long as invite candidates are loaded. What do you think? I would not disable anything though.

src/lib/components/composites/input/InviteUsersInput.svelte
```diff
@@ -1,63 +1,113 @@
...
+    let isErrorWhileLoadingCandidates = $state(false);
+
+    /** Collection of loaded invite candidates based on the current input and already added
+     *  input candidates that should be definitely invited
+     */
+    let loadedInviteCandidates = $state<User[]>([]);
+
+    /**
+     * Checks whether a candidate, represented by the input email or name, already was added as
+     * invitee.
```
domienderle:
```suggestion
     * Checks whether a candidate, represented by the input email or name, already was added as
     * an invitee.
```

src/lib/components/composites/input/InviteUsersInput.svelte
```diff
@@ -1,63 +1,113 @@
...
+     * Fetches a list of invite candidates from the backend based on the given search string.
+     * The backend is responsible for filtering and ranking the users.
      *
-     * Furthermore, the filtered list is sorted by the score from the FZF algorithm, i.e.
-     * the members with the best matching name or email are at the beginning of the list (and will
-     * appear at the top of the suggestions list).
+     * In case of a loading error, an empty list is returned and an error flag is set to
+     * display an error alert.
      *
      * @param input - the content of the input field, i.e. the search string
```
domienderle:
```suggestion
     * @param input - the content of the input field, i.e., the search string
```

src/lib/components/composites/input/InviteUsersInput.svelte
```diff
@@ -1,63 +1,113 @@
...
+            .catch(() => {
+                isErrorWhileLoadingCandidates = true;
+                loadedInviteCandidates = [];
+                return [];
+            });
     }
 
     /**
-     * Checks, whether a given input is a valid name of a registered user or an email.
+     * Checks, whether a given input is valid, i.e.
```
domienderle:
```suggestion
     * Checks whether a given input is valid, i.e.
```

src/lib/components/composites/search-bar/SearchBar.svelte
domienderle, line 1: Can you tell me where you use it? I don't find any search bar where the search is delayed and where the `callDebounce` function would be useful.

tests/unit/utils/common-helper.test.ts
domienderle, line 1: Add a test, "When a debounced function is called multiple times (`callDebounced`), then all function calls are invoked".

src/lib/components/composites/input/InviteUsersInput.svelte
```diff
@@ -48,14 +48,13 @@
...
-                );
-
-                loadedInviteCandidates = [...newInviteCandidates, ...retainedCandidates];
+                const newInviteCandidates = users.filter((user) => !invitees.includes(user.email));
+
+                newInviteCandidates.forEach((candidate) => {
+                    if (!loadedInviteCandidates.includes(candidate)) {
+                        loadedInviteCandidates.push(candidate);
+                    }
+                });
```
domienderle, line 57: You kind of filter two times, while you could do this in one filter call here.
```suggestion
                const newInviteCandidates = users.filter((user) =>
                    !invitees.includes(user.email) && !loadedInviteCandidates.includes(user));
                loadedInviteCandidates.push(...newInviteCandidates);
```

tests/integration/input/invite-users-input.test.ts
```diff
@@ -120,36 +100,36 @@ describe("InviteUsersInput", () => {
...
-        expect(chips).toHaveLength(1);
-        const chip = chips[0];
-        expect(chip).toHaveTextContent(getName(newUser));
+        expect(chips).toHaveLength(2);
+        expect(chips[0]).toHaveTextContent(getName(newUser));
+        expect(chips[1]).toHaveTextContent("jane2.doe@example.com");
     });
 
-    test("When invalid user email is typed and Tab is pressed, then error message is shown", async () => {
+    test("When valid user email is typed and for example 'Tab' pressed, then error message is shown", async () => {
```
domienderle: Why did you change this? It should be an invalid user email, right?

src/lib/components/composites/project-components/CreateProjectDialog.svelte
domienderle, line 1: Thank you for implementing this, too. You can resolve this on your own.

src/lib/components/composites/input/ChipsInput.svelte
domienderle, line 1: My browser (chrome) shows following behaviour: If there is something I have typed in at an earlier time, than this is suggested. Now, if the `getInviteCandidates` call returns some invitees, then they are suggested as well. So we have to different suggestions at the same time, while my browser suggestion is prioritized. What do you think? Should we keep this or should we disable the browser suggestions in this context? Because two overlapping suggestions are ugly imo.

tests/integration/project-components/create-project-dialog.test.ts
domienderle, line 1: If we decided on how to handle the `ESC`/cross behaviour of the dialog add tests for this too.

tests/integration/settings/project-settings/members/invite-users-dialog.test.ts
domienderle, line 1: If we decided on how to handle the `ESC`/cross behaviour of the dialog add tests for this too.

domienderle (changes requested):
Great work. Your code looks really good and there is not much to do anymore. Anyway, here are a couple of thing to talk about.

- At the moment we have the following inconsitency: If a user is added to the invite user dialog, then the dialog is closed using `ESC` or the cross in the top right corner and then opening the dialog again then the already added users are loaded with their email addresses, whereas all now added users are added by their names. I know this a known issue we already talkes about, but what do you think of
  - always showing the email address of the added user?
  - disabling the option to close the dialog using `ESC` or
  - (I would prefer this) set back the invitees `invitees = []` as you do it `onClick` for the cancel button. 

  Or maybe you have a better idea, because imo it is really ugly to have a mixture of emails and names in this field.

- Can you please create an issue for fixing the tests, once the api versions and the mock-backend are compatible again.

src/lib/components/composites/input/InviteUsersInput.svelte
Merseleo: However, your suggestion has not the same behavior as I return the newInviteCandidates and in your version the user doesn't get a suggestion twice. So if you type "bob" then you get a suggestion and then type "bob I" and delete the "I" so have the input "bob" again. Then the suggestion for "bob" is not shown anymore, as it was already loaded and is filtered out. I already implemented your suggestion, but if I remember correctly I noticed the problem I described.

src/lib/components/composites/search-bar/SearchBar.svelte
Merseleo: You are right, we do not use it at the moment, but used it earlier. I think the reason for the change was, that we noticed with our current amount of papers on the papers overview page, it is no problem to fetch all papers and search locally. However, the intention was that you can search and similar to the invite suggestions to only get the papers matching the search query from the backend and then you do not want to request this after every key press. So we implemented the opportunity to change it back quickly and I only preserved this opportunity which is not wrong imo.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: You are right. Most of the times I only test my implementations in Firefox and Safari and I turned off these suggestions. But I think I found a way to disable browser suggestions.

tests/integration/input/invite-users-input.test.ts
Merseleo: A copy paste mistake, I will correct it and squash it (I hope you are ok with that)

tests/unit/utils/common-helper.test.ts
Merseleo: I will squash it into the corresponding commit

tests/integration/settings/project-settings/members/invite-users-dialog.test.ts
Merseleo: I changed it, so it does not explicitly reset the invitee list on cancel click, but whenever the dialog is closed. Do you think we need to test all dialog closing operations separately? Same for your other comment.

src/lib/components/composites/input/ChipsInput.svelte
Merseleo: Yeh, quite a good idea, thank you.

Merseleo (commented):
> Great work. Your code looks really good and there is not much to do anymore. Anyway, here are a couple of thing to talk about.
> 
> * At the moment we have the following inconsitency: If a user is added to the invite user dialog, then the dialog is closed using `ESC` or the cross in the top right corner and then opening the dialog again then the already added users are loaded with their email addresses, whereas all now added users are added by their names. I know this a known issue we already talkes about, but what do you think of
>   
>   * always showing the email address of the added user?
>   * disabling the option to close the dialog using `ESC` or
>   * (I would prefer this) set back the invitees `invitees = []` as you do it `onClick` for the cancel button.
>   
>   Or maybe you have a better idea, because imo it is really ugly to have a mixture of emails and names in this field.

I also prefer to reset the invitees as when clicking "Cancel".


> * Can you please create an issue for fixing the tests, once the api versions and the mock-backend are compatible again.

I wouldn't create an extra issue, but leave @luca-schlecker a note, because he is working on this in parallel. So if he finishes his PRs earlier, than I will drop these commits, otherwise he will do in his PRs (and I will make a note there). Are you ok with this solution?

src/lib/components/composites/input/ChipsInput.svelte
domienderle: This really looks good.

src/lib/components/composites/search-bar/SearchBar.svelte
domienderle: Ok, all right. In this case I'm fine with it. Thanks for the explanation. Resolve this on your own.

src/lib/components/composites/input/InviteUsersInput.svelte
domienderle: I didn't find this problem. Then we should keep your solution.

src/lib/components/composites/input/ChipsInput.svelte
domienderle: Seems like it works now. Thank you.

tests/integration/settings/project-settings/members/invite-users-dialog.test.ts
domienderle: That is a good question. It is kind of ugly in every way we do it. I don't think we have to test each dialog closing operation separately, but I would suggest to test the closing in every dialog. You've already tested it in the `create-project-dialog.test.ts`, but not in `invite-users-input.test.ts`. I know, it is almost the same code, but in case one of the two gets remove for some reason at some point I guess it would be forgotten to add the test two the other class then, if you know what I mean.

domienderle:
> I wouldn't create an extra issue, but leave @luca-schlecker a note, because he is working on this in parallel. So if he finishes his PRs earlier, than I will drop these commits, otherwise he will do in his PRs (and I will make a note there). Are you ok with this solution?

Yes, that is ok for me.

domienderle (changes requested):
Really good work. Now it is really hard to find anything ;) 

I guess we have only one missing test left, then we are good to go.

tests/integration/settings/project-settings/members/invite-users-dialog.test.ts
Merseleo: The `invite-users-input.test.ts` is an input wrapper which cannot be closed. And the `invite-users-dialog.test.ts` does contain a test for checking the closing behavior: "When dialog is closed, then entered invitation candidates are deleted" (it is only called slightly different). So didn't you recognize this test case or do I still dont see what is missing?

domienderle: Ah sorry, no I was just blind. It doesn't matter. My mistake.

domienderle (approved):

Issue #483, opened 2025-07-20
[Feature]: Fetcher Options in Project Settings


luca-schlecker:
### Is your feature request related to a problem? Please describe

Fetcher Options should be settable in the Project Settings. See SE-UUlm/snowballr-backend#124, SE-UUlm/snowballr-backend#125 and SE-UUlm/snowballr-backend#164 for more information on fetcher options.

### Describe the solution you'd like

Each fetcher should be able to be configured using its available options, retrieved using the according proposed gRPC call.
These options should then also be sent to the server when updating project settings.

PR #491, opened 2025-07-30
Implement user email verification

### LINKED ISSUE #490: [Feature]: Implement User Email Verification

_No response_

In order to support the new API version (0.7.0), the frontend needs to be able to verify the email of a user.

Implement the new verifyEmail call.

_No response_

### Anything else?

_No response_

mowi12:
Closes #490 

## What I have made

- Updated to new API version
- Implemented verify email logic
   - Updated old logic

This PR needs to wait for #483.

## Checklist

  - [ ] ~~unit tests~~ no new components
  - [ ] ~~integration tests~~ no new components
  - [x] end-to-end tests (only updated register tests, as mock backend will not receive verify logic updates)


src/routes/(auth)/signin/+page.svelte
```diff
@@ -5,38 +5,21 @@
...
-                .authenticationStatus;
-            if (authStatus === AuthenticationStatus.AUTHENTICATED) {
-                await goto("/");
-            }
-        } catch (error) {
-            console.error("There was an error acquiring the authentication status:", error);
-        }
-
-        isLoadingAuthStatus = false;
-    });
```
luca-schlecker, line 38: Why did you remove this section? It should be responsible for redirecting to the home page in case the user is already signed in and visits the `/signin` page? Is this behavior undesired?

src/routes/(auth)/signin/+page.svelte
```diff
@@ -51,8 +34,9 @@
         await backendService
             .login(userData)
             .then(async () => await goto("/"))
-            .catch((error) => {
-                if (error.code === StatusCodes.UNAUTHENTICATED) {
+            .catch((error: RpcError) => {
+                const errorCodeValue = GrpcStatusCode[error.code as keyof typeof GrpcStatusCode];
+                if (errorCodeValue === GrpcStatusCode.UNAUTHENTICATED) {
```
luca-schlecker: I've seen this throughout this PR. Is there something wrong with the existing `StatusCodes` enum?

src/routes/(auth)/signup/+page.ts
```diff
@@ -0,0 +1,19 @@
...
+            .authenticationStatus;
+        if (authStatus === AuthenticationStatus.AUTHENTICATED) {
+            await goto("/");
+        }
+    } catch (error) {
+        console.error("There was an error acquiring the authentication status:", error);
+    }
+
+    return {};
+};
```
luca-schlecker, line 19: Does sveltekit allow you to move this one directory up? It is the same for `signin` and `signup`.

src/routes/(auth)/verifyemail/+page.svelte
```diff
@@ -0,0 +1,75 @@
...
+</svelte:head>
+
+<main class="flex h-full w-full flex-col items-center justify-center gap-1 text-center">
+    {#await verificationPromise}
+        <h1 class="mb-4 text-8xl">Verifying...</h1>
+        <div class="text-default">Please wait while we check your verification link.</div>
+    {:then}
+        <h1 class="mb-4 text-8xl">Success!</h1>
+        <div class="text-default">
+            Your email address has been successfully verified. You will automatically be redirected
```
luca-schlecker:
```suggestion
            Your email address has been successfully verified. You will be redirected automatically
```

src/routes/(auth)/verifyemail/+page.svelte
```diff
@@ -0,0 +1,75 @@
...
+    import { Button } from "$lib/components/primitives/button";
+    import type { RpcError } from "@protobuf-ts/runtime-rpc";
+    import { GrpcStatusCode } from "@protobuf-ts/grpcweb-transport";
+
+    let { data } = $props();
+    const { verificationPromise } = data;
+
+    type ErrorDetails = { headline: string; body: string } | RpcError;
+
+    function getErrorDetails(error: ErrorDetails): { headline: string; body: string } {
```
luca-schlecker, line 13: Why not just `RpcError` as the type of the `error` parameter?

src/routes/(auth)/verifyemail/+page.svelte
```diff
@@ -0,0 +1,75 @@
...
+        switch (statusCodeValue) {
+            case GrpcStatusCode.INVALID_ARGUMENT:
+                return {
+                    headline: "Verification Failed",
+                    body: "This verification link is invalid. Please try signing up again to receive a new link.",
+                };
+            case GrpcStatusCode.DEADLINE_EXCEEDED:
+                return {
+                    headline: "Verification Failed",
+                    body: "The verification link has expired. Please try signing up again to receive a new link.",
```
luca-schlecker, line 31: Is this currently in the backend? Did your PR handle reregistration?

luca-schlecker (changes requested):
I already verified the behavior in a manual test. This is looking great! 👍 Some minor things still caught my eye though.

src/routes/(auth)/signin/+page.svelte
mowi12: I explained this in the third commit in the commit message. Since I added the `+page.ts` files either way, I moved the logic of the validation if the user is already logged in, into these files. This allows us to remove the hacky way of simply add `opacity-0` while loading.
Now the page isn't even loaded if the user is signed in, only if he is not.

src/routes/(auth)/signin/+page.svelte
mowi12: You observed correctly. The reason for this change is simple. I needed a way to map the `GrpcStatusCode` string to the code. Our custom implementation of this list, the `StatusCodes` enum, only mapped an enum value to a string value, but not the actual code.

Since the there already is an enum, which does this, we can remove our custom implementation of this list (enum) and use the one from the library instead.
It's not as clean as before, but I think this is a better solution since we do not need to manually update the enum as it's not part of the grpc library.

src/routes/(auth)/signup/+page.ts
mowi12: I don't think that's possible as the signout page would also check if you're logged in and redirect instead of logging you out : D

Maybe we can refactor this to check specific paths, but that is too much work for only 10 lines of duplicated code.
The `newpassword` and `resetpassword` functionality isn't even implemented yet, so there is still time to refactor the `+page.ts` files when we know if these pages also need this code. If yes, then we can definitely think about that!

src/routes/(auth)/verifyemail/+page.svelte
mowi12: Reason is this section in the `+page.ts` file:

```ts
const token = url.searchParams.get("token");

    if (!token) {
        return {
            verificationPromise: Promise.reject({
                headline: "Verification Failed",
                body: "The verification link is missing a token. Please check the link provided in your email.",
            }),
        };
    }
```

I check if there is even a token in the URL. If not, I don't even send a request to the backend, but simply return the headline and body. Therefore, I need these two fields in the `ErrorDetails`.

src/routes/(auth)/verifyemail/+page.svelte
mowi12: Yes, the backend is handling this case.

I know the occurrences are sometimes a bit weird, but I have my reasons, let me explain:

In the API documentation I only have the following two error codes:

- DEADLINE_EXCEEDED
- INVALID_ARGUMENT

The invalid argument is pretty straight forward. If there was no token provided (which shouldn't happen since I check this in the frontend as well), this error returns.
The deadline exceeded is a bit more complicated. If there is no token in the database, we could argue, that we should return NOT_FOUND or something like that, but where do we differentiate between the token is simply not there and, the token has expired and therefore been deleted from the database?
Since I implemented this behaviour exactly like that I argued, that we should always return DEADLINE_EXCEEDED as we cannot differentiate in the backend between these two cases. And since the user doesn't really care if the token is wrong or already exceeded and always use this error code.

I hope this explains this : )

mowi12 (commented):
I hope I could explain everything to you!
The small changes in the text are applied.

src/routes/(auth)/signin/+page.svelte
luca-schlecker: Yes, of course. The `+page.ts` file came later in the review and already answers this question. Forgot to remove the comment. 😅

src/routes/(auth)/signin/+page.svelte
luca-schlecker: I do not quite understand yet. As far as I can tell, the `RpcError` for example provides the error code in string representation, the values of which are the ones previously enumerated in `StatusCodes`. I could not find a place where you'd encounter an integer status code, maybe you could point me towards one, if I understood that correctly? Furthermore, this is arguably worse DX, it has to be easier (the first line seems to be just boilerplate screaming to be extracted). While I agree that using the library-provided error enumeration is preferred, I believe so to be only the case if the DX isn't worse (note: these error codes haven't changed in what feels like forever).

I am open to moving forward with dropping our own enum, but the new way should come really close in terms of ease of use.

src/routes/(auth)/signup/+page.ts
luca-schlecker: Ok, we'll move ahead with this. Serves the purpose. 👍

src/routes/(auth)/verifyemail/+page.svelte
```diff
@@ -0,0 +1,75 @@
...
+
+<main class="flex h-full w-full flex-col items-center justify-center gap-1 text-center">
+    {#await verificationPromise}
+        <h1 class="mb-4 text-8xl">Verifying...</h1>
+        <div class="text-default">Please wait while we check your verification link.</div>
+    {:then}
+        <h1 class="mb-4 text-8xl">Success!</h1>
+        <div class="text-default">
+            Your email address has been successfully verified. You will be redirected automatically
+            to the sign-in page.
```
luca-schlecker: I overlooked the next line 🙈. Sorry 😅 .

```
            Your email address has been successfully verified. You will be automatically redirected
            to the sign-in page.
```

src/routes/(auth)/verifyemail/+page.svelte
luca-schlecker: I see, this makes sense. 👍

src/routes/(auth)/verifyemail/+page.svelte
luca-schlecker: I did a bad job explaining myself there (although this is already good to have confirmed). I meant to ask if one can currently sign up again in case the token never arrives.

luca-schlecker (changes requested):
Continuing the discussion.

src/routes/(auth)/verifyemail/+page.svelte
mowi12: Ahh. No, there is no possible to request a new request token or rather a new email with a new token.
We need a new API call for this or need to rethink the logic in the backend for that.

I already have an idea how to handle this, but this will be done in a new issue.

src/routes/(auth)/signin/+page.svelte
mowi12: The `error.code` is a `String` (e.g., `UNAUTHENTICATED`).
The `GrpcStatusCode.UNAUTHENTICATED` is a `number` and evaluates to `16`.

I therefore need a way to map one to the other. You're right—this is definitely not pretty.

To improve this, I added two helper methods, so the conversion logic is centralized. This way, we don’t have to repeat it every time. I hope this solution is satisfactory!

mowi12 (commented):
I refactored the grpc status code comparison. I think this is a good solution.

src/routes/(auth)/verifyemail/+page.svelte
luca-schlecker: I agree, make sure to open an according issue. 👍

src/routes/(auth)/signin/+page.svelte
luca-schlecker: Yes, that's why there originally was the string enum. If we want to use the `GrpcStatusCode`, this conversion becomes necessary. The helper functions you've introduced significantly reduce the boilerplate, we can move forward with this. :+1:

luca-schlecker (approved):
LGTM, great work 🚀

PR #500, opened 2025-08-10
feat: remove double clicks in favor of extra buttons

### LINKED ISSUE #499: [Feature]: Remove Double Clicks in Favor of Buttons

_No response_

Double clicks are an awkward interaction in the web. An icon better indicates that there is an action to be performed. Furthermore, these double clicks seem to be messing with the end-to-end tests, making them fail quite often.

All double clicks should be removed in favor of buttons.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #499 

## What I have made

Introduced icon links instead of double clicks.
Hopefully this fixes the e2e-test flakiness.

## Checklist

  - [ ] unit tests
  - [x] integration tests
  - [x] end-to-end tests

### DIFF
--- a/src/lib/components/composites/paper-components/PaperListEntry.svelte
+++ b/src/lib/components/composites/paper-components/PaperListEntry.svelte
@@ -14,6 +14,7 @@
     import { reviewMode } from "$lib/global-state/review-mode-state.svelte";
     import type { Paper } from "$lib/model/api/paper";
     import type { Project_Paper } from "$lib/model/api/project";
+    import { ExternalLink } from "lucide-svelte";
 
     /**
      * Interface for the paper list entries.
@@ -65,6 +66,7 @@ as well as, if not in review mode, the review information about this paper.
 Furthermore this component is clickable and navigates to the corresponding paper view,
 if the onClick() event handler is not overridden. Otherwise it executes the custom event handler
 on a single click. A double click always causes the navigation to the paper view.
+If an onClick event handler is provided, a link icon is shown to open the paper.
 
 Usage:
 ```svelte
@@ -81,7 +83,7 @@ Usage:
 >
     <div
         class={cn(
-            "flex flex-auto rounded-md px-3 py-2",
+            "flex flex-auto gap-2 rounded-md px-3 py-2",
             !reviewMode.isActivated && isProjectPaper(paper)
                 ? `border-l-4 ${getStatusColor(paper.decision, "border")}`
                 : "",
@@ -91,6 +93,16 @@ Usage:
             loadingPaper={Promise.resolve(asPaper(paper))}
             loadingPaperId={Promise.resolve(paperId)}
         />
+        {#if onClick}
+            <a
+                class="text-muted-foreground self-start"
+                {href}
+                onclick={(e) => e.stopPropagation()}
+                title="Open Paper"
+            >
+                <ExternalLink class="mt-[0.375rem] size-[1rem]" />
+            </a>
+        {/if}
     </div>
     {#if !reviewMode.isActivated && isProjectPaper(paper)}
         {#each paper.reviews as review (review.id)}
--- a/src/lib/components/composites/paper-components/ReadingListEntry.svelte
+++ b/src/lib/components/composites/paper-components/ReadingListEntry.svelte
@@ -5,6 +5,7 @@
     import PaperBookmarkButton from "$lib/components/composites/button/PaperBookmarkButton.svelte";
     import DownloadButton from "$lib/components/composites/button/DownloadButton.svelte";
     import type { Paper } from "$lib/model/api/paper";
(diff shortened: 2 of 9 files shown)


mowi12 (changes requested):
Before I have a closer look at your code, there are two points.

1. Is it possible that the right side (the slide-in window) does not open, if you click on the external-link icon?

Currently, it first opens the side view and the redirects.

https://github.com/user-attachments/assets/ad0c933a-2bfc-4616-9b74-74f8fe8580ad

2. There are 3 teamscale warning. Please check them out and find an appropriate solution.

Generally, on the first look, everything looks pretty good!

luca-schlecker:
1. I fixed the issue you've encountered. Click-events for these link icons are now prevented from being propagated.
2. These findings are about empty blocks in the tests. I deem them irrelevant, as these empty blocks are explicitly there.

mowi12 (changes requested):
Perfect, thank you!
This is the exact behaviour I wanted 👍🏼 

Regarding the teamscale warnings - I agree, this is explicit, but maybe simply add some comments, so this is clear for everyone who looks at the code?

After looking closer at the code, I'm not quite sure about some things.
I checked out `develop` in order to see the old behaviour. I noticed, that there never was any double click behaviour in the reading list, e.g. So my question now is, was this intended? If yes, why didn't the code work, since the `ReadingListEntry` clearly had the code which handled the single/double click code.
And why do we add the new code snippet you added in the three components? I tested in the reading list without the addition, and it still works ...

Could you please test and either confirm or explain why the code is needed if the reading list never displays the external icon?

And maybe it would be easier to create a custom component if we really need to use the code 3 times.

luca-schlecker:
Reintroduced untested double-click ability, like discussed in the meeting. Also rebased onto `develop`, reintroducing the phased out tests.

Merseleo:
Please turn on the test "When navigating to the project paper view, then the page is displayed" before merging. This test should be fixed with your solution as well and used again.

mowi12 (changes requested):
This looks good, but please answer to my last review and @Merseleo's request.

luca-schlecker:
> Please turn on the test "When navigating to the project paper view, then the page is displayed" before merging. This test should be fixed with your solution as well and used again.

Removed the `fixme` from this test :+1:, thank you for the hint.

luca-schlecker:
> Perfect, thank you! This is the exact behaviour I wanted 👍🏼
> 
> Regarding the teamscale warnings - I agree, this is explicit, but maybe simply add some comments, so this is clear for everyone who looks at the code?
> 
> After looking closer at the code, I'm not quite sure about some things. I checked out `develop` in order to see the old behaviour. I noticed, that there never was any double click behaviour in the reading list, e.g. So my question now is, was this intended? If yes, why didn't the code work, since the `ReadingListEntry` clearly had the code which handled the single/double click code. And why do we add the new code snippet you added in the three components? I tested in the reading list without the addition, and it still works ...
> 
> Could you please test and either confirm or explain why the code is needed if the reading list never displays the external icon?
> 
> And maybe it would be easier to create a custom component if we really need to use the code 3 times.

I added comments to these occasions. The reading-list entry was solely adjusted for consistency's sake. I wondered about this too, but I deemed it to be something irrelevant for this PR. Furthermore, I repeated myself because it is a small, limited amount and the other two probably will be obsolete anyway. There was never an issue with the other components, I adjusted it just to be consistent. It stems from the original idea to remove double-click entirely. These components don't even use it as far as I am aware.

mowi12 (changes requested):
This looks good, thanks!

I only have one last question.
I saw, that you added `gap-2` in only one of the modified list entry components. Have you tested, how the other list entries look, if a custom `onclick` function is provided?

Don't we need to modify the surrounding div/component as well, or is this already handled by the current CSS?

luca-schlecker:
> This looks good, thanks!
> 
> I only have one last question. I saw, that you added `gap-2` in only one of the modified list entry components. Have you tested, how the other list entries look, if a custom `onclick` function is provided?
> 
> Don't we need to modify the surrounding div/component as well, or is this already handled by the current CSS?

I did not test the others with the link icon shown. As that functionality is currently not used in the frontend, what would you suggest? Test it now, leave it until needed, or just leave it as is as these parts could likely be refactored in the future?

mowi12 (commented):
I think the best way to continue is to simply test this by testing this now.
If the `gap-2` should be there, I think we should add it now, as we are already aware of it.

luca-schlecker:
I now manually verified that all three of these entry components share the same look. Integration tests were already in place.

mowi12 (dismissed):
Great, then lets finish this!

Only some formatting is left : )

luca-schlecker:
> Great, then lets finish this!
> 
> Only some formatting is left : )

Done, also rebased.

mowi12 (approved):

PR #503, opened 2025-08-13
Show forward and backward references in separate tabs

### LINKED ISSUE #484: [Feature]: Show forward and backward references in separate tabs

_No response_

Currently, we show forward and backward references in the same tab in the paper view.

On a smaller screen, you can merely see 2.5 paper list entries per list. This makes the list harder to read, and the vertical space is not used efficiently.

Split both lists onto two separate tabs.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #484 

## What I have made

See the issue for details.

## Checklist

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only reworked the tabs)


src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
```diff
@@ -33,13 +38,15 @@
     }: PaperResearchContextCardProps = $props();
 
     const reviewInfoTab = { value: "1", label: "Review Information" };
-    const referencesTab = { value: "2", label: "Forward/Backward References" };
+    const forwardReferencesTab = { value: "2", label: "Citations" };
```
Merseleo:
```suggestion
    const forwardReferencesTab = { value: "2", label: "Backward References" };
```

For consistency? Same for the name of the list?

Merseleo: Or is there a reason, why you choose this wording?

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
```diff
@@ -81,6 +88,15 @@ Usage:
         {/if}
     </PaperCardContent>
     <PaperCardContent value="2">
-        <ReferencesAndCitationsCardContent {backwardReferencedPapers} {forwardReferencedPapers} />
+        <ReferencesCardContent
+            loadingReferencedPapers={backwardReferencedPapers}
+            title="Citations"
```
Merseleo:
```suggestion
            title="Backward References"
```

tests/integration/paper-components/paper-view/cards/references-and-citations-card-content.test.ts
Merseleo, line 1: Please rename the test to be consistent with the new component name.

Merseleo (changes requested):
A question that only indirectly relate to this PR: Why did we decide to add the forward /  backward references to the right side of the paper view and not to the left side next to information and document? Actually this would create the two groups "paper related stuff" and "review related stuff" (I know we only have one list related to review, so we would remove the tab pane from the right card and always display review information. 

What are your thoughts on this?

src/lib/components/composites/paper-components/paper-view/cards/PaperResearchContextCard.svelte
Slartibartfass2: No, I thought this was clearer, but as discussed, we keep it like that.

Slartibartfass2 (commented):
Thanks for the fast review.

Slartibartfass2:
> A question that only indirectly relate to this PR: Why did we decide to add the forward / backward references to the right side of the paper view and not to the left side next to information and document? Actually this would create the two groups "paper related stuff" and "review related stuff" (I know we only have one list related to review, so we would remove the tab pane from the right card and always display review information.
> 
> What are your thoughts on this?

As discussed, we keep it as is for now. This would require a redesign and more considerations.

Merseleo (approved):
Good to go 🚀

PR #512, opened 2025-09-05
Fix some flaky tests in the CI

### LINKED ISSUE #511: [Bug]: Fix some flaky tests in the CI

_No response_

### What happened?

When the CI runs in a pull request, then the jobs executing the E2E tests sometimes fails due to flaky tests. 

### Steps to reproduce

1. Open some PR (e.g. #506)
2. Check the pipeline status, especially of the E2E tests

### What did you expect?

The CI should not fail but run successfully.

### Anything else?

Some flaky tests maybe could be fixed by using better locators or check whether some other best practices of playwright were not used. 
However, I could be in general advisable to not fail the CI when some test is flaky, but only provide a warning or something similar.

Merseleo:
Closes #511 

## What I have made

- I fixed some tests that were noticed to be flaky in the pipeline or locally
- the fixed include:
  - do not create papers and add them as project papers using `Promise.all()` but by looping over the promises and manually await all of them so it can be guaranteed that the order in the array is the same as the execution order of the promises
  - adding the helper `openUserMenu` that not only clicks the user avatar to open the user menu, but also waits for the navigation links to be stable as this seemed to be problem (for whatever reason, because Playwright should actually wait, but it doesn't give a damn)
  - use the function `getByTestId` instead of locating the paper details card using the css locator (which is not recommended)

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~
  - ~~[ ] unit tests~~
  - ~~[ ] integration tests~~
  - [x] end-to-end tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,9 +9,16 @@ Closes <!-- issue reference -->
 
 Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does not apply.
 
+### Author
+
 - [ ] I have updated the documentation accordingly and commented my code
+- [ ] I manually tested the changes
 - [ ] I have added tests that prove my fix is effective or that my feature works
   - [ ] unit tests
   - [ ] integration tests
   - [ ] end-to-end tests
-- [ ] (for reviewer) I have checked the implementation against the requirements
+
+### Reviewer
+
+- [ ] I have checked the implementation against the requirements
+- [ ] I have checked for flaky tests in the CI
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -4,7 +4,8 @@ export default defineConfig({
     testDir: "tests/e2e",
     forbidOnly: !!process.env.CI,
     retries: 2,
-    failOnFlakyTests: !!process.env.CI,
+    // decision in team: do not fail on flaky tests in CI during development
+    failOnFlakyTests: false,
     // Opt out of parallel tests on CI.
     workers: process.env.CI ? 1 : undefined,
     reporter: [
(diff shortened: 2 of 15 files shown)


.github/pull_request_template.md
```diff
@@ -15,3 +15,4 @@ Either tick or cross out the items that do not apply (using \~\~example text\~\~
   - [ ] integration tests
   - [ ] end-to-end tests
 - [ ] (for reviewer) I have checked the implementation against the requirements
+- [ ] (for reviewer) I have checked for flaky tests in the CI
```
Slartibartfass2: I propose to restructure this checklist:

```md
## Checklist

### Author

- [ ] I have updated the documentation accordingly and commented my code
- [ ] I manually tested the changes
  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests

### Reviewer

```

What are your thoughts?

playwright.config.ts
```diff
@@ -4,7 +4,8 @@ export default defineConfig({
     testDir: "tests/e2e",
     forbidOnly: !!process.env.CI,
     retries: 2,
-    failOnFlakyTests: !!process.env.CI,
+    // decision in team: do not fail on flaky tests in CI during development
+    // failOnFlakyTests: !!process.env.CI,
```
Slartibartfass2: Is `false` the default value? The documentation doesn't tell. It would probably make sense to define it either way to make it more clear.

Slartibartfass2 (changes requested):
Great changes! Only two things to discuss.

.github/pull_request_template.md
Merseleo: Great idea 👍

playwright.config.ts
Merseleo: I thought all options that are not explicitly set are implicitly set to the default value of the type, in case of boolean to `false` but I agree, that it could be clearer to set it to false instead of not defining it at all.

Merseleo (commented):
Thank you for the hint for the pull request template, very good idea.

Slartibartfass2 (approved):

PR #520, opened 2025-09-21
Feat/519 grpc dev tools

### LINKED ISSUE #519: [Feature]: gRPC Dev Tools

_No response_

As a developer, I would like to have better tooling for debugging gRPC calls in the frontend.

I would like to use [gRPC Web Devtools](https://github.com/SafetyCulture/grpc-web-devtools).
It has extensions for [Firefox](https://addons.mozilla.org/en-US/firefox/addon/grpc-web-developer-tools/) and [Chrome](https://chromewebstore.google.com/detail/grpc-web-developer-tools/cacoibopgjlodngfokahhkphgcohakai).

Image

Using grpcurl and grpcui can be helpful but does not help comprehending api calls in the frontend.

### Anything else?

_No response_

luca-schlecker:
Closes #519 

## What I have made

- Add an interceptor feeding the gRPC Web Devtools browser extension (only in dev mode)

### Author

  - [ ] unit tests
  - [ ] integration tests
  - [ ] end-to-end tests

### Reviewer

### DIFF
--- a/src/lib/grpc-devtools.ts
+++ b/src/lib/grpc-devtools.ts
@@ -1,0 +1,91 @@
+// Modified version of the solution from this comment on github:timostamm/protobuf-ts:
+// https://github.com/timostamm/protobuf-ts/issues/584#issuecomment-1735664168
+
+import { RpcError, type RpcInterceptor } from "@protobuf-ts/runtime-rpc";
+
+declare global {
+    interface Window {
+        __GRPCWEB_DEVTOOLS__?: boolean;
+    }
+}
+
+const type = "__GRPCWEB_DEVTOOLS__";
+
+export const grpcWebDevToolsInterceptor: RpcInterceptor = {
+    interceptUnary(next, method, input, options) {
+        const res = next(method, input, options);
+        if (window.__GRPCWEB_DEVTOOLS__) {
+            const methodType = "unary";
+            const methodName = `${method.service.typeName}/${method.name}`;
+            const request = method.I.toJson(res.request);
+            void res
+                .then((value) => {
+                    window.postMessage({
+                        type,
+                        method: methodName,
+                        methodType,
+                        request,
+                        response: method.O.toJson(value.response),
+                    });
+                })
+                .catch((e) => {
+                    if (e instanceof RpcError) {
+                        window.postMessage({
+                            type,
+                            method: methodName,
+                            methodType,
+                            request,
+                            error: {
+                                code: e.code,
+                                message: e.message,
+                            },
+                        });
+                    }
+                });
+        }
+        return res;
+    },
+    interceptServerStreaming(next, method, input, options) {
+        const res = next(method, input, options);
+        if (window.__GRPCWEB_DEVTOOLS__) {
+            const methodName = `${method.service.typeName}/${method.name}`;
+            const methodType = "server_streaming";
+            window.postMessage({
+                type,
+                method: methodName,
+                methodType,
+                request: method.I.toJson(res.request),
+            });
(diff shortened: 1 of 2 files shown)


src/lib/grpc-devtools.ts
```diff
@@ -0,0 +1,87 @@
...
+// https://github.com/timostamm/protobuf-ts/issues/584#issuecomment-1735664168
+
+import { RpcError, type RpcInterceptor } from "@protobuf-ts/runtime-rpc";
+
+const type = "__GRPCWEB_DEVTOOLS__";
+
+export const grpcWebDevToolsInterceptor: RpcInterceptor = {
+    interceptUnary(next, method, input, options) {
+        const res = next(method, input, options);
+        // @ts-expect-error Variable injected by grpc-web-devtools in the browser
```
mowi12: I know that this is the exactly solution posted in the comment you mentioned, but this doesn't always mean that it is the best.

To simply ignore the error is a solution but there is another way.

```
declare global {
    interface Window {
        __GRPCWEB_DEVTOOLS__?: boolean;
    }
}
```

By declaring this we can extend the global Window object without causing conflicts in the global namespace.

This snippet can either go in this file or into a more appropriate space like `src/types/global.d.ts` or any other file where we already have types (but I don't think we have a file already).

src/lib/grpc-devtools.ts
```diff
@@ -0,0 +1,87 @@
...
+                            },
+                        });
+                    }
+                });
+        }
+        return res;
+    },
+    interceptServerStreaming(next, method, input, options) {
+        const res = next(method, input, options);
+        // @ts-expect-error Variable injected by grpc-web-devtools in the browser
```
mowi12: By applying the solution above this comment can also be removed.

src/lib/grpc-devtools.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const type = "__GRPCWEB_DEVTOOLS__";
+
+export const grpcWebDevToolsInterceptor: RpcInterceptor = {
+    interceptUnary(next, method, input, options) {
+        const res = next(method, input, options);
+        // @ts-expect-error Variable injected by grpc-web-devtools in the browser
+        if (window.__GRPCWEB_DEVTOOLS__) {
+            const methodType = "unary";
+            const m = `${method.service.typeName}/${method.name}`;
```
mowi12:
```suggestion
            const method = `${method.service.typeName}/${method.name}`;
```

I saw something in the comment you referenced about the length of this variable. Is there a problem naming it `method`? Or what exactly is the reason to only call it `m`?

mowi12 (changes requested):
This looks great! Already tested it and found no issues with it.

Just some ideas and questions about the code, but then this can be merged quite quickly.

src/lib/grpc-devtools.ts
luca-schlecker: I think putting it at the top of the `grpc-devtools.ts` file is where it belongs. I've added it and removed the ts-expect-error comments.

src/lib/grpc-devtools.ts
luca-schlecker: `method` cannot be used as it is already a parameter name. I've named it `methodName`, which suits it quite well imo.

src/lib/grpc-devtools.ts
mowi12: Perfect. I also think this can stay in this file.

mowi12 (approved):
Great! You only need to take a look at the Teamscale findings, but then we are good to go!

PR #523, opened 2025-09-24
Add accept project invitation route

### LINKED ISSUE #522: [Feature]: Add Accept Project Invitation Route

_No response_

The backend successfully implements the `AcceptProjectInvitation` call. The frontend not.

Add a new route which handles the accepting of project invitations.

_No response_

### Anything else?

_No response_

mowi12:
Closes #522 

## What I have made

- Added new route `acceptprojectinvitation` route
- Removed obsolete `invitations` route and menu item from user menu

## Checklist

### Author

  - [ ] ~~unit tests~~
  - [ ] ~~integration tests~~
  - [x] end-to-end tests

### Reviewer


src/routes/(auth)/acceptprojectinvitation/+page.ts
```diff
@@ -0,0 +1,19 @@
+import { backendService } from "$lib/grpc-api";
+import type { PageLoad } from "./$types";
+
+export const load: PageLoad = async ({ url }) => {
+    const token = url.searchParams.get("token");
+
+    if (!token) {
+        return {
+            acceptancePromise: Promise.reject({
+                headline: "Accepting Failed",
```
Merseleo:
```suggestion
                headline: "Accepting the Invitation Failed",
```

src/routes/(auth)/acceptprojectinvitation/+page.ts
Merseleo, line 1: Maybe we already discussed this, but why do you do not check whether the user is logged in? Does it make sense to accept an invitation without being logged in?

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
+<script lang="ts">
+    import { goto } from "$app/navigation";
+    import { Button } from "$lib/components/primitives/button";
+    import type { RpcError } from "@protobuf-ts/runtime-rpc";
+    import { GrpcStatusCode } from "@protobuf-ts/grpcweb-transport";
+    import { getGrpcStatusCode } from "$lib/utils/common-helper";
+
+    let { data } = $props();
+    const { acceptancePromise: acceptancePromise } = data;
```
Merseleo:
```suggestion
    const { acceptancePromise } = data;
```
Why not? You are not renaming it?!

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+    type ErrorDetails = { headline: string; body: string } | RpcError;
```
Merseleo: Actually, I noticed that this interface is known to me and I am right: please check the `ApiError` interface, because I think this is exactly the same. 
And maybe you can replace this `{....}` with this interface (same in the invitation page).

Merseleo: And maybe a name like `InvitationAcceptanceError` or something similar would be better? Because actually you use the type `ErrorDetails` to call a function named `getErrorDetails` which is a bit confusing.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+            return customError;
+        }
+
+        const rpcError = error as RpcError;
+        const statusCodeValue = getGrpcStatusCode(rpcError.code);
+
+        switch (statusCodeValue) {
+            case GrpcStatusCode.INVALID_ARGUMENT:
+                return {
+                    headline: "Accepting Failed",
```
Merseleo: Maybe something like "Accepting the Invitation Failed" is better? Because the memory of the user is limited and after 2 seconds he wonders what acceptance failed.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+                    body: "This Acceptance link is invalid. Please check the link and try again.",
```
Merseleo:
```suggestion
                    body: "This acceptance link is invalid. Please check the link and try again.",
```
(same for other bodies)

src/routes/(auth)/acceptprojectinvitation/+page.svelte
Merseleo, line 1: Please check the possible errors again as I changed something in my review (e.g. `DEADLINE_EXCEEDED` should not be used for this purpose, but `NOT_FOUND`) and adapt the error messages displayed to the user.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+                };
+            case GrpcStatusCode.DEADLINE_EXCEEDED:
+                return {
+                    headline: "Accepting Failed",
+                    body: "The Acceptance link has expired. Please check the link and try again.",
```
Merseleo: I don't know whether trying again is the best solution when the link has expired. Maybe the better suggestion is "Please request another invitation" or so?

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+                    body: "Please register before accepting the invitation.",
+                };
+            default:
+                console.error(
+                    "Unexpected error occurred during project invitation acceptance:",
+                    rpcError,
+                );
+                return {
+                    headline: "Server Error",
+                    body: "We couldn't connect to our servers to accept you ivnitation. Please check your internet connection and try again.",
```
Merseleo:
```suggestion
                    body: "We couldn't connect to our servers to accept your invitation. Please check your internet connection and try again.",
```

src/routes/(auth)/acceptprojectinvitation/+page.svelte
```diff
@@ -0,0 +1,84 @@
...
+            sign-in page shortly.
+        </div>
+
+        <Button class="mt-6" onclick={() => goto("/signin")}>Back to Sign In</Button>
+    {:catch error}
+        {@const errorDetails = getErrorDetails(error)}
+        <h1 class="mb-4 text-8xl">{errorDetails.headline}</h1>
+        <div class="text-default">{errorDetails.body}</div>
+
+        <Button class="mt-6" onclick={() => goto("/signup")}>Back to Sign Up</Button>
```
Merseleo, line 91: Why back to sign up and not sign in if an error occurred?

tests/e2e/auth/acceptprojectinvitation/acceptprojectinvitation-page.test.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from "@playwright/test";
+import { test } from "./acceptprojectinvitation-page-fixture";
+
+test.use({ user: null });
+
+test.describe("Accept Project Invitation Verification Navigation", () => {
```
Merseleo:
```suggestion
test.describe("Accept Project Invitation Navigation", () => {
```
Otherwise this sounds very complicated.

tests/e2e/auth/acceptprojectinvitation/acceptprojectinvitation-page.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        page,
+        acceptProjectInvitationPage,
+    }) => {
+        await page.goto("/acceptprojectinvitation");
+        await expect(acceptProjectInvitationPage.heading).toBeVisible();
+    });
+});
+
+test.describe("Accept Project Invitation Tests", () => {
+    test("When navigating to the accept project invitation page with no token, then the page is displayed and the correct message displayed", async ({
```
Merseleo:
```suggestion
    test("When navigating to the accept project invitation page with no token, then the page and 'missing a token' error message is displayed", async ({
```
What do you think?

Merseleo (changes requested):
I looks correct and I do not think that there will be any problems, however I would like to test it tomorrow (maybe even after your changes). Thank you for implementing this as it brings us one step closer to a usable product 👍 

I noticed a flaky test in the CI, but I think it is not related to your changes. Do you think we should investigate this or ignore it?

And as you copied a lot of code from the verify email code, please check possible errors there as well please.

Merseleo:
Ok, I did not figure out how to send emails locally, but I copied the invitation tokens from the local database (which should be actually the same, as the email sending works on the deployed backend) and it worked as expected. 
However, I noticed two things: 1) when I successfully accepted an invitation and was logged in, then I get redirected to the homepage which is fine, but why the button has the text "Go back to Sign In"? Because I wasn't on the sign in page before.
2) You caused an unhandled promise rejecting, probably because of 
```typescript
acceptancePromise?.then(() => {
        setTimeout(async () => await goto("/signin"), 3000);
    });
``` because you only redirect in the success case and if an error occurred, then this promise is unhandled, which is not good. Either you navigate in both cases or you add a catch case here.

src/routes/(auth)/acceptprojectinvitation/+page.ts
mowi12: That was exactly the point. I don't think you need to be logged in to accept the invitation. Is there any reason why you should be logged in?
Maybe it's another barrier to check if you're actually you, but since the email can only be one you own, since it already needs to be verified, there is no reason.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
mowi12: Great idea! It should look a lot cleaner now!

src/routes/(auth)/acceptprojectinvitation/+page.svelte
mowi12: Yeah, why not. Can't hurt to display more information.

mowi12: I've also updated the wording for the email verification page.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
mowi12: I don't know any more, but if I had to think about it, if there is an error, then the most likely reason would be that the person is not registered. So going to signup would be better, or not?

src/routes/(auth)/acceptprojectinvitation/+page.svelte
mowi12: I also updated the verify email page to use the `NOT_FOUND` code instead.

mowi12:
> 1) when I successfully accepted an invitation and was logged in, then I get redirected to the homepage which is fine, but why the button has the text "Go back to Sign In"? Because I wasn't on the sign in page before.

What else should I put there? The user could come from multiple different pages, there is no easy way to know from where.
So the easiest solution is to redirect to sign in, since the user will either log in, or be automatically redirected to the homepage if already logged in. I don't think it will be feasable in the current time to implement a redirect to the currently accepted project e.g.


> 2) You caused an unhandled promise rejecting, probably because of
> 
> ```ts
> acceptancePromise?.then(() => {
>         setTimeout(async () => await goto("/signin"), 3000);
>     });
> ``` because you only redirect in the success case and if an error occurred, then this promise is unhandled, which is not good. Either you navigate in both cases or you add a catch case here.
> ```

I added a catch clause so this won't happen again.

I think I got everything, take a look and maybe you have more luck with testing the email sending this afternoon.

src/routes/(auth)/acceptprojectinvitation/+page.ts
Merseleo: Right, I was dumb.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
Merseleo: Depends. If the token has expired (which can happen as well), then sign in would probably better. But I agree that the case that the user isn't registered yet is more likely. Let's keep it for now.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
Merseleo: Great, thank you.

src/routes/(auth)/acceptprojectinvitation/+page.svelte
Merseleo: Great 👍

Merseleo (approved):
Ok lets check this on the deployed backend.

PR #525, opened 2025-09-30
Archive projects and block actions for archived projects

### LINKED ISSUE #22: [Use Case]: Archive Project

Project Admin

### Describe the use case

As project admin, I want to archive a project.

### Acceptance criteria

- an archived project is not listed in the project list on the landing/home page
- archived project is considered read-only

### Anything else?

Part of #15 

Extended by #23

### LINKED ISSUE #480: [Bug]: Block certain actions in archived projects

#22

### What happened?

When you visit the homepage and started the frontend with the mock backend, there is a chance that you see archived projects. When you open such an archived projects, there is no difference to an active project, which is not correct, as an archived project should be read-only and allow only the action to "re-activate it".

### Steps to reproduce

1. Start the mock backend with a high probability that a user participates in an archived project
2. Go on the homepage of the frontend
3. Go into the archived project

### What did you expect?

I would expect, that I can view everything from the project, but all actions are blocked / disabled except for on button or switch to activate the project again.

### Anything else?

_No response_

Merseleo:
Closes #480
Closes #22 

## What I have made

- I added a settings component `ArchiveProjectSettings` that allows a user to archive an active project and activate a project that is currently archived
- I updated the `buildFieldMask` utility function to optionally set fields that should be excluded from the field mask, as the `id` field should not be contained in the most field masks for example
- I refactored the project settings by extracting identical +page.ts load functions in a single +layout.ts load function for the project settings and by improving the check, whether the current user is project admin
- I removed the page for archived projects
- I converted the `clickOutsideOrEscape` action into the attachment
- I created a `disableIfProjectIsArchived` attachment (not used, idea was to use this for automatically disabling all interactive elements in the project subpages, but our implementation is not ready for such a thing and the attachments still seems to be buggy)
- I refactored the custom contexts and moved all of them into a separate directory
- I added a new context storing whether a project is archived or not
  - All "interactive" components, mainly on the project paper view and project settings, are disabled, whenever the boolean value in this context is true, i.e., the project is currently disabled
- I added a "Archived" badge to the project navigation bar, if the project is archived

Closed #560 as this branch is rebased on that PR.

## Checklist

### Author

  - [ ] ~~unit tests~~
  - [x] integration tests
  - [x] end-to-end tests

### Reviewer


luca-schlecker (dismissed):
This is looking good and fixes the issues I had with the original PR #560. I'd propose just merging this and closing the other one, adding a notice to this to also close the other issue.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2:
```suggestion
        {#if allowEditModeToggle && !isProjectArchived}
```
This way you can remove all the added conditions below. Just don't show them the actions.

Slartibartfass2 (changes requested):
Since your reviewer is currently unavailable, I took a look, and this is great work!
I noticed two missing things where the user shouldn't be allowed to execute an action when the project is archived:
1. Adding a paper to a stage on the project's paper page. (The button should be disabled)
2. Adding highlighted keywords, although one could argue that since this is only stored locally, this has nothing to do with an archived project. (That one is up to you).

Edit: Also, please either correct the Teamscale findings, mark them as false positives or as tolerated.

Merseleo:
> Since your reviewer is currently unavailable, I took a look, and this is great work! I noticed two missing things where the user shouldn't be allowed to execute an action when the project is archived:
> 
> 1. Adding a paper to a stage on the project's paper page. (The button should be disabled)

It is quite hard to disable this "button", as it actually isn't a button but (given the `href` attribute) a `<a>` tag which cannot be disabled. We could alternatively simulate it but not showing the button at all seems the easiest way. What do you think?

> 2. Adding highlighted keywords, although one could argue that since this is only stored locally, this has nothing to do with an archived project. (That one is up to you).

Here I do not quite understand why this should be disabled? I thought this is a helper for finding / reading papers and has nothing todo with the project status in my opinion. Actually I thought we only want to block actions that alter something in the project, but this is only a not-synced, personal visualization helper. So I wouldn't disable this.

Slartibartfass2 (approved):
> It is quite hard to disable this "button", as it actually isn't a button but (given the `href` attribute) a `<a>` tag which cannot be disabled. We could alternatively simulate it but not showing the button at all seems the easiest way. What do you think?

Not showing is fine for me.

> Here I do not quite understand why this should be disabled? I thought this is a helper for finding / reading papers and has nothing todo with the project status in my opinion. Actually I thought we only want to block actions that alter something in the project, but this is only a not-synced, personal visualization helper. So I wouldn't disable this.

Yes, that's why I said it's debatable. One could find it strange that some actions in a project are allowed and some not, but I think it's fine as is. User reports would be helpful for something like that.

PR #539, opened 2025-10-22
Highlight matching patterns in search results

### LINKED ISSUE #360: [Feature]: Add highlighting to search results

#37

When the user either searches for a paper inside the project papers view or in the reading list, there is no highlighting which parts of the title, etc. are matched.

The parts, that are matched by the fuzzy-search should be highlighted, so the user understands, why this search result is at the top, etc.

_No response_

### Anything else?

_No response_

domienderle:
Closes #360 

## What I have made

Implemented the highlighting of search results on the reading list page. 

## Checklist

### Author

  - ~[ ] unit tests~ not necessary
  - ~[ ] integration tests~ not necessary
  - [x] end-to-end tests

### Reviewer

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,8 @@
                 "fetch-cookie": "^3.1.0",
                 "fzf": "^0.5.2",
                 "globals": "^16.5.0",
+                "highlight-words": "2.0.0",
+                "globals": "^16.5.0",
                 "jsdom": "^26.1.0",
                 "license-checker": "^25.0.1",
                 "lucide-svelte": "^0.544.0",
@@ -5611,6 +5613,17 @@
             },
             "engines": {
                 "node": ">= 0.4"
+            }
+        },
+        "node_modules/highlight-words": {
+            "version": "2.0.0",
+            "resolved": "https://registry.npmjs.org/highlight-words/-/highlight-words-2.0.0.tgz",
+            "integrity": "sha512-If5n+IhSBRXTScE7wl16VPmd+44Vy7kof24EdqhjsZsDuHikpv1OCagVcJFpB4fS4UPUniedlWqrjIO8vWOsIQ==",
+            "dev": true,
+            "license": "MIT",
+            "engines": {
+                "node": ">= 20",
+                "npm": ">= 9"
             }
         },
         "node_modules/hosted-git-info": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,6 @@
                 "fzf": "^0.5.2",
                 "globals": "^16.5.0",
                 "highlight-words": "2.0.0",
-                "globals": "^16.5.0",
                 "jsdom": "^26.1.0",
                 "license-checker": "^25.0.1",
                 "lucide-svelte": "^0.544.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
                 "fetch-cookie": "^3.1.0",
                 "fzf": "^0.5.2",
                 "globals": "^16.5.0",
-                "highlight-words": "2.0.0",
+                "highlight-words": "^2.0.0",
                 "jsdom": "^26.1.0",
                 "license-checker": "^25.0.1",
                 "lucide-svelte": "^0.544.0",
--- a/src/app.css
+++ b/src/app.css
@@ -225,3 +225,9 @@
         @apply text-primary text-lg font-semibold tracking-normal;
     }
 }
+
+.highlight {
+    border-radius: 2px;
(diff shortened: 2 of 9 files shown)


package-lock.json
```diff
@@ -37,6 +37,8 @@
                 "fetch-cookie": "^3.1.0",
                 "fzf": "^0.5.2",
                 "globals": "^16.5.0",
+                "highlight-words": "2.0.0",
+                "globals": "^16.5.0",
```
Slartibartfass2:
```suggestion
                "globals": "^16.5.0",
                "highlight-words": "2.0.0",
```
globals is duplicated here. Just run `npm i` to fix it.

src/lib/components/composites/paper-components/PaperInfo.svelte
```diff
@@ -34,11 +37,25 @@ for a project paper, the local / relative project paper id.
...
         <div class="text-hint flex flex-row items-center truncate">
             {#if paper.authors.length > 0}
-                <span class="place-content-start truncate">{getNames(paper.authors)}</span>
+                {#each highlightWords( { text: getNames(paper.authors), query: safeSearchQuery, matchExactly: false }, ) as chunk (chunk.key)}
+                    {#if chunk.match}
+                        <span class="highlight">{chunk.text}</span>
+                    {:else}
+                        {chunk.text}
+                    {/if}
+                {/each}
```
Slartibartfass2: This still has to be in the span like before.
```suggestion
                <span class="place-content-start truncate">
                    {#each highlightWords( { text: getNames(paper.authors), query: safeSearchQuery, matchExactly: false }, ) as chunk (chunk.key)}
                        {#if chunk.match}
                            <span class="highlight">{chunk.text}</span>
                        {:else}
                            {chunk.text}
                        {/if}
                    {/each}
                </span>
```

src/routes/readinglist/+page.svelte
```diff
@@ -7,14 +7,33 @@
...
+            callDebounced(
+                () =>
+                    goto(`?${searchParameters.toString()}`, {
+                        replaceState: true,
+                        keepFocus: true,
+                    }),
+                250,
+            );
+        }
+    });
```
Slartibartfass2, line 29: What is the purpose of this effect?

src/app.css
```diff
@@ -225,3 +225,9 @@
         @apply text-primary text-lg font-semibold tracking-normal;
     }
 }
+
+.highlight {
+    border-radius: 2px;
+    background-color: yellow;
+    color: black;
+}
```
Slartibartfass2: Use Tailwind syntax for this:
```suggestion
@utility highlight {
    @apply text-black bg-amber-300;
}
```
I omitted the border radius in this example and used amber instead of yellow because I thought it was more comfortable to read than just bright yellow.

Slartibartfass2 (changes requested):
Great work. I really like this feature. There are some minor questions/changes left open.

Edit: You can also rebase right away. There's a minor conflict, which should be no problem.

src/lib/components/composites/paper-components/PaperInfo.svelte
domienderle: Good catch.

src/routes/readinglist/+page.svelte
domienderle: This is used to update the search parameters in the url. It recognizes a difference between the search parameters and the url and updates the url and the location. But since we use this in `/project/[projectId]/papers/+page.svelte` too I refactored this to a helper method now.

src/app.css
domienderle: Good catch. I have never changed this since my initial tries. I'm fine with that.

domienderle (commented):
Thanks for the review.

src/routes/readinglist/+page.svelte
Slartibartfass2: Great refactoring!

Slartibartfass2 (changes requested):
This is basically done. There are still changes when I run `npm install`, can you reproduce this?

Also, usually the dependencies are ordered alphabetically. This is done automatically when using npm in the command line. It appears that you have been doing this manually. I recommend removing and adding the `highlight-words` dependency again, i.e., by executing the following commands:
```sh
npm remove highlight-words
npm install highlight-words -D
```
This way, the order is restored, and the package-lock.json is also correctly set.

After that, set this on auto-merge so we can merge it right away.

domienderle:
All right. I have removed and added the highlight-words again and have run `npm install`. There should not be any difference anymore.

Slartibartfass2 (approved):

Issue #513, opened 2025-09-09
[Feature]: Add more sophisticated input validation for edit paper details


Slartibartfass2:
### Is your feature request related to a problem? Please describe

In #61, we added basic input validation for updated paper details. This should be more sophisticated.

### Describe the solution you'd like

Same validation as in the backend.

Agreement from last meeting:
> authors are separated by semicolon and if a comma appears, then it is interpreted as last name, first name, otherwise as first name last name and add a help behind authors explaining the syntax

Edit: Only enable the save button when paper data is valid (also show a hint why the button is disabled).

Merseleo:
Alongside with an improved input validation, the user should get more information on how to properly input paper details by adding help tooltips next to the paper details. These tooltips should explain, what inputs are allowed, what is not allowed and maybe what consequences a change have.

Slartibartfass2:
Before we implement this, I would like the input elements to be refactored/improved. Currently, we use a text area for all inputs, but it really only makes sense for the abstract. I would like to use our input components for that and toggle between spans and those input elements. When using input elements, we can then also apply Zod schemas for input validation.

mowi12:
> Alongside with an improved input validation, the user should get more information on how to properly input paper details by adding help tooltips next to the paper details. These tooltips should explain, what inputs are allowed, what is not allowed and maybe what consequences a change have.

In addition, we need to add good placeholders.
These should be valid by the help-suggestions

PR #540, opened 2025-10-22
Changed how authors are represented and parsed

### LINKED ISSUE #536: [Feature]: Change how authors are represented and parsed

_No response_

Adding papers is more complex than we expected; now we have two calls for that.

Use the new calls and implement the logic.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #536 

## What I have made

Similar to the changes in the API and the backend, this PR is an adjustment to the authors being no entities anymore.
The author list is now represented in the format we discussed, and a hint is shown on how the format works.

## Checklist

### Author

  - [x] unit tests

### Reviewer

### DIFF
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -5,8 +5,10 @@
     import ToggleableInput from "$lib/components/composites/input/ToggleableInput.svelte";
     import type { HTMLAttributes } from "svelte/elements";
     import type { Paper } from "$lib/model/api/paper";
-    import ErrorIndicator from "../../utils/ErrorIndicator.svelte";
+    import ErrorIndicator from "$lib/components/composites/utils/ErrorIndicator.svelte";
     import type { StringifiedPaper } from "$lib/model/general";
+    import CircleHelp from "lucide-svelte/icons/circle-help";
+    import Tooltip from "$lib/components/composites/utils/Tooltip.svelte";
 
     export interface PaperDetailProp {
         key: keyof StringifiedPaper;
@@ -19,9 +21,17 @@
         loadingPaper: Promise<Paper>;
         paper: StringifiedPaper;
         isInEditMode: boolean;
+        hint?: string;
     };
 
-    let { prop, index = 0, loadingPaper, paper = $bindable(), isInEditMode }: Props = $props();
+    let {
+        prop,
+        index = 0,
+        loadingPaper,
+        paper = $bindable(),
+        isInEditMode,
+        hint = undefined,
+    }: Props = $props();
     const { key, label } = prop;
 
     const skeletonValues = [
@@ -61,13 +71,31 @@ Usage:
             />
         </div>
     {:then}
-        <ToggleableInput
-            isEditable={isInEditMode}
-            {key}
-            onInputChange={updateValue}
-            placeholder={`No ${label} available`}
-            value={paper[key]}
-        />
+        <div class="flex w-full flex-row gap-1">
+            <ToggleableInput
+                isEditable={isInEditMode}
+                {key}
+                onInputChange={updateValue}
+                placeholder={`No ${label} available`}
+                value={paper[key]}
+            />
+            {#if hint && isInEditMode}
+                <Tooltip
+                    class="p-0 [&_svg]:size-6"
+                    aria-label={hint}
+                    openOnClick
+                    triggerSize="fit"
+                    triggerVariant="none"
+                >
+                    {#snippet trigger()}
+                        <CircleHelp class="text-neutral-500" />
+                    {/snippet}
+                    {#snippet content()}
+                        {hint}
+                    {/snippet}
+                </Tooltip>
+            {/if}
+        </div>
     {:catch}
         <ErrorIndicator errorMessage={`Couldn't load ${label}`} />
     {/await}
--- a/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
@@ -60,7 +60,25 @@ Usage:
 -->
 <div id={key} class="flex flex-row gap-2" data-testid="paper-detail">
     <!-- Match top padding of input -->
-    <span class="w-24 pt-[0.3125rem] xl:w-42">{label}</span>
+    <span class="flex w-24 flex-row items-center gap-2 pt-[0.3125rem] xl:w-42">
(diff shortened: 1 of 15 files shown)


src/lib/utils/common-helper.ts
Merseleo, line 1: This `common-helper` file got quite big, what do you think of refactoring it and split our utils functions in a separate PR? I thought about a different structure together with AI and I think there is big potential.

tests/unit/utils/common-helper.test.ts
```diff
@@ -276,3 +277,118 @@ describe("Check String equality", () => {
...
+        const text = "Doe, John; Jane Smith; Brown, Bob";
+        const authors = stringToAuthors(text);
+        expect(authors).toEqual([
+            { firstName: "John", lastName: "Doe" },
+            { firstName: "Jane", lastName: "Smith" },
+            { firstName: "Bob", lastName: "Brown" },
+        ]);
+    });
+
+    test("When an author has multiple commas in the name, then only the first comma is used to split first and last names", () => {
```
Merseleo, line 386: Maybe add the note, that this only applies for the format 'Last, First', right?

src/lib/utils/common-helper.ts
```diff
@@ -368,6 +369,52 @@ function isStringEqual(obj1: unknown, obj2: unknown): boolean {
...
+
+/**
+ * Extracts a single author from a string.
+ *
+ * We assume that the author is either in the format "first_name last_name" or "last_name, first_name".
+ *
+ * @param text - The string containing the author.
+ * @returns An Author object.
+ */
+function stringToAuthor(text: string): Author {
```
Merseleo, line 396: Did you tested this with your new backend version?
Because, this methods allows to set the first name to an empty string and then the backend will complain as you defined in the backend, then both first name and last name must be given, so where do you handle this error? Do you have any idea on how to properly show this problem to the user?

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
```diff
@@ -61,13 +71,31 @@ Usage:
...
         </div>
     {:then}
-        <ToggleableInput
-            isEditable={isInEditMode}
-            {key}
-            onInputChange={updateValue}
-            placeholder={`No ${label} available`}
-            value={paper[key]}
-        />
+        <div class="flex w-full flex-row gap-1">
```
Merseleo, line 92: In my opinion this isn't the best layout. Nice addition with the hint (and we have to add the remaining hints later, but not in this PR), but I think it would be better, if the help symbol would be next to the "Authors" text and not the last thing in this row. Is there any special reason why you placed it there?

Merseleo (changes requested):
Nice work, I wish every review would be that easy. No functional problems, probably tested by yourself, no spelling mistakes or missing documentation, edge cases considered, barely something to complain about.

P.S. Same as in the backend. I think you do not updated the API submodule properly as I still get errors?! Probably it would be advantageous to first release a new version of the API, merge this and then update the backend and frontend accordingly? Or am I missing something?

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Slartibartfass2: I thought this made the most sense.
What do you think of this?
Image

I generally believe that this section requires a rework, and I have some ideas.

tests/unit/utils/common-helper.test.ts
Slartibartfass2: There are no commas in 'First Last', so I'm not sure what you mean. Please explain.

src/lib/utils/common-helper.ts
Slartibartfass2: I split the file up into 3 files. What's your opinion on that?

src/lib/utils/common-helper.ts
Slartibartfass2: Yes, I noticed this yesterday, but forgot it today. Since we tell the user to provide a first and last name, we can expect them to follow the format. Otherwise, they cannot update the information.

Slartibartfass2 (commented):
Thanks for the review

src/lib/utils/common-helper.ts
Merseleo: As I said, I already started to refactor the utils functions and split them, so (imo) it is unnecessary start this in this PR and do it in another PR as well, as these changes have nothing to do with this fix (so I would revert these changes). I actually wanted to know, whether you recognized as well that this single helper file has grown a lot, but I guess your answer is yes ^^.

tests/unit/utils/common-helper.test.ts
Merseleo: First I thought, that 'First, Second first Last' is also possible so I was a bit confused, but as I read the hint again, it is clearer. So ignore my comment.

src/lib/utils/common-helper.ts
Merseleo: Yes, we can expect them to follow the format. But if (for whatever reason) they mix something up and input an author "Doe; John, Jonathan" instead of "Doe, John, Jonathan", then you only get a toast "Failed to update paper", but in my opinion then it is not quite obvious what I've done wrong. But we can also extend this is in #513.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Merseleo: I think, your suggestion in this conversation does make more sense as the help symbol is directly behind the label it explains (like in the criteria list). 

I know, but not in this PR, other PRs exist for this.

Merseleo (changes requested):

src/lib/utils/common-helper.ts
Slartibartfass2: Ah sorry, I misread your comment, I dropped the commit.

src/lib/utils/common-helper.ts
Slartibartfass2: Yes, we have to turn the response of the backend into alerts in the frontend.

src/lib/components/composites/paper-components/paper-view/PaperDetail.svelte
Slartibartfass2: Alright, I implemented the change.

src/lib/utils/common-helper.ts
Merseleo: Ok, I guess then we do this in #513

Merseleo (approved):
I think this PR is ready to be merged now.

PR #541, opened 2025-10-26
Add papers to stage

### LINKED ISSUE #35: [Use Case]: Add Papers To Stage

Project Member

### Describe the use case

As project member, I want to add papers to any stage.

### Acceptance criteria

- ability to add papers to any stage (0-n)
- the papers of a stage must be unique
- already existing papers cannot be added
- papers can only be added to the current stage, i.e. not to any previous stages

### Anything else?

Related to #30 and #36

Slartibartfass2:
Closes #35

## What I have made

This PR adds the functionality to manually create a paper inside a project.
The commits are well-structured, so I recommend reviewing this commit-by-commit.

There are two commits that are rather unrelated but were noticed on the way:
- 99520a5635e6bef71c1524d886853ec11f397232: When the project-independent paper view is opened, the whole UUID is shown, which is a lot. This shortens the UUID to its first 8 characters, which is still enough to differentiate them.
- 4c8301ca1365612974c32fa5ebad54cf5a24b68b: I noticed some UX crimes in the fetcher settings (loading buttons without fixed width and loading label, non-disabled buttons while the page was loading, ...). I fixed it.

## Checklist

### Author

  - ~~[ ] unit tests~~ (no "units" to test)
  - [x] integration tests
  - [x] end-to-end tests

### Reviewer

### DIFF
--- a/src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
@@ -14,6 +14,8 @@
     import { isStringEqual, stringToAuthors } from "$lib/utils/common-helper";
     import { beforeNavigate } from "$app/navigation";
     import { buildFieldMask } from "$lib/utils/fieldmask-helper";
+    import { goto } from "$app/navigation";
+    import type { Project_Paper } from "$lib/model/api/project";
 
     interface Props {
         loadingPaper: Promise<Paper>;
@@ -82,11 +84,16 @@
             backwardReferencedIds: paper.backwardReferencedIds.split(/,\s*/g),
         };
 
+        let promise;
         if (isInCreationMode) {
-            await createPaper(Paper.create(paperData));
+            promise = createPaper(Paper.create(paperData));
         } else {
-            await updatePaper(paperData);
+            promise = updatePaper(paperData);
         }
+
+        // Add noop-catch so that promise rejection is not uncaught in tests
+        // noop-catch cannot be added before toast.promise because error case wouldn't be handled
+        await promise.catch(() => {}).finally(() => (isMakingApiCall = false));
     }
 
     async function updatePaper(paperData: Partial<Paper>) {
@@ -106,13 +113,7 @@
             error: "Failed to update the paper.",
         });
 
-        // Add noop-catch so that promise rejection is not uncaught in tests
-        // noop-catch cannot be added before toast.promise because error case wouldn't be handled
-        await updateCall
-            .catch(() => {})
-            .finally(() => {
-                isMakingApiCall = false;
-            });
+        return updateCall;
     }
 
     async function createPaper(paperData: Paper) {
@@ -121,20 +122,27 @@
             .response.then((createdPaper) => {
                 originalPaper = stringifyPaper(createdPaper);
                 paper = stringifyPaper(createdPaper);
-                return createdPaper;
+
+                return backendService.addPaperToProject({
+                    paperId: createdPaper.id,
+                    projectId: "5",
+                    stage: 0n,
+                }).response;
             });
 
         toast.promise(creationCall, {
-            loading: "Creation paper...",
-            success: (paper) => `Successfully created the paper with the ID ${paper.id}.`,
+            loading: "Creating paper...",
+            success: (projectPaper: Project_Paper) =>
+                `Successfully created the paper '${projectPaper.paper?.title}'.`,
             error: "Failed to create the paper.",
         });
 
-        const createdPaper = await creationCall
-            .catch(() => {})
-            .finally(() => {
-                isMakingApiCall = false;
-            });
+        return creationCall.then(async (projectPaper) => {
+            await goto(`/project/5/paper/${projectPaper.localId}`);
+            return projectPaper;
+        });
+
+        // TODO: add to specific project
     }
 
     beforeNavigate(({ cancel }) => {
--- a/src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
+++ b/src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
@@ -19,9 +19,10 @@
         loadingPaper: Promise<Paper>;
         allowEditModeToggle: boolean;
         startInEditMode: boolean;
+        isInCreationMode: boolean;
     }
 
-    let { loadingPaper, allowEditModeToggle, startInEditMode }: Props = $props();
+    let { loadingPaper, allowEditModeToggle, startInEditMode, isInCreationMode }: Props = $props();
 
     let paper: StringifiedPaper = $state(stringifyPaper(Paper.create()));
 
@@ -81,7 +82,11 @@
             backwardReferencedIds: paper.backwardReferencedIds.split(/,\s*/g),
         };
 
-        await updatePaper(paperData);
+        if (isInCreationMode) {
+            await createPaper(Paper.create(paperData));
+        } else {
+            await updatePaper(paperData);
+        }
     }
 
     async function updatePaper(paperData: Partial<Paper>) {
@@ -104,6 +109,28 @@
         // Add noop-catch so that promise rejection is not uncaught in tests
         // noop-catch cannot be added before toast.promise because error case wouldn't be handled
         await updateCall
+            .catch(() => {})
+            .finally(() => {
+                isMakingApiCall = false;
+            });
+    }
+
+    async function createPaper(paperData: Paper) {
+        const creationCall = backendService
+            .createPaper(paperData)
+            .response.then((createdPaper) => {
+                originalPaper = stringifyPaper(createdPaper);
+                paper = stringifyPaper(createdPaper);
+                return createdPaper;
+            });
+
+        toast.promise(creationCall, {
+            loading: "Creation paper...",
+            success: (paper) => `Successfully created the paper with the ID ${paper.id}.`,
+            error: "Failed to create the paper.",
+        });
+
+        const createdPaper = await creationCall
             .catch(() => {})
             .finally(() => {
                 isMakingApiCall = false;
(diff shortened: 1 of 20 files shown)


src/routes/project/[projectId]/paper/new/+page.svelte
```diff
@@ -1,25 +1,13 @@
...
-        hasPdf: true,
-        authors: [
-            { firstName: "John", lastName: "Doe" },
-            { firstName: "Jane", lastName: "Smith" },
-        ],
-        backwardReferencedIds: ["1", "2"],
-    };
+
+    let paper: Paper = Paper.create({
+        year: new Date().getFullYear(),
```
Comment (domienderle) [inline src/routes/project/[projectId]/paper/new/+page.svelte]: Is the most probable year for the new paper the current year? Does it make sense to provide it here?

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
```diff
@@ -110,6 +115,28 @@
             });
     }
 
+    async function createPaper(paperData: Paper) {
+        const creationCall = backendService
+            .createPaper(paperData)
+            .response.then((createdPaper) => {
+                originalPaper = stringifyPaper(createdPaper);
+                paper = stringifyPaper(createdPaper);
+                return createdPaper;
```
domienderle, line 131: Is there any reason why we call `stringifyPaper` two times?

tests/integration/paper-components/paper-view/cards/paper-details-card.test.ts
```diff
@@ -173,103 +174,161 @@ describe("PaperDetailsCard", () => {
...
-        const mockCall = mockApiCall("updatePaper", paper);
-
-        render(PaperDetailsCard, {
-            target: document.body,
-            props: {
-                isInCreationMode: false,
-                loadingPaper: loading(paper),
-                allowEditModeToggle: true,
-                startInEditMode: true,
+    describe.each([
```
domienderle, line 177: This would make a new describe block for each running test? Isn't it better to use `test.each`? Or otherwise move this out of the current describe block.

tests/e2e/project/paper/project-paper-view-page-model.ts
```diff
@@ -80,6 +80,17 @@ export class ProjectPaperViewPageModel {
...
+    /**
+     * Navigates to a create project paper view.
+     *
+     * @param projectId - The id of the project
+     * @param paperId - The local project paper id
+     */
+    async openCreateProjectPaperView(projectId: string, stage: string) {
+        await this.page.goto(`/project/${projectId}/paper/new?stage=${stage}`);
+        await expect(this.savePaperChangesButton).toBeVisible();
+    }
```
domienderle, line 92: Update the documentation correctly: Something like:
```javascript
* @param stage - The stage of the newly created paper
```

tests/e2e/project/paper/project-paper-view-page-model.ts
```diff
@@ -80,6 +80,17 @@ export class ProjectPaperViewPageModel {
...
+
     /**
      * Navigates to the first referenced paper and ensures the page is loaded.
      */
```
domienderle: This isn't related to you PR, but the method `navigateToReferencedPaper` is not used anymore. Please remove it.

domienderle (changes requested):
This looks a lot on the first glance, but overall this was really great work. It was really hard to find something to improve especially in your code. But a couple of questions and change request (especially regarding the behaviour) are still left:

- Following scenario. You click on edit a paper an type in some changes. Then you click on edit again the changes are not saved, but the view looks like it is changed. What do you think of disabling the edit button as soon as something is changed as long as the changes are manually saved? This would be more intuitive for me.
- If I add/edit a paper and initally just remove the year (so all fields are empty) the save button is enabled and I could press it. I would expect, that the save button is only shown if the necessary fields are not empty.
- If a paper with an empty title should be added the backend prevents this and a toast is shown with `Failed to create the Paper` and `Failed to add the paper`. This is theoretically the correct behaviour, but for a better user feedback I would prefer to change the input field for the paper name to don't accept empty inputs (or even better to disable the save button as long as the field is empty)
-  If I click on `AddPaper`, then the nothing gets disabled or loading, indicating, that the click has an effect. Please add a loading state to at least the Add Paper button (maybe also to the filter and sort button) 
- I noticed, that every time we add a paper we also get an error dialog showing `error loading the next paper`. This is a bug which is not directly related to your PR, because it originates from the `nextPaper` and `nextPaperToReview` implementation. If you want to you can fix this, otherwise please create a bug fix issue for this or tell me and I will create it.
- I also noticed, that adding authors with only one name is rejected, but adding authors with multiple names `firstname secondname lastname` is allowed. I guess this is intended, but just to make sure: is it intended behaviour?
- If I remove the year of the paper the default is 0. I have to admit this will be a very rare case, but imo the current year would be a better default value.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: No, not in particular. We can also call it once and then assign the value. I changed it that way.

src/routes/project/[projectId]/paper/new/+page.svelte
Comment (Slartibartfass2) [reply src/routes/project/[projectId]/paper/new/+page.svelte]: Probably not, but which would be? The default value is 0, which is also not very likely. Depending on the scientific field, the most probable year could be 20 to 60 or even 100 years ago.

Another option would be to do something like this: `year: "" as unknown as number`, which is likely the best option.

tests/e2e/project/paper/project-paper-view-page-model.ts
Slartibartfass2: Ah yes, copy-pasting.

tests/integration/paper-components/paper-view/cards/paper-details-card.test.ts
Slartibartfass2: I'm not sure what you mean. The reason behind this is that we can run each test once for the updatePaper call and once for the createPaper call and therefore don't have to duplicate these tests.

Slartibartfass2 (commented):
Thank you for reviewing this PR.

> * Following scenario. You click on edit a paper an type in some changes. Then you click on edit again the changes are not saved, but the view looks like it is changed. What do you think of disabling the edit button as soon as something is changed as long as the changes are manually saved? This would be more intuitive for me.

Good catch, I changed it. For undoing changes, I added [this issue](https://github.com/SE-UUlm/snowballr-frontend/issues/557).

> * If I add/edit a paper and initally just remove the year (so all fields are empty) the save button is enabled and I could press it. I would expect, that the save button is only shown if the necessary fields are not empty.

Yes, I agree. This is part of https://github.com/SE-UUlm/snowballr-frontend/issues/513.

> * If a paper with an empty title should be added the backend prevents this and a toast is shown with `Failed to create the Paper` and `Failed to add the paper`. This is theoretically the correct behaviour, but for a better user feedback I would prefer to change the input field for the paper name to don't accept empty inputs (or even better to disable the save button as long as the field is empty)

Yes, I agree. This is part of https://github.com/SE-UUlm/snowballr-frontend/issues/513.

> * If I click on `AddPaper`, then the nothing gets disabled or loading, indicating, that the click has an effect. Please add a loading state to at least the Add Paper button (maybe also to the filter and sort button)

Can you tell me where? In the paper view, we have a toast, and the save button has a loading state.

> * I noticed, that every time we add a paper we also get an error dialog showing `error loading the next paper`. This is a bug which is not directly related to your PR, because it originates from the `nextPaper` and `nextPaperToReview` implementation. If you want to you can fix this, otherwise please create a bug fix issue for this or tell me and I will create it.

This is because there are no papers to review, right? Should we just show nothing if this occurs?

> * I also noticed, that adding authors with only one name is rejected, but adding authors with multiple names `firstname secondname lastname` is allowed. I guess this is intended, but just to make sure: is it intended behaviour?

Yes, currently it is. Leonhard already mentioned it [here](https://github.com/SE-UUlm/snowballr-frontend/pull/540#discussion_r2454613389), and we decided to solve it in #513.

> * If I remove the year of the paper the default is 0. I have to admit this will be a very rare case, but imo the current year would be a better default value.

Good catch. This is because `Number("")` just returns 0 and not `NaN`. I added an extra check. More on this in #513.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
domienderle: Can you change this for `updatePaper` (line 106) and the `effect` (line 45) as well, please.

tests/integration/paper-components/paper-view/cards/paper-details-card.test.ts
domienderle: Yes you are right. Forget it. I just didn't look good enough.

domienderle (changes requested):
Really good work. There is not really much to add. Just answer the to points below, then we can merge.

>> If I click on AddPaper, then the nothing gets disabled or loading, indicating, that the click has an effect. Please add a loading state to at least the Add Paper button (maybe also to the filter and sort button)

> Can you tell me where? In the paper view, we have a toast, and the save button has a loading state.

It is not a problem of the paper view but of the view where the project papers are shown.

> I noticed, that every time we add a paper we also get an error dialog showing error loading the next paper. This is a bug which is not directly related to your PR, because it originates from the nextPaper and nextPaperToReview implementation. If you want to you can fix this, otherwise please create a bug fix issue for this or tell me and I will create it.

What about this? Did you already create one and I just didn't find it? Otherwise I can create one, just that it is not duplicated.

src/lib/components/composites/paper-components/paper-view/cards/PaperDetailsCard.svelte
Slartibartfass2: Didn't see them, thanks!

Slartibartfass2 (commented):
> It is not a problem of the paper view but of the view where the project papers are shown.

Now I see what you mean. I tested it with a long response delay and saw that it takes a while to navigate. Since this is an a tag, it should navigate instantly, but I think due to preloading, it takes a while to do that. However, nothing that should concern us for now.

> What about this? Did you already create one and I just didn't find it? Otherwise I can create one, just that it is not duplicated.

I already answered this one with some more questions 😅:
> This is because there are no papers to review, right? Should we just show nothing if this occurs?

What are your thoughts about this?

domienderle (changes requested):
Oh sorry, I must have overseen this comment. Yes, I guess it would be the best to just don't show anything, since it is not really an error that occurs if no next or previous paper is shown but intended behaviour. Maybe you could just remove the errors, than we can merge the PR.

Slartibartfass2:
> Oh sorry, I must have overseen this comment. Yes, I guess it would be the best to just don't show anything, since it is not really an error that occurs if no next or previous paper is shown but intended behaviour. Maybe you could just remove the errors, than we can merge the PR.

I removed the error toasts.

domienderle (approved):

Issue #54, opened 2024-10-31
[Use Case]: View Highlighted Keywords


Slartibartfass2:
### Who is the primary actor in this use case?

Project Member

### Describe the use case

As project member, I want to see the keywords highlighted in the abstract.

### Acceptance criteria

- each keyword, which is configured in the project settings, is highlighted in the abstract

### Anything else?

Related to #34

PR #559, opened 2025-11-06
Refactored ToggleableInput component

### LINKED ISSUE #501: [Feature]: Rework `ToggleableInput`

_No response_

Currently, we have the `ToggleableInput` component, which comes with some downsides. We initially tried to put both the editing and the displaying behavior in one HTML element instead of just using a conditional expression. Also, we should differentiate between single input elements and the text area for the abstract.

I propose using a conditional expression for the toggle logic and then splitting the component into a `ToggleableInput` and a `ToggleableTextarea`. The input component can then also accept other parameters such as the `type`.

Using the conditional expression and splitting it up makes it later easier to implement the keyword highlighting.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #558: [Refactor]: Refactor ToggleableInput to use actual input elements

_No response_

Before we add input validation in #513, we need the correct structure to do so.

Refactor the component so that we can use input validation.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #558 
Closes #501

## What I have made

I refactored the component to make input validation later when we implement it in #513. It's also important for #54.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only refactoring)

### Reviewer

### DIFF
--- a/src/lib/components/composites/input/ToggleableInput.svelte
+++ b/src/lib/components/composites/input/ToggleableInput.svelte
@@ -2,6 +2,7 @@
     import type { HTMLInputAttributes } from "svelte/elements";
     import type { WithElementRef } from "bits-ui";
     import { cn } from "$lib/utils/shadcn-helper";
+    import Input from "./Input.svelte";
 
     type Props = WithElementRef<HTMLInputAttributes> & {
         key: string;
@@ -21,7 +22,7 @@
 
 <!--
 @component
-A textarea that can be toggled between read-only and editable mode.
+An input that can be toggled between read-only and editable mode.
 
 The `onInputChange` method is called whenever the input of the input changes.
 
@@ -30,16 +31,27 @@ Usage:
     <ToggleableInput {isEditable} {key} onInputChange={(input) => foo(input)} />
 ```
 -->
-<textarea
-    class={cn(
-        "bg-background text-default min-h-8 w-full resize-none border px-1.5 py-1 focus-visible:outline-hidden",
-        isEditable ? "border-input rounded-md" : "border-transparent",
-        className,
-    )}
-    data-testid={`toggleable-input-${key}`}
-    oninput={(event) => onInputChange((event.target as HTMLTextAreaElement)?.value as string)}
-    {placeholder}
-    readonly={!isEditable}
-    rows="1"
-    {value}
-></textarea>
+{#if isEditable}
+    <Input
+        class="bg-background text-default w-full"
+        data-testid={`toggleable-input-${key}`}
+        inputClass={cn("border-input rounded-md border px-1.5", className)}
+        inputId={`toggleable-input-${key}`}
+        oninput={(event) => onInputChange((event.target as HTMLInputElement)?.value as string)}
+        placeholder={placeholder ?? ""}
+        required
+        type="text"
+        {value}
+    />
+{:else}
+    <div
+        class={cn(
+            "bg-background text-default flex h-10 w-full items-center overflow-hidden px-2",
+            !value ? "text-slate-500" : "",
+            className,
+        )}
+        data-testid={`toggleable-input-${key}`}
+    >
+        <span class="line-clamp-1" title={value}>{value ? value : placeholder}</span>
+    </div>
+{/if}
--- a/src/lib/components/composites/input/ToggleableInput.svelte
+++ b/src/lib/components/composites/input/ToggleableInput.svelte
@@ -46,7 +46,7 @@ Usage:
 {:else}
     <div
         class={cn(
-            "bg-background text-default flex h-10 w-full items-center overflow-hidden px-2",
+            "bg-background text-default flex h-10 w-full items-center overflow-hidden px-[0.475rem]",
             !value ? "text-slate-500" : "",
             className,
         )}
--- a/src/lib/components/composites/input/ToggleableTextArea.svelte
+++ b/src/lib/components/composites/input/ToggleableTextArea.svelte
@@ -41,7 +41,7 @@ Usage:
 {:else}
     <div
         class={cn(
-            "bg-background text-default h-1 min-h-8 w-full border border-transparent px-1.5 py-1",
+            "bg-background text-default h-full w-full overflow-y-hidden border border-transparent px-1.5 py-1",
             !value ? "text-neutral-500/85" : "",
         )}
         data-testid={`toggleable-textarea-${key}`}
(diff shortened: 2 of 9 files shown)


Merseleo:
Then we close #501 as duplicate right?

src/lib/components/composites/input/ToggleableInput.svelte
```diff
@@ -30,16 +31,27 @@ Usage:
...
-></textarea>
+{#if isEditable}
+    <Input
+        class="bg-background text-default w-full"
+        data-testid={`toggleable-input-${key}`}
+        inputClass={cn("border-input rounded-md border px-1.5", className)}
+        inputId={`toggleable-input-${key}`}
+        oninput={(event) => onInputChange((event.target as HTMLInputElement)?.value as string)}
+        placeholder={placeholder ?? ""}
+        required
```
Merseleo, line 42: Why it must be `required` now?

src/lib/components/composites/input/ToggleableInput.svelte
```diff
@@ -30,16 +31,27 @@ Usage:
...
+        type="text"
+        {value}
+    />
+{:else}
+    <div
+        class={cn(
+            "bg-background text-default flex h-10 w-full items-center overflow-hidden px-2",
```
Merseleo:
```suggestion
            "bg-background text-default flex h-10 w-full items-center overflow-hidden px-1.5",
```
Or why it should be different than in the input?

Slartibartfass2:
> Then we close #501 as duplicate right?

I totally forgot about this. I will link it in the pr.

src/lib/components/composites/input/ToggleableTextArea.svelte
```diff
@@ -0,0 +1,51 @@
...
+        oninput={(event) => onInputChange((event.target as HTMLTextAreaElement)?.value as string)}
+        {placeholder}
+        rows="1"
+        {value}
+    ></textarea>
+{:else}
+    <div
+        class={cn(
+            "bg-background text-default h-1 min-h-8 w-full border border-transparent px-1.5 py-1",
+            !value ? "text-neutral-500/85" : "",
```
Merseleo: Multiple questions: Why dont you use `text-slate-500` as color and why h-1 and then set the min height to 32px?

Merseleo (changes requested):
In general the code looks good, I think this a good improvement to differentiate the inputs from the textarea and I guess this will simplify the input validation for the paper input.

However, I noticed two problems: 

1) The abstract overflow seems to not working correctly (see next screenshot)

Image

2) and the div containing the abstract is not scrollable when there is not enough vertical space. I think there is missing an overflow or something similar.

src/lib/components/composites/input/ToggleableInput.svelte
Slartibartfass2: It doesn't really matter in this context. It was a required parameter for the input component, so I just added it.

src/lib/components/composites/input/ToggleableInput.svelte
Slartibartfass2: Because the input also has a border. If you set it to 1.5 and toggle the edit mode, you will see the elements move. However, they were a bit off, and I corrected it.

src/lib/components/composites/input/ToggleableTextArea.svelte
Slartibartfass2: `h-1` is a leftover from testing. Good catch!
The color matches the color of the placeholder of the textarea so that there's no difference.

Slartibartfass2 (commented):
> However, I noticed two problems:
> 
>     1. The abstract overflow seems to not working correctly (see next screenshot)
> 
>     2. and the div containing the abstract is not scrollable when there is not enough vertical space. I think there is missing an overflow or something similar.

Yes, I forgot that. I fixed it, and now it's scrolling.


Thank you for the review.

src/lib/components/composites/input/ToggleableInput.svelte
Merseleo: Ok.

src/lib/components/composites/input/ToggleableInput.svelte
Merseleo: I think they are still moving a little bit, that are details that can be ignored in my opinion.

src/lib/components/composites/input/ToggleableTextArea.svelte
Merseleo: Ah ok.

Merseleo:
> > However, I noticed two problems:
> > ```
> > 1. The abstract overflow seems to not working correctly (see next screenshot)
> > 
> > 2. and the div containing the abstract is not scrollable when there is not enough vertical space. I think there is missing an overflow or something similar.
> > ```
> 
> Yes, I forgot that. I fixed it, and now it's scrolling.

It is still not quite correct, when I am not in edit mode and have a smaller screen. But for the textarea component it is fine (i.e. when you are in edit mode)

Bildschirmfoto 2025-11-09 um 12 56 20

Merseleo (changes requested):
I think there is only one little problem left.

Merseleo:
> It is still not quite correct, when I am not in edit mode and have a smaller screen. But for the textarea component it is fine (i.e. when you are in edit mode)
> 
> Bildschirmfoto 2025-11-09 um 12 56 20

Did you worked on this?

Slartibartfass2:
> Did you worked on this?

I thought I had fixed this before, but then you still found this behavior and I could reproduce it on my Linux laptop.
I tried for a while, but couldn't find a fix since.

To get the overflow the way we want was always a struggle or we are just not good enough.

Merseleo:
> > Did you worked on this?
> 
> I thought I had fixed this before, but then you still found this behavior and I could reproduce it on my Linux laptop. I tried for a while, but couldn't find a fix since.
> 
> To get the overflow the way we want was always a struggle or we are just not good enough.

So what do you propose? Just ignore it?

Slartibartfass2:
> So what do you propose? Just ignore it? 

No. I will have another look. You can too, if you like. There's a solution, we just have to find it.

Merseleo:
> > So what do you propose? Just ignore it?
> 
> No. I will have another look. You can too, if you like. There's a solution, we just have to find it.

Ok, i we do not find any by next week, then we ignore this.

Merseleo:
Ok, i have an idea, but let's discuss this in another issue.

Merseleo (approved):

PR #560, opened 2025-11-09
Add setting for archiving / activating a (archived) project

### LINKED ISSUE #22: [Use Case]: Archive Project

Merseleo:
Closes #22 

## What I have made

- I added a settings component `ArchiveProjectSettings` that allows a user to archive an active project and activate a project that is currently archived
- I updated the `buildFieldMask` utility function to optionally set fields that should be excluded from the field mask, as the `id` field should not be contained in the most field masks for example
- I refactored the project settings by extracting identical +page.ts load functions in a single +layout.ts load function for the project settings and by improving the check, whether the current user is project admin

## Checklist

### Author

  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests

### Reviewer


src/lib/components/composites/settings/project-settings/general/ArchiveProjectSettings.svelte
```diff
@@ -0,0 +1,90 @@
...
+
+        await backendService
+            .updateProject({
+                project: Project.create(projectData),
+                mask: buildFieldMask(projectData, "project"),
+            })
+            .response.then(async () => {
+                await invalidate("data:getProjectById");
+                toast.success(
+                    `Successfully ${isAlreadyArchived.value ? "activated" : "archived"} the project.`,
```
luca-schlecker, line 59: This does not seem to work correctly. It always says "Successfully archived the project."

https://github.com/user-attachments/assets/607019ae-d2d8-4d54-b8f1-5e958cd2111e

luca-schlecker (changes requested):
~There was an issue with the toast.~
~Furthermore, the rename and invite user buttons are not properly disabled and will fail.~ Seems to be addressed separately in #525 

_Edit: #525 also fixes the toast issue for me._ As #525 includes the changes from this PR, should we just merge that one and close this one?

src/lib/components/composites/settings/project-settings/general/ArchiveProjectSettings.svelte
Merseleo: As you correctly mentioned, this problem is fixed in #525, sorry I should have separated these functionalities in a better way and fix this problem in this PR. So you are ok with this feature here?

PR #562, opened 2025-11-12
Exclude tests from fast-failing E2E run in preceeding full run

### LINKED ISSUE #561: [Feature]: Only run tests in E2E full run that haven't been executed in fail fast run

_No response_

Our E2E tests work in a way that we first make a fail-fast run that only runs changed tests (Playwright only-changed feature) and then run the whole test suite. If the fail-fast run will include all E2E tests, we run all tests twice. While this is the extreme case, there's always redundancy when the fail-fast run is executed.

We can exclude all tests that have been executed in the fail-fast run when we execute the full run. This way we only execute each test once and therefore have full coverage with reduced test time.

This can be achieved by listing all executed tests of the fail-fast run and excluding those tests in the full run afterward.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #561 

## What I have made

Refer to the issue for the detailed problem description and solution idea. This works because the `only-changed` feature of Playwright leads to executing all tests of a file even if only one test of it is added/modified/removed. In the full test run, we then can exclude whole test files because of this.

[Here](https://github.com/SE-UUlm/snowballr-frontend/actions/runs/19308355938/job/55221755523?pr=562) you can see the E2E workflow result without any changes to the tests (using only the first commit).

[Here](https://github.com/SE-UUlm/snowballr-frontend/actions/runs/19308543275/job/55222420981?pr=562) you can see the E2E workflow result with changes to test tests (using the two first commits).

Currently, we have 372 E2E tests (including those I added for verification purposes). You can see that the fast-failing test runs execute 66 tests, and the full test runs then execute the remaining 306.

The commit with the test additions will be removed before merging.

## Checklist

### Author

### Reviewer

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -80,13 +80,42 @@ jobs:
                       e2e-tests:
                         - 'tests/e2e/**'
 
+            - name: Set E2E test command alias
+              run: |
+                  rc=/tmp/rcfile
+                  echo "shopt -s expand_aliases" > $rc
+                  echo "alias run-e2e-tests='npm run test:e2e -- --shard ${{ matrix.shard }}/${{ strategy.job-total }}'" >> $rc
+
             - name: Run only tests for changed code (fail fast)
               # Skip fast failing test run when not triggered by a pull request and no changes were made to the e2e test directory
               if: ${{ github.event_name == 'pull_request' && steps.changes.outputs.e2e-tests == 'true' }}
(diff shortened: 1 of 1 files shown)


mowi12 (dismissed):
Nicely done!
I found no logical mistakes and since your demo worked I think this works.

I probably need to approve again after you remove the demo commits, but I will approve now for you to see that everything is approved : )

mowi12 (approved):

PR #563, opened 2025-11-13
Add export project functionality

### LINKED ISSUE #23: [Use Case]: Export Project

Project Admin

### Describe the use case

As project admin, I want to export a project (archived or not).

### Acceptance criteria

- the export contains the following information
  - project name, admins, members
  - stages of project
  - papers of each stage
    - paper title
    - reviewers and their decision + final decision
    - metadata
- the export is available in the following formats
  - JSON

### Anything else?

Part of #15

Slartibartfass2:
Closes #23

## What I have made

- I added the ExportProjectSettings component that lets the user export the project in a selected format defined by the backend.
- I standardized the way errors are handled and visualized in the application since there was a lot of duplication before. Now we can also control when we want to log the error and when not.

## Checklist

### Author

  - ~~[ ] unit tests~~ (no logic to test)
  - [x] integration tests
  - ~~[ ] end-to-end tests~~ (this feature not supported by the mock backend)

### Reviewer


src/lib/components/composites/select/SingleSelect.svelte
```diff
@@ -0,0 +1,41 @@
...
+    }: Props = $props();
+
+    let label = $derived(getSelectLabel(selectedValue));
+
+    function getSelectLabel(selectedValue: string | undefined): string {
+        return !selectedValue ? `No ${categoryLabel} selected` : selectedValue;
+    }
+</script>
+
+<Select.Root {disabled} type="single" bind:value={selectedValue}>
```
mowi12, line 37: Please add some doucmentation inlcuding an example usage

tests/setupTest.ts
```diff
@@ -342,3 +342,6 @@ if (typeof window !== "undefined") {
     window.HTMLElement.prototype.hasPointerCapture = vi.fn();
     window.HTMLElement.prototype.releasePointerCapture = vi.fn();
 }
+
+global.URL.createObjectURL = vi.fn();
```
mowi12, line 346: Please add a short sentence, mentioning that this is for the download

tests/integration/select/TestMultiSelect.svelte
mowi12, line 1: Please also test the single select component.

src/lib/utils/download-file.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Offers a download of the given data with the specified filename.
+ *
+ * @param byteArray - the data to be downloaded
+ * @param filename - the desired filename for the downloaded file
+ */
+export function downloadBlob(byteArray: Uint8Array<ArrayBufferLike>, filename: string): void {
+    const normalizedData = new Uint8Array(byteArray);
```
mowi12: This is not necessary

src/lib/components/composites/settings/project-settings/general/ExportProjectSettings.svelte
```diff
@@ -0,0 +1,106 @@
...
+<SettingsSection sectionTitle="Export Project">
+    {#await formatsPromise}
+        <span class="italic">Loading formats ...</span>
+    {:then}
+        <span>
+            You can export the project in the following
+            {pluralize(formats, "format", "formats")}:
+            {formats.join(", ")}.
+        </span>
+    {:catch}
```
mowi12: I don't think we need to list all the possible options. Either simply remove it or think of a better solution.

mowi12 (changes requested):

src/lib/components/composites/select/SingleSelect.svelte
Slartibartfass2: I added one.

src/lib/components/composites/settings/project-settings/general/ExportProjectSettings.svelte
Slartibartfass2: I removed it.

src/lib/utils/download-file.ts
Slartibartfass2: It is. Otherwise I get the following error:
```
Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'BlobPart'.
  Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'ArrayBufferView<ArrayBuffer>'.
    Types of property 'buffer' are incompatible.
      Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
        Type 'SharedArrayBuffer' is missing the following properties from type 'ArrayBuffer': resizable, resize, detached, transfer, transferToFixedLengthts(2322)
```

tests/integration/select/TestMultiSelect.svelte
Slartibartfass2: We don't need one. We used one for the multi-select to check which values are selected, but for the single select, we can easily determine this by reading the label.

tests/setupTest.ts
Slartibartfass2: Done.

Slartibartfass2 (commented):
This should be all then.

src/lib/utils/download-file.ts
mowi12:
```ts
export function downloadBlob(byteArray: Uint8Array<ArrayBufferLike>, filename: string): void {
    // create a browser Blob and object URL
    const file = new window.Blob([byteArray], {
        type: "application/octet-stream",
    });
```

This worked for me. I tested it and got no error.

mowi12 (changes requested):
There is one last conversation you need to take another look at. I think you simply forgot the brackets there.

And last time I forgot to mention, I would really like it, if the output would already be formatted. I don't think that should be too hard to do.

src/lib/utils/download-file.ts
Slartibartfass2: Weird, I could swear that I tested it this way.

mowi12 (approved):

PR #565, opened 2025-11-14
Make adjustments according to API version v0.12.0

### LINKED ISSUE #564: [Feature]: Make adjustments according to API version v0.12.0

_No response_

The current version is v0.9.0 and there have been some breaking changes that need to be addressed.

Address the breaking changes and adjust the code according to API version v0.12.0.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #564

## What I have made

There have been changes since v0.9.0. This makes the required adjustments.

I refactored `example-data.ts` to get type safety for the example model objects. Before, we only saw error in the usages of those objects but not in their definition because their type wasn't declared.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no new features)

### Reviewer

### DIFF
--- a/tests/e2e/project/settings/members/project-member-settings-page.test.ts
+++ b/tests/e2e/project/settings/members/project-member-settings-page.test.ts
@@ -237,27 +237,27 @@ test.describe("Remove Members Tests", () => {
         ).toBeDisabled();
     });
 
-    test("When other users are listed, then they can be removed from the project.", async ({
-        projectMembersSettingsPage,
-        page,
-    }) => {
-        const userEmail = "john.doe@example.com";
-        await projectMembersSettingsPage.inviteUser(userEmail, projectMembersSettingsPage);
+    test.fixme(
+        "When other users are listed, then they can be removed from the project.",
+        async ({ projectMembersSettingsPage, page }) => {
+            const userEmail = "john.doe@example.com";
+            await projectMembersSettingsPage.inviteUser(userEmail, projectMembersSettingsPage);
 
-        await projectMembersSettingsPage.openRemoveMemberDialog({ email: userEmail });
-        await projectMembersSettingsPage.removeMemberDialog.remove();
-        await expect(projectMembersSettingsPage.removeMemberDialog.dialog).not.toBeVisible();
-        await expect(page.getByText(`Removed ${userEmail} from the project`)).toBeVisible();
-        await projectMembersSettingsPage.checkForUser(userEmail, "email", false);
-        await projectMembersSettingsPage.assertNumberOfProjectMembers({
-            all: 1,
-            admins: 1,
-            members: 0,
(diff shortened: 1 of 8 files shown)


tests/integration/project-components/create-project-dialog.test.ts
```diff
@@ -54,7 +54,7 @@ describe.sequential("CreateProjectDialog", () => {
         await user.tab();
 
         const membersInput = screen.getByRole("textbox", { name: "Members" });
-        await user.type(membersInput, Members.demoMember1.user.email);
+        await user.type(membersInput, Members.demoMember1.user!.email);
```
Merseleo, line 57: Why is this now necessary?

tests/e2e/project/settings/members/project-member-settings-page.test.ts
```diff
@@ -237,27 +237,27 @@ test.describe("Remove Members Tests", () => {
...
-        await projectMembersSettingsPage.assertNumberOfProjectMembers({
-            all: 1,
-            admins: 1,
-            members: 0,
-            invitees: 0,
-            invitedAdmins: 0,
-            invitedMembers: 0,
-        });
-    });
+    test.fixme(
```
Merseleo, line 240: But then it would be nice, if you open an issue that all test.fixme() e2e tests (not only these) should be fixed and implemented. What you do me this favor?

Merseleo (changes requested):
It looks very good to me. I tested to remove invitations using the real backend and this seems to work, so it wouldn't know what do complain about. Good work!

tests/integration/project-components/create-project-dialog.test.ts
Slartibartfass2: Previously those demo members weren't declared as ProjectMember objects. Now they are, and in the API definition of this class, `user` can be undefined.

tests/e2e/project/settings/members/project-member-settings-page.test.ts
Slartibartfass2: See https://github.com/SE-UUlm/snowballr-frontend/issues/567

Slartibartfass2 (commented):
Thank you for the review!

tests/integration/project-components/create-project-dialog.test.ts
Merseleo: Ah right, this was the weird thing with the nested objects.

tests/e2e/project/settings/members/project-member-settings-page.test.ts
Merseleo: Great, thank you.

Merseleo (approved):
I think everything is fine, so lets merge this so you can continue working.

PR #569, opened 2025-11-24
Improve overflow handling on paper view

### LINKED ISSUE #568: [Feature]: Improve overflow handling on paper view

_No response_

I'm not happy with the way we handle vertical overflow on the paper view: 

Image

When I open the paper view with only few vertical space, then the card components should handle the overflow and should be either scrollable or hide the content, but not the components itself.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #568

## What I have made

- I changed the overflow behavior of the components on the paper view page, so that
  - the abstract does not overflow the paper card
  - the criteria lists does not scroll and hide all criteria
- I changed the behavior, so that all components inside the paper card component vertically take all possible space and if there is not enough space, then the entire content of the paper card gets scrollable

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/lib/components/composites/criteria/CriteriaList.svelte
+++ b/src/lib/components/composites/criteria/CriteriaList.svelte
@@ -37,7 +37,7 @@ Usage:
     <CriteriaList listTitle="Hard Exclusion" {reviewers} {criteria} />
 ```
 -->
-<section class="flex flex-col gap-5 overflow-hidden">
+<section class="flex flex-col gap-5">
     <NamedList
         {emptyHint}
         errorHint="Couldn't load criteria"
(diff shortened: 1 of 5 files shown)


Slartibartfass2:
This doesn't seem to work when I try it. The page just keeps scrolling:
image

Slartibartfass2 (changes requested):
See my other comment.
I found a fix and committed it on `tmp/fix-overflow`. Please check whether this works in your browser as well.

Merseleo:
> See my other comment. I found a fix and committed it on `tmp/fix-overflow`. Please check whether this works in your browser as well.

Probably the best solution we will find.

Slartibartfass2 (approved):

PR #574, opened 2025-11-28
Use actions from snowballr-ci

### LINKED ISSUE #573: [Feature]: Use Actions from snowballr-ci

_No response_

We newly introduced the snowballr-ci repo, which collects commonly used actions that we can reuse in this repo.

Replace the current workflows with the reusable actions where possible.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #573 

## What I have made

With the use of the actions and workflows from snowballr-ci, I made the following changes:
- deleted `deploy.yml` (now `docker.yml`)
- deleted `add_auto_gen_wiki_hint.sh` (now part of wiki)
- deleted `wiki-replace-github-urls.js` (now part of lint-md)

We now use the following actions and workflows from snowballr-ci:
- release (we don't have releases now, but if we add one, we already have the CI)
- teamscale-upload
- docker
- ensure-linear-history
- lint-md
- wiki

## Checklist

### Author

### Reviewer


.github/workflows/build.yml
```diff
@@ -2,11 +2,9 @@ name: Build
 
 on:
     push:
-        branches:
-            - main
-            - develop
+        branches: ["develop"]
```
mowi12, line 5: What about main?

mowi12 (changes requested):
Just one question.
I'm not sure if I overlooked this in the backend but let's first answer this.

.github/workflows/build.yml
Slartibartfass2: We don't really need to build again on main since we do this on develop and in the PR before that. Main really is just the branch with the latest "stable" release.

In the backend, we also only build on develop and on tags.

mowi12 (approved):

PR #590, opened 2025-12-20
Fix PaperBookmarkButton

### LINKED ISSUE #589: [Bug]: PaperBookmarkButton isn't reloaded when navigating paper

#99

### What happened?

When navigating papers, the paper bookmark status isn't correctly loaded.

### Steps to reproduce

Image

See that paper with ID 4 and title "a" is shown as bookmarked but isn't on the reading list.

### What did you expect?

Paper with ID 4 shouldn't be shown as bookmarked.

### Anything else?

_No response_

Slartibartfass2:
Closes #589 

## What I have made

bug-fix

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/lib/components/composites/button/PaperBookmarkButton.svelte
+++ b/src/lib/components/composites/button/PaperBookmarkButton.svelte
@@ -6,16 +6,17 @@
     import { resource } from "$lib/resource.svelte";
     import { backendService } from "$lib/grpc-api";
     import { cn } from "$lib/utils/shadcn-helper";
+    import { LoaderCircle } from "lucide-svelte";
 
     interface Props {
-        loadingPaperId: Promise<string>;
+        paperId: string;
         isBookmarkedDefault?: boolean;
         onPaperChangedBookmarkStatus?: () => void;
         class?: string;
     }
 
     const {
-        loadingPaperId,
+        paperId,
         isBookmarkedDefault = false,
         onPaperChangedBookmarkStatus = undefined,
         class: className,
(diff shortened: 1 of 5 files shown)


src/lib/components/composites/button/PaperBookmarkButton.svelte
```diff
@@ -6,30 +6,28 @@
     import { resource } from "$lib/resource.svelte";
     import { backendService } from "$lib/grpc-api";
     import { cn } from "$lib/utils/shadcn-helper";
+    import { LoaderCircle } from "lucide-svelte";
 
     interface Props {
-        loadingPaperId: Promise<string>;
+        paperId: string;
```
Merseleo, line 12: You forgot to adapt the comment.

src/lib/components/composites/button/PaperBookmarkButton.svelte
```diff
@@ -6,30 +6,28 @@
...
     const {
-        loadingPaperId,
+        paperId,
         isBookmarkedDefault = false,
         onPaperChangedBookmarkStatus = undefined,
         class: className,
     }: Props = $props();
 
+    console.log("foo");
+
```
Merseleo: Please remove this.

src/lib/components/composites/button/PaperBookmarkButton.svelte
```diff
@@ -128,7 +120,9 @@ Usage:
     onmouseleave={onMouseLeave}
 >
     {#snippet trigger()}
-        {#if isHovered}
+        {#if isUpdatingBookmarkStatus}
+            <LoaderCircle class="animate-spin" />
```
Merseleo, line 122: In theory we could also use the new `Spinner` shadcn component, what do you think?

Merseleo (changes requested):
I think only minor things have to be cleaned up

src/lib/components/composites/button/PaperBookmarkButton.svelte
Slartibartfass2: Fixed it.

src/lib/components/composites/button/PaperBookmarkButton.svelte
Slartibartfass2: Forgot that. I removed it.

src/lib/components/composites/button/PaperBookmarkButton.svelte
Slartibartfass2: There appears to be a problem with this. Also, something changed with the lucide-svelte package. I would do this in another PR.

Merseleo (approved):
LGTM

PR #592, opened 2025-12-20
Use shadcn ScrollArea for scollable content

### LINKED ISSUE #591: [Feature]: Use shadcn ScrollArea for scollable content

_No response_

So far, we've managed scrollable content ourselves. The scrollbars depend on the OS. The shadcn ScrollArea adds a common scrollbar which looks good.

Use the ScrollArea component.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #591

## What I have made

I wrapped scrollable content in the ScrollArea component from shadcn.

What do you think about the padding on the right side? I'm not sure about it.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only visual changes)

### Reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
         "compile:proto": "protoc --ts_out src/lib/model/api --ts_opt generate_dependencies --proto_path node_modules/snowballr-api/ node_modules/snowballr-api/main.proto"
     },
     "devDependencies": {
+        "@internationalized/date": "^3.10.1",
         "@playwright/test": "^1.57.0",
         "@protobuf-ts/grpcweb-transport": "^2.11.1",
         "@protobuf-ts/plugin": "^2.11.1",
@@ -51,7 +52,7 @@
         "@types/node": "^24.1.0",
         "@vitest/coverage-v8": "^3.2.4",
         "@vitest/ui": "^3.0.2",
-        "bits-ui": "^2.1.0",
+        "bits-ui": "^2.14.4",
         "clsx": "^2.1.1",
         "cross-fetch": "^4.1.0",
         "d3": "^7.9.0",
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
         "prettier-plugin-svelte": "3.4.0",
         "prettier-plugin-tailwindcss": "^0.6.14",
         "protobuf-fieldmask": "^2.0.0",
+        "resize-observer-polyfill": "1.5.1",
         "svelte": "^5.45.2",
         "svelte-check": "^4.3.1",
         "svelte-sonner": "^1.0.5",
--- a/src/lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte
+++ b/src/lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte
@@ -15,7 +15,7 @@
         selectedTab: ProjectTabValue;
         children?: Snippet | undefined;
         isCurrentUserAdmin: boolean;
-        isScrollable?: boolean;
+        allScrollable?: boolean;
     }
 
     let {
@@ -23,7 +23,7 @@
         selectedTab,
         children = undefined,
         isCurrentUserAdmin,
-        isScrollable = true,
+        allScrollable = true,
     }: Props = $props();
 
     const ALL_TABS = [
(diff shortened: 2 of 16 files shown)


src/lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte
mowi12, line 1: Any reason why only the `ProjectSettingsLayout` is explicitly set to scrollable and the user settings layout not?

This directly leads to the general question, why you explicitly set scrollable to `true` here, if the default is already `true`.

package.json
```diff
@@ -32,6 +32,7 @@
         "compile:proto": "protoc --ts_out src/lib/model/api --ts_opt generate_dependencies --proto_path api/proto/ api/proto/main.proto"
     },
     "devDependencies": {
+        "@internationalized/date": "^3.10.1",
```
mowi12, line 37: What is this needed for?

package.json
```diff
@@ -71,6 +72,7 @@
         "prettier-plugin-svelte": "^3.3.3",
         "prettier-plugin-tailwindcss": "^0.6.14",
         "protobuf-fieldmask": "^2.0.0",
+        "resize-observer-polyfill": "1.5.1",
```
mowi12, line 77: And could you please explain shortly, what this is doing as well and why we need it now?

mowi12 (changes requested):
Nice work. I really like the changes, very subtle.

Regarding the padding on the right side, I'm also not sure, but I would be fine with both. And if I'm honest, nobody will see this change. So let's keep it this way.

But I have one general question:

What about this scroll bar here if I want to invite a member. I'm not sure if this was always here, but maybe we can disable it if it's not scrollable?

Image

Can you tell me, why there are two different scroll bar designs here?
This should be the member list and settings section of a project.
Is it simply because the content is different? One time it's actually a list of something , and one time a page kind of?

Image

Image

src/lib/components/composites/settings/project-settings/ProjectSettingsLayout.svelte
Slartibartfass2: Currently, we only use it in the project member settings. All the other settings are scrollable as a whole setting. In the project member settings, we want the members list to be scrollable.

package.json
Slartibartfass2: This was automatically added when adding the ScrollArea component.

package.json
Slartibartfass2: The ResizeObserver interface is used by the ScrollArea component. It is provided by the browser. When running the test using Node, it is not available. The polyfill provides an implementation that we can use for the tests.

Slartibartfass2 (commented):
> What about this scroll bar here if I want to invite a member. I'm not sure if this was always here, but maybe we can disable it if it's not scrollable?

I can't reproduce this. Can you please provide some steps?

> Can you tell me, why there are two different scroll bar designs here? This should be the member list and settings section of a project. Is it simply because the content is different? One time it's actually a list of something , and one time a page kind of?

From which component is the right picture? I didn't find it.

Thank you for taking your time to review this PR.

mowi12:
> > What about this scroll bar here if I want to invite a member. I'm not sure if this was always here, but maybe we can disable it if it's not scrollable?
> 
> I can't reproduce this. Can you please provide some steps?

This is literally the member invite dialog in the project settings. No steps needed here imo. Simply create a project and try to invite a member that exists. As soon as there are some users to suggest, the scroll bar appears on the right side of the list. This was not the case before your changes.

> > Can you tell me, why there are two different scroll bar designs here? This should be the member list and settings section of a project. Is it simply because the content is different? One time it's actually a list of something , and one time a page kind of?
> 
> From which component is the right picture? I didn't find it.

The right side is simply the general section of the project settings, but the height of my window is very small, so a scroll bar appears on the right side. It's not the browser scroll bar itself, that I'm sure of.

Slartibartfass2 (commented):
> This is literally the member invite dialog in the project settings. No steps needed here imo. Simply create a project and try to invite a member that exists. As soon as there are some users to suggest, the scroll bar appears on the right side of the list. This was not the case before your changes.

This is what I got:
Image

Can you investigate this or provide additional information? Maybe this is an OS-specific problem.

> The right side is simply the general section of the project settings, but the height of my window is very small, so a scroll bar appears on the right side. It's not the browser scroll bar itself, that I'm sure of.

That's the native scrollbar. I changed the code so that it also uses the ScrollArea. Did you find another place where we need to replace the native one with the component?

mowi12 (changes requested):
Awesome. I couldn't find any other section where the new scroll bar is missing.

I also took anther look at the invite user member scroll bar problem:

First I compared the pages in Firefox and Chrome, but this lead to nothing, since both browsers do show the scrollbar (left Firefox, right Chrome:

Image

Then I simply added more users so I could verify it actually is the scrollbar:

Image

So maybe also use the custom scrollbar here as well?

Slartibartfass2 (commented):
I added the ScrollArea component to the ChipsInput.

mowi12 (approved):
Perfect! Now everything looks consistent!

PR #599, opened 2026-01-08
Delete project

### LINKED ISSUE #20: [Use Case]: Delete Project

Project Admin

### Describe the use case

As project admin, I want to delete a project.

### Acceptance criteria

- a deleted project is merely marked as deleted and not completely removed from the database (soft deletion)
- the project admin is warned that they are about to delete the project
- the project admin has to confirm their decision to delete the project (e.g. type in the name of project, similar to when deleting a GitHub repository)

### Anything else?

_No response_

Slartibartfass2:
Closes #20

## What I have made

One can delete a project.

The project can still be accessed via the URL, this should be fixed in the getProjectById call.

## Checklist

### Author

  - [x] unit tests
  - [x] integration tests
  - [x] end-to-end tests

### Reviewer

### DIFF
--- a/src/lib/components/composites/settings/SettingsSection.svelte
+++ b/src/lib/components/composites/settings/SettingsSection.svelte
@@ -1,15 +1,42 @@
+<script lang="ts" module>
+    export const settingsSectionVariants = tv({
+        base: "flex w-full flex-col gap-3",
+        variants: {
+            variant: {
+                default: "",
+                destructive: "border-error rounded-xl border py-4",
+            },
+        },
+        defaultVariants: {
+            variant: "default",
+        },
+    });
+
+    export type SettingsSectionVariant = VariantProps<typeof settingsSectionVariants>["variant"];
+
+    export interface SettingsSectionProps {
+        sectionTitle: string;
+        children: Snippet;
+        loading?: boolean;
+        variant?: SettingsSectionVariant;
+    }
+</script>
+
 <script lang="ts">
     import type { Snippet } from "svelte";
     import { Separator } from "$lib/components/primitives/separator";
     import { LoaderCircle } from "lucide-svelte";
+    import { tv, type VariantProps } from "tailwind-variants";
+    import { cn } from "$lib/utils/shadcn-helper";
 
-    interface SettingsSectionProps {
-        sectionTitle: string;
-        children: Snippet;
-        loading?: boolean;
-    }
+    const {
+        sectionTitle,
+        children,
+        loading = false,
+        variant = "default",
+    }: SettingsSectionProps = $props();
 
-    const { sectionTitle, children, loading = false }: SettingsSectionProps = $props();
+    const hasSeparator = variant === "default";
 </script>
 
 <!--
@@ -33,11 +60,15 @@ Usage:
     data-testid={`settings-section-${sectionTitle.toLowerCase().replace(" ", "-")}`}
 >
     <div class="flex flex-row items-center gap-3">
-        <h2>{sectionTitle}</h2>
+        <h2 class:text-error={variant !== "default"}>{sectionTitle}</h2>
         {#if loading}
             <LoaderCircle class="animate-spin" />
         {/if}
     </div>
-    <Separator />
-    {@render children?.()}
+    {#if hasSeparator}
+        <Separator />
+    {/if}
+    <div class={cn(settingsSectionVariants({ variant }))}>
+        {@render children?.()}
+    </div>
 </section>
--- a/src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+    import { goto } from "$app/navigation";
     import AlertDialog from "$lib/components/composites/dialog/AlertDialog.svelte";
     import { buttonVariants } from "$lib/components/primitives/button";
     import Input from "$lib/components/primitives/input/input.svelte";
@@ -7,6 +8,7 @@
     import type { Project } from "$lib/model/api/project";
     import { loadingWrapper } from "$lib/utils/common-helper";
     import { onMount } from "svelte";
+    import { toast } from "svelte-sonner";
 
     interface Props {
         projectId: string;
@@ -51,8 +53,9 @@
 
         await backendService
             .softDeleteProject({ id: projectId })
-            .response.then(() => {
-                open = false;
+            .response.then(async () => {
+                toast.success("Successfully deleted the project.");
+                await goto("/");
             })
             .catch((error) => {
                 deleteProjectError = error;
--- a/src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
+++ b/src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
@@ -81,6 +81,9 @@ Usage:
         disabled: loadingProjectName || !canDeleteProject,
         onclick: (args) => loadingWrapper(loading, deleteProject, args),
     }}
+    cancelProps={{
+        onclick: () => (confirmationText = ""),
+    }}
     error={deleteProjectError}
     errorText="Couldn't delete this project"
     loading={loading.value}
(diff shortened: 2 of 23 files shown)


Merseleo:
> The project can still be accessed via the URL, this should be fixed in the getProjectById call.

Can you create an issue for this problem?

src/lib/utils/common-helper.ts
```diff
@@ -1,15 +1,30 @@
...
+ * @param person - the person object
+ * @returns the name of the person as string
  */
 function getName(person: Person): string {
-    return `${person.firstName} ${person.lastName}`;
+    return `${person.firstName} ${person.lastName}`.trim();
+}
+
+/**
+ * Same as {@link getName} but returns the email of the person if the name is empty.
```
Merseleo: "if the name is empty" sounds a bit weird. What about "if the name of the person is not given / unknown". But actually this isn't necessary at all. You can also close this conversation.

src/lib/components/composites/settings/SettingsSection.svelte
Merseleo, line 1: Very clever using the variant for different styling, nice one ;)

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
```diff
@@ -0,0 +1,104 @@
...
+
+    let confirmationText = $state("");
+    let canDeleteProject = $derived(
+        confirmationText.length > 0 && confirmationText === projectName,
+    );
+
+    onMount(async () => {
+        loadingProjectName = true;
+
+        await loadingProject
```
Merseleo, line 35: If you wait for the project name on mount, why not on the project id as well and not pass them all the way down through the components?

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
```diff
@@ -0,0 +1,104 @@
...
+        "aria-label": "Delete this project",
+    }}
+    bind:open
+>
+    {#snippet trigger()}
+        Delete Project
+    {/snippet}
+    {#snippet description()}
+        Once deleted, the project cannot be accessed anymore. Ensure that no member is still
+        conducting an SLR.
```
Merseleo, line 102: really "an" SLR and not "a" SLR

tests/integration/settings/project-settings/general/delete-project-settings.test.ts
```diff
@@ -0,0 +1,167 @@
...
+import { render, screen, waitFor } from "@testing-library/svelte";
+import { Projects } from "$tests/example-data";
+import userEvent, { type UserEvent } from "@testing-library/user-event";
+import { mockApiCall, mockFailedApiCall } from "$tests/setupTest";
+import { loading } from "$tests/model-builder";
+import DeleteProjectSettings from "$lib/components/composites/settings/project-settings/general/DeleteProjectSettings.svelte";
+
+describe("DeleteProjectSettings", () => {
+    async function openDialog(user: UserEvent) {
+        const trigger = screen.getByTestId("alert-dialog-trigger");
```
Merseleo, line 11: Is it not possible to get the component by role (button) and text? Or isn't the trigger a button?

Merseleo (changes requested):
Very great work, everything works as it should, hardly something to complain about in the code, this is how I like reviewing.

Merseleo:
But could you add one E2E test for deleting a project?

Slartibartfass2:
> > The project can still be accessed via the URL, this should be fixed in the getProjectById call.
> 
> Can you create an issue for this problem?

https://github.com/SE-UUlm/snowballr-backend/issues/420

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
Slartibartfass2: Yes. It's about how it sounds, not how it's written. See https://www.scribbr.com/commonly-confused-words/a-vs-an/

src/lib/components/composites/settings/SettingsSection.svelte
Slartibartfass2: Thanks :)

src/lib/utils/common-helper.ts
Slartibartfass2: Changed it.

tests/integration/settings/project-settings/general/delete-project-settings.test.ts
Slartibartfass2: Not necessarily. Check the implementation for more information. Also, the text of the button has more than one occurrence, which makes it a bit more difficult. So far, we have used the test ID for the AlertDialog trigger.

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
Slartibartfass2: I'm not sure if I understand you correctly. Why should I wait for the project ID? It is already there.

Slartibartfass2 (commented):
I added an e2e test. Thanks for your review. Somehow there are some other e2e tests failing now. I couldn't spot the problem yet, do you?

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
Merseleo: Because you could also extract it from the project as you wait for the project name anyway at the beginning and then you have to pass less properties down in the component tree. But you can also ignore this and close this conversation on your own.

src/lib/components/composites/settings/project-settings/general/DeleteProjectSettingsDialog.svelte
Merseleo: Ok, because you pronounce it as acronym.

tests/e2e/project/settings/general/project-general-settings-page-model.ts
```diff
@@ -37,4 +47,15 @@ export class ProjectGeneralSettingsPageModel {
         await this.renameButton.click();
         this.projectName = projectName;
     }
+
+    async deleteProject() {
+        await this.archiveButton.hover();
+        await this.page.mouse.wheel(0, 1000); // Scroll down to make the delete button visible
```
Merseleo, line 53: Why not use `scrollIntoViewIfNeeded`?

tests/e2e/project/settings/general/project-general-settings-page-model.ts
```diff
@@ -37,4 +47,15 @@ export class ProjectGeneralSettingsPageModel {
...
+
+        await this.deleteButton.click();
+        expect(this.deleteProjectDialog).toBeVisible();
```
Merseleo:
```suggestion
        await expect(this.deleteProjectDialog).toBeVisible();
```

Merseleo (changes requested):
> I added an e2e test. Thanks for your review. Somehow there are some other e2e tests failing now. I couldn't spot the problem yet, do you?

Yes, I have an idea: Because you wrapped the settings content in another `div`
```html
<div class={cn(settingsSectionVariants({ variant }))}>
        {@render children?.()}
    </div>
```
The test "When the user deletes a fetcher, then it is not visible anymore" fails as the command `getFetcherRow` doesnt work anymore. I think there you have to specify more precisely which div you want (actually the person who wrote the test did not a good job, as the locator is too specific). Do you understand the problem from my description?

tests/e2e/project/settings/general/project-general-settings-page-model.ts
Slartibartfass2: It didn't work, so I went for this solution.

Slartibartfass2 (commented):
Thanks for the review. I fixed the E2E locator for the fetcher settings. I made it a semantically correct list and then got the row by the `listitem` role.

tests/e2e/project/settings/general/project-general-settings-page-model.ts
Merseleo: Weird

Merseleo (approved):
> Thanks for the review. I fixed the E2E locator for the fetcher settings. I made it a semantically correct list and then got the row by the `listitem` role.

Great solution. I tested everything again and did not found any problem, so great work (especially with the fetcher settings refactor). Thank you for continually improving this software.

PR #602, opened 2026-01-15
Redirect to previously accessed URL when not signed in on sign in

### LINKED ISSUE #412: [Use Case]: Redirect to previously accessed URL after required sign in

User

### Describe the use case

As a user, I want to access a certain route e.g., `/project/1/dashboard`, but I am not signed in, so I get redirected to the sign-in page.
After the sign in, I would like to be redirected to the previously accessed route `/project/1/dashboard` instead of the home page.

### Acceptance criteria

After signing in, the user is redirected to the previously accessed route.

### Anything else?

Low priority, nice to have.

Slartibartfass2:
Closes #412

## What I have made

Ensure that you're signed out and open:
- http://localhost:4173/project/3a5d5a02-6afb-4fb3-b892-c90119aa3a0f/dashboard
- http://localhost:4173/project/3a5d5a02-6afb-4fb3-b892-c90119aa3a0f/settings/members
- http://localhost:4173/readinglist
- http://localhost:4173/settings/account

First, you should be redirected to the sign-in page and then to the respective page you initially wanted to open.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (none required)

### Reviewer

### DIFF
--- a/src/lib/utils/search-parameters.ts
+++ b/src/lib/utils/search-parameters.ts
@@ -11,12 +11,16 @@ import {
 import { callDebounced, stringToEnumValue } from "$lib/utils/common-helper";
 import { goto } from "$app/navigation";
 
+function getUrlSearchParam(key: string): string | null {
+    return page.url.searchParams.get(key);
+}
+
 /**
  * Returns the 'searchText' query parameter from the current URL.
  * If this query parameter is not set, an empty string is returned as default.
  */
 export function getSearchTextFromURL() {
-    return page.url.searchParams.get("searchText") ?? "";
+    return getUrlSearchParam("searchText") ?? "";
 }
 
 /**
@@ -51,8 +55,7 @@ export function updateSearchTextParam(
  * Each filter is expected to be a comma-separated list (e.g., ?stages=stage1,stage2).
  */
 export function getFilterFromURL(): ProjectPaperFilter {
-    const getArray = (key: string) =>
-        page.url.searchParams.get(key)?.split(",").filter(Boolean) ?? [];
+    const getArray = (key: string) => getUrlSearchParam(key)?.split(",").filter(Boolean) ?? [];
 
     return {
         stages: getArray("stages"),
@@ -96,8 +99,8 @@ export function updateFiltersParam(
  * If this query parameter is not set, 'Id: Low to High' is returned as default.
  */
 export function getSortOptionFromURL(): SortOptionLabel {
-    const sortCriterion = page.url.searchParams.get("sort");
-    const sortDirection = page.url.searchParams.get("order");
+    const sortCriterion = getUrlSearchParam("sort");
+    const sortDirection = getUrlSearchParam("order");
 
     if (!sortCriterion || !sortDirection) {
         return "Id: Low to High";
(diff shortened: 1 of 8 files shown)


src/lib/utils/search-parameters.ts
```diff
@@ -155,3 +158,33 @@ export function updateUrlParams(searchParameters: SvelteURLSearchParams): void {
...
+ * Adds the 'redirect' query parameter to the given URL if it exists in the current URL.
+ * I.e., it passes the redirect parameter along to another URL.
+ *
+ * @param url - The URL to which the redirect parameter should be added.
+ * @returns The URL with the added redirect parameter if it exists.
+ */
+export function addRedirectUrlIfExists(url: string) {
+    const redirectUrl = getRedirectParam();
+    return url + (redirectUrl ? `?redirect=${encodeURIComponent(redirectUrl)}` : "");
+}
```
Merseleo, line 190: Why dont you use the `SvelteUrlSearchParams` object and manipulate this one as in the other `update...` functions?

src/routes/+layout.ts
```diff
@@ -76,19 +77,25 @@ export const load: LayoutLoad = async ({ depends, url, fetch }) => {
...
+
 /**
  * Redirects to the sign-in page and clears the user cache.
  * This function is used when the user is not authenticated or when an error occurs
  * that implies the user is not authenticated.
  *
  * @returns An empty user object.
  */
-async function redirectToSignIn() {
+async function redirectToSignIn(url: URL) {
```
Merseleo, line 97: You forgot to describe the usage of this parameter in the comment.

src/lib/utils/search-parameters.ts
Merseleo: I think this could especially simplify the logic in the `redirectToSignIn` in the root layout.ts. Or am I mistaken?

Merseleo (changes requested):
I manually tested it and it worked most of the time, nice feature! However, sometimes I got stuck in the loading screen: 

Image

I got this error after logging out and then accessing the _/settings/account_ account and signing in. Can you reproduce this error?

And I think it would be great to add an E2E test in the signin tests that demonstrate, that the redirect works. What do you think?

src/lib/utils/search-parameters.ts
Slartibartfass2: I don't understand how I would solve this. Could you please provide a solution?

src/routes/+layout.ts
Slartibartfass2: I added it.

Slartibartfass2 (commented):
Thanks for the review. I fixed the bug and added the E2E test.

src/lib/utils/search-parameters.ts
Merseleo: I had another look into this and I think it wouldn't make sense to use the `SvelteUrlSearchParams` as I misunderstood your function and it actually does something different than I initially thought. So everything is ok.

Merseleo (approved):
The feature seems to work, the E2E test looks great and I didn't found any more issues. However, the solution with fetching the user from the cache when their is no user in the context seems a bit sketchy to me and I do not understand completely why this must always be correct and it is ok to return `user!`. But I think the entire authentication is not an issue or thing we should try to clean up here.

PR #605, opened 2026-01-31
Set Snowballing Type

### LINKED ISSUE #60: [Use Case]: Set Snowballing Type

Project Admin

### Describe the use case

As project admin, I want to set the type of snowballing used in a project.

### Acceptance criteria

- allowed types are: "forwards", "backwards", and "both"
- only one can be chosen
- "both" is the default

### Anything else?

Part of #15

Slartibartfass2:
Closes #60

## What I have made

- I implemented the Snowballing Type settings
- In 144dc4a3827a419ddf1eb1936849f9d605a8a2a4 I migrated lucide from the old to the new version because when we add components from shadcn it uses the new version. There's no big difference, only the package name.

## Checklist

### Author

  - ~~[ ] unit tests~~ (nothing added that can be tested by unit tests)
  - [x] integration tests
  - [x] end-to-end tests

### Reviewer


src/lib/components/composites/settings/project-settings/slr/SnowballingTypeSettings.svelte
```diff
@@ -0,0 +1,156 @@
...
+        },
+        {
+            id: "both",
+            label: "Both",
+            value: SnowballingType.BOTH,
+            description: "Both forward and backward references are fetched",
+        },
+    ];
+    let selectedTypeId: (typeof options)[number]["id"] = $state("");
+    let initialTypeId = $state("");
```
Merseleo: Only a question of interest? Why do you store the initial and selected id and not the value directly? So the `SnowballingType`? I do not recognize any advantage, but unnecessary finds.

tests/integration/settings/project-settings/slr/snowballing-type-settings.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import userEvent from "@testing-library/user-event";
+
+describe("SnowballingTypeSettings", () => {
+    const projectData = createProject({
+        settings: createProjectSettings({
+            snowballingType: SnowballingType.BOTH,
+        }),
+    });
+
+    test("When all props are provided and the SLR settings not locked, then the component renders correctly", async () => {
```
Merseleo:
```suggestion
    test("When all props are provided and the SLR settings are not locked, then the component renders correctly", async () => {
```

tests/e2e/project/settings/slr/project-slr-settings-page.test.ts
```diff
@@ -66,3 +66,48 @@ test.describe("SLR Settings Tests", () => {
...
 });
+
+test.describe("SLR Settings Tests - Snowballing Type", () => {
+    test("When the SLR settings are not locked, then the radio buttons are enabled", async ({
+        projectSLRSettingsPage,
+    }) => {
+        await expect(projectSLRSettingsPage.snowballingTypeForwardRadio).toBeEnabled();
+        await expect(projectSLRSettingsPage.snowballingTypeBackwardRadio).toBeEnabled();
+        await expect(projectSLRSettingsPage.snowballingTypeBothRadio).toBeEnabled();
+    });
```
Merseleo: I think this test is not necessary, as you already test this in your integration tests and it slows down the e2e test suite further.

Merseleo: The same applies for the next tests. I would rather add a more general test, that all settings on the SLR page are disabled if the project is locked, but not every single setting separably, as we therefore have our integration tests.

Merseleo: Same for the last test, although this could make more sense. But actually you should then search for a paper and check, whether after the change only backward references are checked. This would be true end to end testing with the backend (which is not possible at the moment).

Merseleo (changes requested):
Great work, everything seems to work as expected (although the update sometimes lags and it seems a bit weird to wait for a conformation when using a radio button). 

One small thing, I want to discuss with you. What do you think of increasing the spacing between the settings sections to `gap-9` instead of `gap-5` as in the Mockup? Because now everything feels so cluttered?

tests/e2e/project/settings/slr/project-slr-settings-page.test.ts
Slartibartfass2: I removed those unnecessary tests.

src/lib/components/composites/settings/project-settings/slr/SnowballingTypeSettings.svelte
Slartibartfass2: I thought I tried this and it didn't work, but now I tried again and it works. Thanks for reminding me again.

Slartibartfass2 (commented):
> Great work, everything seems to work as expected (although the update sometimes lags and it seems a bit weird to wait for a conformation when using a radio button).
> 
> One small thing, I want to discuss with you. What do you think of increasing the spacing between the settings sections to `gap-9` instead of `gap-5` as in the Mockup? Because now everything feels so cluttered?

I like it. I added it. Thanks for the review.

src/lib/components/composites/settings/project-settings/slr/SnowballingTypeSettings.svelte
Slartibartfass2: I now remember. It was because we would have to convert back and forth between the numeric and string representation of the type.

I updated the code to handle this. Please take a look and tell me if you would still prefer this solution. I'm fine with both.

Merseleo: I prefer your current solution, looks good 👍🏻

Merseleo (approved):
LGTM

PR #625, opened 2026-03-05
Only run fast failing E2E tests in PR

### LINKED ISSUE #614: [Feature]: Only run fast-failing E2E tests on PR

_No response_

Currently, we first run the fast-failing E2E tests and then continue with the rest. In general, this takes quite some time. @Merseleo had the idea that we should only run the fast-failing tests on PRs and then run all E2E tests after merging. The fast-failing tests are sufficient for detecting errors in the changes made in the PR.

Only run fast-failing tests on PRs.

_No response_

### Anything else?

It's possible to read the PR description, and we could use this to add a flag, e.g., `#e2e-run-all`, which then triggers all E2E tests to run in a PR instead of just the fast-failing ones. This could be useful if we want to trigger the full execution manually.

Slartibartfass2:
Closes #614 

## What I have made

- Separated the fast-failing and complete e2e test runs
- The fast failing is now only triggered on PRs
- The complete run is only triggered on pushes to main or develop
- Previous E2E runs are cancelled when new changes are pushed
- Added an `#e2e-run-all` flag that can be inserted into a PR so that the complete run can be triggered (the fast-failing run will be skipped in this case)
- There's only one catch: We cannot make those runs required because one of them is skipped, and we can't make a possibly skipped job required. This is now the job of the reviewer.
- I will change the ruleset after this has been approved

Proof:
- [This](https://github.com/SE-UUlm/snowballr-frontend/actions/runs/22714947136) is a run where only the fast-failing tests were executed
- [This](https://github.com/SE-UUlm/snowballr-frontend/actions/runs/22715141012) is a run where the complete E2E test suite was executed

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only CI changes

### Reviewer

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,7 +21,7 @@ Either tick or cross out the items that do not apply (using \~\~example text\~\~
 ### Reviewer
 
 - [ ] I have checked the implementation against the requirements
-- [ ] I have checked for flaky tests in the CI
+- [ ] I have checked that the E2E tests were performed and that they were not flaky.
 
 <!-- Uncomment if the whole E2E test suite should run -->
 <!-- #e2e-run-all -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -22,3 +22,6 @@ Either tick or cross out the items that do not apply (using \~\~example text\~\~
 
 - [ ] I have checked the implementation against the requirements
 - [ ] I have checked for flaky tests in the CI
+
+<!-- Uncomment if the whole E2E test suite should run -->
+<!-- #e2e-run-all -->
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -9,7 +9,7 @@ on:
     workflow_dispatch:
 
 concurrency:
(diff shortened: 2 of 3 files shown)


.github/pull_request_template.md
Merseleo, line 1: Maybe also add a checkbox for verifying that the E2E runs were executed successfully (because they are not required anymore as far as I understand your comment?) Or integrate this check into the flaky e2e tests check.

.github/workflows/e2e_tests.yml
```diff
@@ -6,12 +6,15 @@ on:
             - main
             - develop
     pull_request:
-    merge_group:
     workflow_dispatch:
 
+concurrency:
+    group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
Merseleo: Only a question of interest: Why not only using "gitHub.ref"?

Merseleo (changes requested):
LGTM. I did not found any big issue and I think this should accelerate our PR pipeline.

Merseleo (approved):
LGTM

PR #630, opened 2026-04-01
Use snowballr-api npm package

### LINKED ISSUE #629: [Feature]: Use snowballr-api NPM package

_No response_

Let's use the API NPM package instead of the current workflow.

Using the NPM package.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #629 

## What I have made

We now use the [snowballr-api NPM package](https://www.npmjs.com/package/snowballr-api).

This removes this whole fetch and compile workflow that we've been using for a while.
Most changes are import changes. I've added an import alias so that if we change the location in the future, the alias stays the same and we don't have to adjust all imports all over again.

We can't use the typescript code in the node_modules when compiling the code because it's prohibited to do so (packages should be compiled when publishing). Therefore, as postinstall script, we just copy the files to the previous location.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no new features added)

### Reviewer

- [x] I have checked that the E2E tests were performed and that they were not flaky.

#e2e-run-all


mowi12 (changes requested):
I also did let run copilot over the changes:

 1. Medium: postinstall is Unix-only (cp -r ...) and will fail on Windows shells.
package.json (script postinstall)
 2. Medium: Generated API files are copied without cleanup, so removed/renamed files in newer snowballr-api releases can leave 
stale .ts files in src/lib/model/api.
package.json (postinstall copy strategy), target dir src/lib/model/api/
 3. Medium: snowballr-api is in devDependencies, but postinstall depends on it. Installs using npm ci --omit=dev will fail because
 the source path is missing.
package.json (devDependencies + postinstall coupling)
 4. Low: Docker build still installs Git “for tiged”, but tiged is gone; this is outdated and adds unnecessary image weight.
Dockerfile (comment + apk add git)

Regarding 1. - I'm not sure if we support Windows, so we can keep it as is.
But the rest, you should take a look at.

Slartibartfass2:
@mowi12 I checked the issues and fixed those that were correct.
Regarding 3., we don't use that command, so I don't know why this should be a problem.

Edit: I verified this by running the docker image locally.

mowi12 (approved):
LGTM! Let's finish this

